PR #13, opened 2018-07-04
Add build command

marcomorain:
Port the majority of the current bash script that runs the existing CLI tool.

The allows people to run `circleci build` locally using the new CLI.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -1,10 +1,7 @@
 package cmd
 
 import (
-	"bufio"
-	"context"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -12,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/CircleCI-Public/circleci-cli/settings"
 	"github.com/pkg/errors"
@@ -136,18 +134,7 @@ func picardImage() (string, error) {
 	return fmt.Sprintf("%s@%s", picardRepo, sha), nil
 }
 
-func streamOutout(stream io.Reader) {
-	scanner := bufio.NewScanner(stream)
-	go func() {
-		for scanner.Scan() {
-			Logger.Info(scanner.Text())
-		}
-	}()
-}
-
 func runBuild(cmd *cobra.Command, args []string) error {
-
-	ctx := context.Background()
 
 	pwd, err := os.Getwd()
 
@@ -164,7 +151,7 @@ func runBuild(cmd *cobra.Command, args []string) error {
 	// TODO: marc:
 	// I can't find a way to pass `-it` and have carriage return
 	// characters work correctly.
-	arguments := []string{"run", "--rm",
+	arguments := []string{"docker", "run", "--rm",
 		"-e", fmt.Sprintf("DOCKER_API_VERSION=%s", dockerAPIVersion),
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
 		"-v", fmt.Sprintf("%s:%s", pwd, pwd),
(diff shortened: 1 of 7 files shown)


cmd/build.go
```diff
@@ -0,0 +1,203 @@
...
+
+var job = "build"
+var nodeTotal = 1
+var checkoutKey = "~/.ssh/id_rsa"
+var skipCheckout = true
+
+func newBuildCommand() *cobra.Command {
+
+	command := &cobra.Command{
+		Use:   "build",
```
dnephin: Since this command is still running a job (not a full build) I've been wondering if we should rename it to either `run` or `job run`.  We can always add `build` as an alias for backwards compatbility.

cmd/build.go
```diff
@@ -0,0 +1,203 @@
...
+				--branch string         Git branch
+			-e, --env -e VAR=VAL        Set environment variables, e.g. -e VAR=VAL
+				--index int             node index of parallelism
+				--repo-url string       Git Url
+				--revision string       Git Revision
+			-v, --volume stringSlice    Volume bind-mounting
+	*/
+
+	flags := command.Flags()
+	flags.StringVarP(&dockerAPIVersion, "docker-api-version", "d", dockerAPIVersion, "The Docker API version to use")
```
dnephin: I don't think you want to duplicate all the flags here. There is a better way!

One option is to use `flags.SetIntersperced(false)` https://godoc.org/github.com/spf13/pflag#SetInterspersed. That way any flags after a positional argument (or a `--`) are kept in `Args`. It might also work for any not recognized flags, I don't remember.

You're already appending args so that should be it.

cmd/build.go
```diff
@@ -0,0 +1,203 @@
...
+
+		if err != nil {
+			return "", err
+		}
+
+	} else {
+		// Otherwise pulling latest image in background
+		// TODO - this should write to a file to record
+		// the fact that we have the latest image.
+		newPullLatestImageCommand().Start()
```
dnephin: I think we should reconsider this "pull latest in background on every run". I think it's unexpected behaviour and not a good user experience. This is Windows Updates.

Instead I think we could provide a subcommand to update it.

cmd/build.go
```diff
@@ -0,0 +1,203 @@
...
+
+	image, err := picardImage()
+
+	if err != nil {
+		return errors.Wrap(err, "Could not find picard image")
+	}
+
+	// TODO: marc:
+	// I can't find a way to pass `-it` and have carriage return
+	// characters work correctly.
```
dnephin: Because you are setting stdout to a pipe, which is not a tty. You could use a pty, but I think there is a better option.

Instead of fork/exec, why not just exec (https://golang.org/pkg/syscall/#Exec)? That way you don't need to stream output (not sure how that works on windows). Alternatively just set Stdout/Stderr to os.Stdout/os.Stderr, which also fixes the tty issue and removes the need to stream output.

cmd/build.go
```diff
@@ -0,0 +1,203 @@
...
+	arguments := []string{"run", "--rm",
+		"-e", fmt.Sprintf("DOCKER_API_VERSION=%s", dockerAPIVersion),
```
dnephin: This doesn't need to be a flag. Use `-e DOCKER_API_VERSION` and it will use the value from `cmd.Env`, which defaults to `os.Environ()` which is exactly what you want I think.

cmd/build.go
marcomorain: Oh sweet, that will make things easier 👌

cmd/build.go
marcomorain: That's a good idea. We will need to do something else on Windows I imagine?

dnephin: Ya, I imagine there will be other problems with windows as well. I don't think bind mounting `/var/run/docker.sock` works, because it uses a named pipe if I remember correctly.

marcomorain: Done in 781bebeb9bd5cfd9f22afb3aeb4adca0e91240ea

ndintenfass:
:thatmotivatesme:

PR #17, opened 2018-07-09
[CIRCLE-12198] Update instructions circleci-beta->circleci-cli

hannahhenderson:
Paired with @eric-hu 

This card isn't technically part of [CIRCLE-12198](https://circleci.atlassian.net/browse/CIRCLE-12198), but it has a fair amount of overlap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ You may first need to generate a CircleCI API Token from the [Personal API Token
 $ circleci configure 
 ```
  
-Accept the provided default `CircleCI API End Point`.
+If you are using this tool on `.com`. accept the provided default `CircleCI API End Point`. If you are using it on Server, change the value to your Server address (i.e. `circleci.my-org.com`).
 
 
 ## Validate A Build Config
--- a/README.md
+++ b/README.md
@@ -6,15 +6,25 @@ This project is the seed for CircleCI's new command-line application.
 
 ### 1. Get the latest binary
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,9 +83,11 @@ fragment FullType on __Type {
 You can now pipe that file to the `query` command to send it.
```
johnswanson: Unclear what file I'm piping to the `query` command here now--why was the example removed? At the least we should say something like "If you have a query in the file `query.gql`, you can pipe that file ..."

README.md
```diff
@@ -70,9 +83,11 @@ fragment FullType on __Type {
...
 
 ```
-$ cat query.gql | circleci-beta query
+$ cat query.gql | circleci-cli query
 ```
 
+> You could also copy the query above, and run it directly from your Clipboard `pbpaste | circleci-cli query`.
```
johnswanson: mac specific but probably okay (linux would be `xclip -sel clipboard -o | circleci-cli query`)

README.md
```diff
@@ -6,16 +6,26 @@ This project is the seed for CircleCI's new command-line application.
...
 ```
 
+### 3. Add a Token
+You may first need to generate a CircleCI API Token from the [Personal API Token tab](https://circleci.com/account/api).
+
+```
+$ circleci configure 
+```
+ 
+Accept the provided default `CircleCI API End Point`.
```
eric-hu: Oh I didn't realize this was a prompt.  I thought it printed out only.

In this case, maybe the instructions should instruct Server customers to use their Server address (i.e. `circleci.my-org.com`)

PR #18, opened 2018-07-09
Fix gometalinter flakes by setting CGO_ENABLED=0

dnephin:
I looked into the lint CI failures in #17 by re-running the job with ssh enabled. In the shell I ran 

```
gometalinter --debug --vendor ./...
```
In the debug output I noticed 

```
/go/bin/structcheck returned exit status 1: exec: "gcc": executable file not found in $PATH
```
Often (maybe always) when you get linter errors reported for `/usr/local/go` this indicates that linting is failing because `go build` failed. 

The `gometalinter` docker image uses alpine and doesn't include `gcc` and all the c header files, so running with `CGO_ENABLED=0` fixes the build by using the pure-go compiler.

Also added a config file to explicitly define and configure linters.  I added a few additional linters that aren't in the [default set](https://github.com/alecthomas/gometalinter#supported-linters) but which I find helpful (`gosimple`, `misspell`, `goimports`, `nakedret`, `unparam`). 

I also removed `dupl` from the list, which is enabled by default. In my experience the warnings this reports are often not very valuable, but I would be happy to re-enabled it if you would like.

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -1,0 +1,34 @@
+{
+  "Vendor": true,
+  "Deadline": "5m",
+  "Concurrency": 2,
+  "Linters": {
+    "gofmt": {"Command": "gofmt -l -s -w"},
+    "goimports": {"Command": "goimports -l -w"}
+  },
+
+  "Enable": [
+    "deadcode",
+    "errcheck",
+    "gas",
+    "goconst",
+    "gocyclo",
+    "gofmt",
+    "goimports",
+    "golint",
+    "gosimple",
+    "gotype",
+    "gotypex",
+    "ineffassign",
+    "interfacer",
+    "megacheck",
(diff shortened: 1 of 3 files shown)


.gometalinter.json
```diff
@@ -0,0 +1,34 @@
+{
+  "Vendor": true,
+  "Deadline": "5m",
+  "Concurrency": 2,
+  "Linters": {
+    "gofmt": {"Command": "gofmt -l -s -w"},
+    "goimports": {"Command": "goimports -l -w"}
```
dnephin: This is just for convenience. It adds the `-w` flags for these two commands so that if you run it locally, and there are  errors, the fixes for the errors will be written.

marcomorain:
Awesome, thanks @dnephin

PR #23, opened 2018-07-16
[CIRCLE-12186] optionally read config file from stdin

johnswanson:


cmd/orb.go
```diff
@@ -41,10 +41,10 @@ func newOrbCommand() *cobra.Command {
 
 	orbCommand.AddCommand(orbListCommand)
 
-	orbValidateCommand.PersistentFlags().StringVarP(&orbPath, "path", "p", "orb.yml", "path to orb file")
+	orbValidateCommand.PersistentFlags().StringVarP(&orbPath, "path", "p", "orb.yml", "path to orb file ('-' for STDIN)")
```
zzak: Having to type `-p` when I just want to pipe some file as stdin to a program is painful.

For certain flags, which aren't optional (behavior changing) such as the path to your config, a truly vital option for this config (or we default) shouldn't be a flag IMO.

I'd like if we could just type `circleci orb validate path/to/orb.yml` or leave `path/to/orb.yml` out entirely for the same results (if it were default) and in this use-case `circleci orb validate -`. But I may just be dreaming. :thinking:

ndintenfass: +1 - my understanding is we added path as a flag because we would have a default value of the path, but we should probably kill the default value, have path be required, pass it without a flag. I think we talked about a pattern where we'd try to limit required arguments and let you pass them without a flag. But if that's a major effort we should discuss its relative value.

marcomorain: Let's not make the config path be required. The current CLI tool has a command `circleci config validate` and we don't want to break our customer's workflows.

cmd/orb_test.go
```diff
@@ -32,6 +33,56 @@ var _ = Describe("Orb", func() {
...
+				command = exec.Command(pathCLI,
+					"orb", "validate",
+					"-t", token,
+					"-e", testServer.URL(),
+					"-",
+				)
+				stdin, err := command.StdinPipe()
+				Expect(err).ToNot(HaveOccurred())
+				go func() {
+					defer stdin.Close()
```
eric-hu: 🆒

cmd/orb_test.go
```diff
@@ -32,6 +33,56 @@ var _ = Describe("Orb", func() {
...
+						"config": "{}"
+					}
+				}`
+
+				appendPostHandler(testServer, token, http.StatusOK, expectedRequestJson, gqlResponse)
+				session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+
+				Expect(err).ShouldNot(HaveOccurred())
+				// the .* is because the full path with temp dir is printed
+				Eventually(session.Out).Should(gbytes.Say("Orb at - is valid"))
```
eric-hu: The feedback message from piping in is awkward.  Maybe this should be something like `orb passed from stdin is valid` instead?

```
$ cat ./.circleci/config.yml | go run main.go orb validate -
Orb at - is valid
```

zzak:
ca96a7f is a quick refactor to clean up the cruft of building an expect json response that matches the formatting given from `api` package.

screenshot from 2018-07-23 12-08-22

cmd/cmd_suite_test.go
```diff
@@ -71,19 +71,25 @@ func (f tmpFile) write(fileContent string) error {
 	return err
 }
 
-func openTmpFile(path string) (tmpFile, error) {
+func openTmpDir(prefix string) (string, error) {
```
marcomorain: I don't think `prefix` is ever used here - copy + paste bug?

zzak:
We can now support optionally a positional arg to specify the input file, even as stdin with "-", or fall back to the default.

This includes the following commands:

* `config expand`
* `config validate`
* `orb expand`
* `orb publish`
* `orb validate`

## config expand

```
/tmp/tmp.7fGNduimQt => ./circleci help config expand
Expand the config.

Usage:
  circleci config expand [config.yml] [flags]

Flags:
  -h, --help   help for expand

Global Flags:
  -e, --endpoint string   the endpoint of your CircleCI GraphQL API (default "https://circleci.com/graphql-unstable")
  -t, --token string      your token for using CircleCI
  -v, --verbose           Enable verbose logging.
/tmp/tmp.7fGNduimQt => ./circleci config expand
Error: required key [version] not found
/tmp/tmp.7fGNduimQt => cat .circleci/config.yml 
{}
```

## config validate

```
/tmp/tmp.7fGNduimQt => ./circleci help config validate
Check that the config file is well formed.

Usage:
  circleci config validate [config.yml] [flags]

Aliases:
  validate, check

Flags:
  -h, --help   help for validate

Global Flags:
  -e, --endpoint string   the endpoint of your CircleCI GraphQL API (default "https://circleci.com/graphql-unstable")
  -t, --token string      your token for using CircleCI
  -v, --verbose           Enable verbose logging.
/tmp/tmp.7fGNduimQt => ./circleci config validate
Error: required key [version] not found
/tmp/tmp.7fGNduimQt => ./circleci config validate .circleci/config.yml 
Error: required key [version] not found
/tmp/tmp.7fGNduimQt => cat .circleci/config.yml | ./circleci config validate -
Error: required key [version] not found
```

## orb expand

```
/tmp/tmp.7fGNduimQt => ./circleci help orb expand
expand an orb.yml

Usage:
  circleci orb expand [orb.yml] [flags]

Flags:
  -h, --help   help for expand

Global Flags:
  -e, --endpoint string   the endpoint of your CircleCI GraphQL API (default "https://circleci.com/graphql-unstable")
  -t, --token string      your token for using CircleCI
  -v, --verbose           Enable verbose logging.
```

## orb publish

```
/tmp/tmp.7fGNduimQt => ./circleci help orb publish
publish a version of an orb

Usage:
  circleci orb publish [orb.yml] [flags]

Flags:
  -h, --help                 help for publish
  -i, --orb-id string        id of orb to publish
  -o, --orb-version string   version of orb to publish

Global Flags:
  -e, --endpoint string   the endpoint of your CircleCI GraphQL API (default "https://circleci.com/graphql-unstable")
  -t, --token string      your token for using CircleCI
  -v, --verbose           Enable verbose logging.
```

## orb validate

```
/tmp/tmp.7fGNduimQt => ./circleci help orb validate
validate an orb.yml

Usage:
  circleci orb validate [orb.yml] [flags]

Flags:
  -h, --help   help for validate

Global Flags:
  -e, --endpoint string   the endpoint of your CircleCI GraphQL API (default "https://circleci.com/graphql-unstable")
  -t, --token string      your token for using CircleCI
  -v, --verbose           Enable verbose logging.
```

cmd/cmd_suite_test.go
zzak: It's used here:
https://github.com/CircleCI-Public/circleci-cli/pull/23/files/1c600eaec59f3c63688668b38f7e6753e88de3bb#diff-bee1de2bffa142cede85d49cea1a90f7R76

So you can create a temp directory without specifying the prefix, just do `openTmpDir("")`.

PR #27, opened 2018-07-18
Circle-11123 publish orb

eric-hu:
User can publish an orb from the CLI

```
$ go run main.go orb publish --path ./tmp-orb.yml -i bb604b45-b6b0-4b81-ad80-796f15eddf87 -o 0.0.2
Orb published

$ go run main.go orb publish --path ./tmp-orb.yml -i bb604b45-b6b0-4b81-ad80-796f15eddf87 -o 0.0.2
Error: orb revision already exists exit status 255
```

PR #28, opened 2018-07-20
[CIRCLE-12217] Create namespace

hannahhenderson:
This PR adds "create namespace" to the list of orb commands (or, more specifically `createns`, which is open to alternate naming suggestions).

I used a lot of copy-pasting to make this work, thank you @eric-hu . I also benefited from some good old, "Please help me see the forest, oh :rubberduck: @johnswanson, because at this point in time I can only see the trees," unsticking. 

Which is all to say, this PR works as expected, but I believe it is a good candidate for some DRY refactoring.

[CIRCLE-12217](https://circleci.atlassian.net/browse/CIRCLE-12217)

Examples of this in action in dev:
_Ignore the weird UUID output in earlier responses, I deleted some print lines_
_Also, "GITHUB" and "BITBUCKET" can now be lower case_
screen shot 2018-07-24 at 10 16 38 am

Help lists this information:
screen shot 2018-07-24 at 11 11 32 am

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -5,6 +5,8 @@ import (
 	"io/ioutil"
 	"strings"
 
+	"fmt"
+
 	"github.com/CircleCI-Public/circleci-cli/client"
 	"github.com/CircleCI-Public/circleci-cli/logger"
 	"github.com/pkg/errors"
@@ -36,6 +38,17 @@ type PublishOrbResponse struct {
 	Orb struct {
 		CreatedAt string
 		Version   string
+	}
+
+	GQLResponseErrors
+}
+
+// CreateNamespaceResponse type matches the data shape of the GQL response for
+// creating a namespace
+type CreateNamespaceResponse struct {
+	Namespace struct {
+		CreatedAt string
+		ID        string
 	}
 
 	GQLResponseErrors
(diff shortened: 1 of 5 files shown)


zzak:
@hannahhenderson I will do a formal review soon, but I think we should make `--name` a positional argument.

johnswanson:
This is all UX, not go code review :)

What about `orb ns create`? (Similar to `docker network ls`, `docker network rm`, `docker network create`--specifying the thing you're operating on, then a command to run) Alternatively, `orb create-ns`?

I also think we should wire it up so you can do:

```go run main.go orb create-ns --org "circleci" foo```

rather than having to manually look up your org ID. This will require another roundtrip to query for the ID of the org called `circleci`, but I don't think we can force the user to look up the ID for their orgs :)

Super awesome to see this working!

File: api/api.go [hunk 1]
```diff
@@ -101,3 +101,70 @@ func OrbQuery(ctx context.Context, logger *logger.Logger, configPath string) (*C
...
+		Message string
+		Type string
+	}
+}
+
+func (response CreateNamespaceResponse) ToError() error {
+	messages := []string{}
+
+	for i := range response.Errors {
+		messages = append(messages, response.Errors[i].Message)
```
marcomorain: Looks like you are not using the `type` field here. You should either use it, or drop it from the GQL request.

api/api.go
```diff
@@ -101,3 +101,70 @@ func OrbQuery(ctx context.Context, logger *logger.Logger, configPath string) (*C
...
+	}
+
+	return errors.New(strings.Join(messages, ": "))
+
+}
+
+func createNamespaceQuery(ctx context.Context, logger *logger.Logger, name string, ownerId string, response interface{}, query string) error {
```
marcomorain: can you in-line this function into OrbCreateNamespace please? - I asked Eric to do the same at https://github.com/CircleCI-Public/circleci-cli/pull/27#discussion_r203738439 

Indeed if we merge Eric's PR and you re-base on top you can re-use his new Errors struct to remove `(response CreateNamespaceResponse) ToError()`

api/api.go
hannahhenderson: ✅

api/api.go
hannahhenderson: ✅

hannahhenderson:
@johnswanson I think that not forcing folks to use a `namespaceId` when they're adding orbs makes sense, because we are enforcing uniqueness on namespace names.

Organization names aren't unique. Does it make more sense to add a separate "get organization id" CLI call (name + vcs type necessary), or to add all of that info to this call?

api/api.go
```diff
@@ -164,3 +176,92 @@ func OrbPublish(ctx context.Context, logger *logger.Logger,
...
+
+	request := client.NewAuthorizedRequest(viper.GetString("token"), query)
+	request.Var("organizationName", organizationName)
+	request.Var("organizationVcs", organizationVcs)
+
+	graphQLclient := client.NewClient(viper.GetString("endpoint"), logger)
+
+	err := graphQLclient.Run(ctx, request, &response)
+
+	fmt.Println(response)
```
eric-hu: I think you want to delete this line before merging? (In my editor, doing this would also get rid of the "fmt" import at the top on save.  I think a default go tool does that, but not sure which.)

PR #30, opened 2018-07-23
Update README.md: add instructions for cleaning up development install

eric-hu:


zzak:
Does `make clean` not take care of this?

zzak:
@eric-hu Thanks for updating this! Though I think this information belongs in `CONTRIBUTING.md` as that is geared for development docs.

eric-hu:
@zzak 👍 updated

PR #32, opened 2018-07-25
Create namespace

zzak:


api/api.go
```diff
@@ -63,7 +74,7 @@ func (response GQLResponseErrors) ToError() error {
 		messages = append(messages, response.Errors[i].Message)
 	}
 
-	return errors.New(strings.Join(messages, ": "))
+	return fmt.Errorf(strings.Join(messages, ": "))
```
marcomorain: Why the change from the `errors` package to string-formatting (`fmt`) package here? IMO using the errors package is  more clear of the intent, and more clear still since there is no string-formatting going on.

api/api.go
```diff
@@ -243,7 +254,7 @@ func getOrganization(ctx context.Context, logger *logger.Logger, organizationNam
 	err := graphQLclient.Run(ctx, request, &response)
 
 	if err != nil || response.Organization.ID == "" {
-		err = errors.Wrap(err, fmt.Sprintf("Unable to find organization %s of vcs-type %s", organizationName, organizationVcs))
+		err = fmt.Errorf("Unable to find organization %s of vcs-type %s", organizationName, organizationVcs)
```
marcomorain: This change is destructive - the original error is lost. Let's suppose that error is a network failure, we want the error message to include this information, so that the user can take action.

api/api.go
```diff
@@ -54,6 +54,17 @@ type CreateNamespaceResponse struct {
 	GQLResponseErrors
 }
 
+// CreateOrbResponse type matches the data shape of the GQL response for
+// creating an orb
+type CreateOrbResponse struct {
+	Orb struct {
+		CreatedAt string
```
marcomorain: `createdAt` doesn't seem to be used anywhere. Let's remove from the GQL query and the response struct.

api/api.go
```diff
@@ -256,11 +267,95 @@ func CreateNamespace(ctx context.Context, logger *logger.Logger, name string, or
...
+	request := client.NewAuthorizedRequest(viper.GetString("token"), query)
+	request.Var("name", name)
+
+	graphQLclient := client.NewClient(viper.GetString("endpoint"), logger)
+
+	err := graphQLclient.Run(ctx, request, &response)
+
+	if err != nil || response.RegistryNamespace.ID == "" {
+		err = fmt.Errorf("Unable to find namespace %s", name)
+	}
```
marcomorain: A couple of comments here:

- Let's split the case of when `err` is nil from the checking of `response.RegistryNamespace.ID`. We want to either return the error, or a wrapper around the error when it is not null.
- When can `response.RegistryNamespace.ID` be empty/null? That seems like a fundamental failure in the API schema.

cmd/diagnostic.go
```diff
@@ -25,7 +25,7 @@ func diagnostic(cmd *cobra.Command, args []string) error {
 	Logger.Infof("GraphQL API endpoint: %s\n", endpoint)
 
 	if token == "token" || token == "" {
-		return errors.New("please set a token")
+		return fmt.Errorf("please set a token")
```
marcomorain: Why use a formatting function here for a static string?

api/api.go
```diff
@@ -256,11 +267,95 @@ func CreateNamespace(ctx context.Context, logger *logger.Logger, name string, or
...
+		return nil, err
+	}
+
+	orb, err := createOrbWithNsID(ctx, logger, name, namespaceID)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return orb, err
```
marcomorain: If you `return createOrbWithNsID(ctx, logger, name, namespaceID)` then you can delete the next 5 lines of code.

api/api.go
```diff
@@ -256,11 +267,95 @@ func CreateNamespace(ctx context.Context, logger *logger.Logger, name string, or
...
+	request := client.NewAuthorizedRequest(viper.GetString("token"), query)
+	request.Var("name", name)
+	request.Var("registryNamespaceId", namespaceID)
+
+	graphQLclient := client.NewClient(viper.GetString("endpoint"), logger)
+
+	err := graphQLclient.Run(ctx, request, &response)
+
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("Unable to create orb %s for namespaceID %s", name, namespaceID))
```
marcomorain: In looking into the difference between `fmt.Errorf` and `errors.New` I learned that [`errors.Wrapf`](https://github.com/pkg/errors/blob/master/errors.go#L197) exists. That will neaten a bunch of code up.

api/api.go
zzak: This may be due to the nature of how these 2 PRs were merged, sorry about that.

I will fix it and address the other comments tomorrow unless @johnswanson or @hannahhenderson get to it.

hannahhenderson: The linter actually screamed at me for `errors.New` and demanded `fmt.Errorf`. I assumed that was because the error, in this case, is only a formatted string.

api/api.go
hannahhenderson: Fair enough. This will need to become an if-else statement, then. 

From what @johnswanson and I could tell, when the error is `nil` (like when the response is empty), if you wrap it with a string the end result is still `nil` (so the subsequent `err != nil` checks don't behave as-desired).

Do you know if there's a better way to handle this?

hannahhenderson:
@zzak #28 should have come before #31. This is the work for creating an orb, although it has a few string-updates for creating a namespace.

api/api.go
eric-hu: @marcomorain @hannahhenderson The change isn't destructive.  These two lines have equivalent return values.  `fmt.Errorf` returns an `error` object with string interpolation:

https://golang.org/pkg/fmt/#Errorf

See also: https://www.reddit.com/r/golang/comments/6ffrie/fmterrorf_or_errorsnew/

api/api.go
eric-hu: @hannahhenderson If linter warnings failing builds becomes a nuisance, I think we can check for them in bash and choose to not fail the build:

https://github.com/alecthomas/gometalinter#exit-status

PR #33, opened 2018-07-26
[CIRCLE-12217] Require flags (better error messaging)

hannahhenderson:
Per @eric-hu 's feedback on [CIRCLE-12217](https://circleci.atlassian.net/browse/CIRCLE-12217), the error messaging was a little opaque.

As can be seen in my first commit, it _looks_ like there's a super-handy "require flag" option in Cobra, but on closer inspection, it seems like implementing that solution requires a fair amount of DIY force-updates (see [cobra#206](https://github.com/spf13/cobra/issues/206)).

We should also update our `orb publish` option to accept a namespace-qualified orb name instead of an orbID.

screen shot 2018-07-26 at 4 52 15 pm


johnswanson:
We might need this elsewhere, so doing it in a reusable way would be cool--what if we write something like:

```
func requireArgs(required ...string) func(*cobra.Command, []string) error {
    return func(cmd *cobra.Command, args []string) error {
        // checks for missing flags, returns error if some are missing
    }
}
command := cobra.Command{ PreRunE: requireArgs("foo") }
```

cmd/orb.go
```diff
@@ -84,7 +85,7 @@ func newOrbCommand() *cobra.Command {
 
 	orbCommand.AddCommand(orbPublishCommand)
 
-	orbCreateNamespace.PersistentFlags().StringVar(&organizationName, "org-name", "", "organization name")
+	orbCreateNamespace.PersistentFlags().StringVar(&organizationName, "org-name", "", "organization name (required)")
```
marcomorain: You can use cobra to mark flags required:

```go
orbCreateNamespace.MarkPersistentFlagRequired("org-name")
```

cmd/orb.go
```diff
@@ -276,6 +277,10 @@ func publishOrb(cmd *cobra.Command, args []string) error {
 		orbPath = args[0]
 	}
 
+	if orbPath == "" || orbID == "" {
+		return fmt.Errorf("Must include orb-path and orb-id flags")
```
marcomorain: using cobra required flags means that you don't need to handle

cmd/orb.go
hannahhenderson: @marcomorain it doesn't look like `MarkPersistentFlagRequired` exists in Cobra. I used the same `panic` loop that you did in PR #35 ([see also](https://github.com/CircleCI-Public/circleci-cli/pull/35/files#diff-75d982f75dc436959ce23e09bf08d0fdR55)) and that works... but a little too aggressively. Now I get a panic result even when I _do_ include the flag. Any idea how to fix this? 

```
✗ go run main.go orb ns create sdf --org-name "circleci"
panic: no such flag -org-name

goroutine 1 [running]:
github.com/CircleCI-Public/circleci-cli/cmd.newOrbCommand(0xc420180000)
	/Users/hannah/go/src/github.com/CircleCI-Public/circleci-cli/cmd/orb.go:95 +0xcf4
github.com/CircleCI-Public/circleci-cli/cmd.MakeCommands(0x0)
	/Users/hannah/go/src/github.com/CircleCI-Public/circleci-cli/cmd/root.go:45 +0x2d4
github.com/CircleCI-Public/circleci-cli/cmd.Execute()
	/Users/hannah/go/src/github.com/CircleCI-Public/circleci-cli/cmd/root.go:21 +0x22
main.main()
	/Users/hannah/go/src/github.com/CircleCI-Public/circleci-cli/main.go:9 +0x20
exit status 2```

File: orb.yml [hunk 3]
```diff
@@ -0,0 +1 @@
+@@@
```
marcomorain: I guess you didn't mean to add this

hannahhenderson: oh. no. i added this as part of local testing. good catch!

marcomorain:
Might be worth squashing before merging.

PR #34, opened 2018-07-27
Some UX issues

marcomorain:


cmd/orb.go
```diff
@@ -49,8 +49,14 @@ func newOrbCommand() *cobra.Command {
 		RunE:  publishOrb,
 		Args:  cobra.MaximumNArgs(1),
 	}
-	orbPublishCommand.PersistentFlags().StringVarP(&orbVersion, "orb-version", "o", "", "version of orb to publish")
-	orbPublishCommand.PersistentFlags().StringVarP(&orbID, "orb-id", "i", "", "id of orb to publish")
+	orbPublishCommand.Flags().StringVarP(&orbVersion, "orb-version", "o", "", "version of orb to publish")
+	orbPublishCommand.Flags().StringVarP(&orbID, "orb-id", "i", "", "id of orb to publish")
+
+	for _, flag := range [2]string{"orb-version", "orb-id"} {
+		if err := orbPublishCommand.MarkFlagRequired(flag); err != nil {
```
johnswanson: I'm pretty sure this doesn't actually work. See: https://github.com/spf13/cobra/issues/206

johnswanson: Aha, nevermind, it only didnt't work with PersistenFlags, which we don't want here anyway. :+1:

PR #38, opened 2018-07-31
[CIRCLE-12762] Command cosmetics cleanup

zzak:


zzak:
This is ready for review, short of the change to VCS to short-form only.

PR #40, opened 2018-08-02
Generate docs for the CLI

zzak:
You can see a preview here:
https://852-137903064-gh.circle-artifacts.com/0/docs/circleci.html

Which was generated by this build:
https://circleci.com/gh/CircleCI-Public/circleci-cli/852#artifacts/containers/0

---

screenshot from 2018-08-02 16-36-55


zzak:
Pending review, this has been deployed here:
https://circleci-public.github.io/circleci-cli/

cmd/usage.go
```diff
@@ -0,0 +1,86 @@
...
+// nolint: gosec
+func isCmdAvailable(name string) bool {
+	cmd := exec.Command("/bin/sh", "-c", "command", "-v", name)
+	if err := cmd.Run(); err != nil {
+		return false
+	}
+	return true
+}
+
+func usage(cmd *cobra.Command, args []string) error {
```
marcomorain: I’d rather we did the call to Pandoc from outside go. I’d like to generate the .md content from go, and the exit. 

Then in config.yml call out to pandoc to generate HTML.

.circleci/deploy-gh-pages.sh
```diff
@@ -0,0 +1,52 @@
...
+else
+    git checkout --orphan gh-pages
+fi
+
+# copy over or recompile the new site
+cp -a "../docs/." .
+
+# move index and update links
+mv circleci.html index.html
+sed -i -- 's#<a href="circleci.html">#<a href="index.html">#g' *.html
```
marcomorain: Can you do this moving and replacement from somewhere else, so that you his script is just concerned with pushing to gh-pages. 

Having these two lines in the middle of this script is surprising to the reader.

cmd/root.go
```diff
@@ -13,6 +13,8 @@ import (
 
 var defaultEndpoint = "https://circleci.com/graphql-unstable"
 
+var rootCmd *cobra.Command
```
marcomorain: Why did you make a global here?

zzak:
After further review, it looks like the change to `Gopkg.lock` reflects the latest changes to `go dep`, including those added in this PR:
https://github.com/golang/dep/pull/1912

Which seems to have been released about 10 days ago:
https://github.com/golang/dep/releases/tag/v0.5.0

.circleci/generate-docs.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# generate markdown docs
+go run main.go usage
+
+# setup output dir
+mkdir -p out
+
+# pandoc markdown to html into out dir
+for f in docs/*.md;  do
+    pandoc "$f" -s -o "out/$(basename ${f%.md}.html)";
```
marcomorain: Nice

PR #42, opened 2018-08-03
CIRCLE-12811: Ensure consistency on positional arguments vs. flags

zzak:


cmd/config.go
```diff
@@ -12,32 +12,41 @@ import (
...
 
 	validateCommand := &cobra.Command{
-		Use:     "validate [config.yml]",
+		Use:     "validate PATH (use \"-\" for STDIN)",
 		Aliases: []string{"check"},
 		Short:   "Check that the config file is well formed.",
 		RunE:    validateConfig,
 		Args:    cobra.MaximumNArgs(1),
 	}
+	validateCommand.PersistentFlags().StringVarP(&configPath, "path", "p", ".circleci/config.yml", "path to build config")
```
marcomorain: The arg in build agent is not -p - it’s something else, -c I think. I assume that’s what you meant here?

marcomorain: There is a constant at the top of this file that defines the default path.

cmd/namespace.go
```diff
@@ -15,7 +15,7 @@ func newNamespaceCommand() *cobra.Command {
 	}
 
 	createCmd := &cobra.Command{
-		Use:   "create [name] [vcs] [org-name]",
+		Use:   "create NAME github|bitbucket ORG-NAME",
 		Short: "create an namespace",
```
marcomorain: This needs more docs about what an org name is, and how to know whether to specify GitHub/ButBucket.

cmd/namespace.go
```diff
@@ -15,7 +15,7 @@ func newNamespaceCommand() *cobra.Command {
...
 		RunE:  createNamespace,
```
marcomorain: Typo: create a namespace.

cmd/config.go
zzak: True, I actually copied this from earlier in the history of this repo when it was originally added but looks like we use `-c` or `--config`.

cmd/namespace.go
zzak: There is a card to add positional args to the `help` docs for command usage, that should clear this up. I'm afraid of adding too much to the `use` line that would just become clutter.

zzak:
VCSType is documented here: https://github.com/CircleCI-Public/circleci-cli/pull/43/commits/99d88af526db45bbb3c2a7fee567001998c6d1a3#diff-a740d250c5d2f5f2bee7208ac3c67732

Waiting to merge this PR before I will merge #43

api/api.go
```diff
@@ -399,7 +400,8 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, name string, namespac
 }
 
 // OrbSource gets the source or an orb
-func OrbSource(ctx context.Context, logger *logger.Logger, name string) (string, error) {
+func OrbSource(ctx context.Context, logger *logger.Logger, namespace string, orb string) (string, error) {
```
DanToml: Code consistency:

Is an `orb` `namespace/name` or is a `name` `namespace/orb`? (CreateOrb uses `namespace/name`, but this uses `namespace/orb`, which will be potentially confusing to future contributors)

cmd/config.go
```diff
@@ -47,13 +56,21 @@ func newConfigCommand() *cobra.Command {
 	return configCmd
 }
 
+// The PATH arg is actually optional, in order to support compatibility with the --path flag.
 func validateConfig(cmd *cobra.Command, args []string) error {
-	configPath := defaultConfigPath
+	path := defaultConfigPath
```
DanToml: Could potentially extract the code for resolving against the default path into it's own function?

cmd/config.go
```diff
@@ -63,17 +80,13 @@ func validateConfig(cmd *cobra.Command, args []string) error {
 		return response.ToError()
 	}
 
-	Logger.Infof("Config file at %s is valid", configPath)
+	Logger.Infof("Config file at %s is valid", path)
 	return nil
 }
 
 func expandConfig(cmd *cobra.Command, args []string) error {
 	ctx := context.Background()
```
DanToml: Is there a reason why `expand` has different config search behaviour than `validate`? - Seems like something that will be confusing.

cmd/config.go
zzak: Only validate allows a default, for compatibility reasons, so there is no point in refactoring it out.

cmd/config.go
zzak: Validate is treated different for compatibility reasons, those should be documented in the comments -- but if that's not clear then we should fix it.

api/api.go
zzak: That's a good point, and one inconsistency we should fix before shipping.

I will make a follow up PR to fix this though.

FWIW, they should all be `orb [command] NAMESPACE ORB` as we don't want to start inventing a uri-like syntax.

PR #44, opened 2018-08-06
Update readme to reflect new install and upgrade paths

zzak:


README.md
```diff
@@ -4,19 +4,34 @@ This project is the seed for CircleCI's new command-line application.
...
-### 1. Get the latest binary
+### From Scratch
 
-Download the [latest release](https://github.com/CircleCI-Public/circleci-cli/releases/latest) from GitHub for your operating system. If you're on a Mac, this would be `circleci-cli_0.1.XXX_darwin_amd64.tar.gz`.
+If you haven't already installed `circleci` on your machine, run the following command:
 
-### 2. Put the binary in your $PATH
+```
+curl -o /usr/local/bin/circleci \
+  https://raw.githubusercontent.com/zzak/local-cli/switch/circleci.sh && \
```
ndintenfass: This should probably be in the circleci-public/circleci-cli repo - the only reason for "local-cli" was because we had a difference between the build agent and local, but now that they are the same we should unify, yes? In any case, we can't have "zzak" in our critical paths.

README.md
```diff
@@ -4,19 +4,34 @@ This project is the seed for CircleCI's new command-line application.
...
+  chmod +x /usr/local/bin/circleci
+```
+
+The CLI, `circleci`, is downloaded to the `/usr/local/bin` directory.
+
+If you do not have write permissions for `/usr/local/bin`, you might need to run the above commands with `sudo`.
+
+### Upgrade from existing CLI
```
ndintenfass: We should probably put this first, to avoid confusion (especially during the preview period)

PR #48, opened 2018-08-07
Add `circleci tests glob` for backwards compatibility

zzak:


zzak:
Shellcheck was fixed here: 263746b

Just waiting for that PR to get merged.

cmd/root_test.go
```diff
@@ -16,7 +16,7 @@ var _ = Describe("Root", func() {
 
 		It("can create commands", func() {
 			commands := cmd.MakeCommands()
-			Expect(len(commands.Commands())).To(Equal(10))
+			Expect(len(commands.Commands())).To(Equal(11))
```
eric-hu: Nit: I think we can delete this test.  It's not adding much value to have to increment it each time we add a command.

marcomorain:
LGTM

PR #49, opened 2018-08-07
WIP: Fix broken goreleaser config

zzak:
This fixes a config issue:

```
line 21: field command not found in type config.SnapcraftAppMetadata
```

Though, I'm not sure why `command` is invalid metadata as that's exactly what we do for `local-cli`:
https://github.com/circleci/local-cli/blob/dfdab774f712cefb0a897c5b23a498cc8b434e70/snap/snapcraft.yaml#L27

In any case, we still need access to `SNAPCRAFT_LOGIN_FILE` in order to push to the snap store. /cc @felicianotech

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,9 +127,9 @@ workflows:
       - docs:
           requires:
             - deploy
-      - snap:
-          requires:
-            - deploy
+      #- snap:
+      #    requires:
+      #      - deploy
       - deploy:
           requires:
             - test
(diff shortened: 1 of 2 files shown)


marcomorain:
:shipit:

eric-hu:
> Though, I'm not sure why command is invalid metadata as that's exactly what we do for local-cli

That file in `local-cli` is `snapcraft.yaml`, whereas your change is to `.goreleaser.yml`.  Are these two yaml files meant for the same purpose?

zzak:
@eric-hu as in, it's supported data in snapcraft's config! Why wouldn't it be supported in both? :)

PR #53, opened 2018-08-08
Hide build command, and add 'local execute' as the canonical command

dnephin:
This is my proposal for renaming the `build` command to something that makes a bit more sense with the new terminology.

A `build` is one or more workflows, and this command only runs a job, so I don't think the term build can appear in the name.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -32,11 +32,11 @@ type proxyBuildArgs struct {
 	envVarArgs  []string
 }
 
-func newBuildCommand() *cobra.Command {
+func newLocalExecuteCommand() *cobra.Command {
 	buildCommand := &cobra.Command{
-		Use:                "build",
-		Short:              "Run a build",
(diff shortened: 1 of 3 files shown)


marcomorain:
I'd rather we mark `build` as deprecated, rather than hidden, for some period, maybe 2 months, and then hide it after that. The goal being to make sure users are not confused by the change.

dnephin:
We synced on this briefly yesterday. `circleci help build` and `circleci build --help` will still work if the user remembers the command. If they are looking for what command to run `circleci --help` will show `local` with a description that should identify it as the command to run.

marcomorain:
LGTM

dnephin:
Let me know what I should do about the failing build. If I should wait until the issue is resolved and rebase, or merge now.

zzak:
@dnephin Thank you!!

PR #55, opened 2018-08-10
Proxy in-build commands to old CLI

zzak:


proxy/proxy.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"os"
+	"os/exec"
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+func Exec(command string, args []string) error {
+	agent, err := exec.LookPath("picard")
```
marcomorain: @zzak @dnephin what binary name should we be using here - can we use `circlec-agent`?

zzak: `picard` is for sure supported, my guess was until we transition the name this can be a working placeholder and update the path later?

dnephin: `/bin/picard` only work with the docker executor, machine executor uses only `circleci` so unfortunately there is no consistent name we can use right now. 

I opened a PR yesterday to make it use `circleci-agent` consistently. Once that merges we should use that.

proxy/proxy.go
```diff
@@ -0,0 +1,19 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Could not find `picard` executable on $PATH; please ensure that build-agent is installed")
```
dnephin: The error returned by `exec.LookPath` (https://golang.org/src/os/exec/lp_unix.go#L34, https://golang.org/src/os/exec/exec.go#L44) will be something like 

```
exec: picard: executable not found in $PATH
```

So in the wrap message we probably don't want to duplicate those details. I think we could say something like `errors.Wrapf(err, "failed to proxy command %s", command)`. 

If we only expect this to work inside of a circleCI job we could explain that as well.

proxy/proxy.go
```diff
@@ -0,0 +1,19 @@
...
+	}
+
+	err = syscall.Exec(agent, args, os.Environ()) // #nosec
```
dnephin: `syscall.Exec` doesn't work on windows (I believe), but that's fine for now.

It seems the args are wrong here
* command is not used anywhere. I noticed that command is currently a string with spaces. `exec` will not perform shell splitting on these args, so it will need to be `[]string{"step", "halt"}`, which I guess needs to get prepended to the `args`
* I believe you also need to prepend `agent` to `args` as well as pass it as the first argument (https://en.wikipedia.org/wiki/Exec_(system_call)#Arguments)

zzak: Looks like exec uses argv0 for the first arg:

`func Exec(argv0 string, argv []string, envv []string)`

cmd/config.go
```diff
@@ -49,9 +50,21 @@ func newConfigCommand() *cobra.Command {
...
 
+	migrateCommand := &cobra.Command{
+		Use:                "migrate",
+		Short:              "migrate a pre-release 2.0 config to the official release version",
+		RunE:               migrateConfig,
+		Hidden:             true,
+		DisableFlagParsing: true,
+	}
+	// These flags are for documentation and not actually parsed
+	migrateCommand.PersistentFlags().StringP("config", "c", ".circleci/config.yml", "path to config file (default \".circleci/config.yml\")")
```
dnephin: I believe the `(default ...)` is automatically added by cobra, so it would be included twice if it's in the help text.

proxy/proxy.go
```diff
@@ -0,0 +1,23 @@
...
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+// Exec will invoke the given command and proxy any arguments for backwards compatibility.
+func Exec(command []string, args []string) error {
+	agent, err := exec.LookPath("circleci-agent")
+	if err != nil {
+		return errors.Wrap(err, "Could not find `picard` executable on $PATH; please ensure that build-agent is installed")
```
dnephin: I believe https://github.com/CircleCI-Public/circleci-cli/pull/55#discussion_r209290410 is still relevant. The text about "could not find x in path" will be duplicated because it's already in `err`.

dnephin:
`circleci-agent` is now live, so this should be unblocked

proxy/proxy.go
zzak: Whoops, that is correct.

The error was going to be:

```
Error: Could not find `picard` executable on $PATH; please ensure that build-agent is installed: exec: "picard": executable file not found in $PATH
```

Now I've changed it to be:

```
Error: please ensure that build-agent is installed: exec: "picard": executable file not found in $PATH
```

cmd/config.go
zzak: Nice catch, here's what it would have been :face_with_head_bandage: 

```
  -c, --config string   path to config file (default ".circleci/config.yml") (default ".circleci/config.yml")
```

PR #56, opened 2018-08-14
Snap

zzak:


zzak:
The tests are failing for unrelated reasons, would like a review of this approach and sanity check :bow:

PR #57, opened 2018-08-14
Use gometalinter docker image now that it's updated

zzak:
https://github.com/supinf/dockerized-tools/commit/3030b7033b8c0c69fb957406814f9ecc77cf968a


zzak:
It looks like `gometalinter` hasn't released yet since renaming `gosec`:
https://github.com/alecthomas/gometalinter/pull/505

So we'll just have to hold off on this, since we can't add this to our `go dep`, and using a tag with `go get` seems to be near impossible -- plus I don't want to have to manually install a binary release on every machine I own.

PR #72, opened 2018-08-23
Allow resetting endpoint via `circleci setup` if its not default

zzak:
## Don't reset, keeping old endpoint

```
$ go run main.go setup
✔ CircleCI API Token: 
API token has been set.
✔ CircleCI Host: https://circleci.com
CircleCI host has been set.
✗ Do you want to reset the endpoint? (default: graphql-unstable): 
Setup complete. Your configuration has been saved.
```

Config:

```yaml
endpoint: https://circleci.com/graphql
host: https://circleci.com
token: ""
verbose: false
```

## Reset to default endpoint

```
$ go run main.go setup
✔ CircleCI API Token: 
API token has been set.
✔ CircleCI Host: https://circleci.com
CircleCI host has been set.
✔ Do you want to reset the endpoint? (default: graphql-unstable): y
Setup complete. Your configuration has been saved.
```

Config:

```yaml
endpoint: graphql-unstable
host: https://circleci.com
token: ""
verbose: false
```

## Other changes

Also, don't accept host unless it's absolute. (Note, this change affects commands and is not a `setup`-time validation).

For example, a non-absolute host:

```
$ go run main.go diagnostic --host bar.bzasdf
Error: Host (bar.bzasdf) must be absolute URL, including scheme!
```

PR #75, opened 2018-08-23
Ignore setup.go for Codecov

eric-hu:
Following up on [this PR](https://github.com/CircleCI-Public/circleci-cli/pull/72#issuecomment-415568649):

Ignore `setup.go` for code coverage purposes. `setup_test.go` only makes assertions on what `setup.go` prints to STDOUT and will miss 1 of 2 code paths for any conditional. This was adding noise to our PRs and to the implementation code, as `setup.go` accepts a hidden `--testing` flag that makes it ignore user input.

Cleaning up test code from `setup.go` is more involved and can wait until after EOL. I'll make a JIRA for that if/when this PR is accepted.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - cmd/setup.go


zzak:
I'd have to agree with Daniel's comment from the other PR.

We should find a way to test this code, if promptui is making that difficult we should abstract the business logic parts so their easier to test.

This project uses codecov, but it's not a requirement for merging PRs or making progress on the CLI. I take our low (45%) coverage as a result of prioritizing integration tests which don't count towards that number. In the end, it's just a number :ok_man:

PR #76, opened 2018-08-23
[CIRCLE-12608] Add validateArgs to newLocalExecute command

hannahhenderson:
Originally proposed: https://github.com/circleci/build-agent/pull/796

There are some linting/testing issues that need to be sorted out before this PR is ready to go.
https://github.com/spf13/cobra#positional-and-custom-arguments

Co-authored by @dnephin (You reach a point in pairing assists where you're a co-author. We reached that point.)


cmd/build.go
```diff
@@ -77,6 +80,41 @@ func newLocalCommand() *cobra.Command {
...
+	var configWithVersion struct {
+		Version string "yaml:version"
+	}
+	if err := mapstructure.WeakDecode(raw, &configWithVersion); err != nil {
+		return "", false
+	}
+	return configWithVersion.Version, true
+}
+
+func validateArgs(cmd *cobra.Command, args []string) error {
```
dnephin: Arg validators only receive the non-flag args, not the full list of args, so the filename won't be available from here.

I think this function is generally the right idea, it would just change to accept a `filename string` as arguments and be called from `runExecute`

hannahhenderson:
Update: This is now working the way I think it should work from the CLI.

screen shot 2018-08-24 at 4 56 08 pm


The tests are clearly bad / not returning correctly.

cmd/build.go
```diff
@@ -154,7 +158,69 @@ func picardImage() (string, error) {
...
+		if v == v1 || v == v2 {
+			return i
+		}
+	}
+	return -1
+}
+
+func validateConfigVersion(args []string) error {
+	indexConfigFlag := containsIdx(args, "--config", "--c")
+	configPath := opts.configFilename
```
dnephin: I believe this line is not doing anything because  `DisableFlagParsing` is enabled.

cmd/build.go
```diff
@@ -154,7 +158,69 @@ func picardImage() (string, error) {
...
+
+	if indexConfigFlag > -1 && indexConfigFlag+1 < len(args) {
+		configPath = args[indexConfigFlag+1]
+	}
```
dnephin: I don't think we need to re-implement flag parsing. `spf13/pflag` (the flag library that is used by cobra) can be used to parse the `args`.

My suggestion would be:
* move `flags.StringVarP(&opts.configFilename, "config", "c", defaultConfigPath, "config file")` into an `addConfigFlag(opts proxyBuildArgs, flagset *pflag.FlagSet)`
* call `addConfigFlag` from `newLocalExecuteCommand()` to add it to the flag set used for documentation
* [create a new `pflag.FlagSet`](https://godoc.org/github.com/spf13/pflag#NewFlagSet). You can set ` ParseErrorsWhitelist.UnknownFlags = true` to ignore the other flags.
* create a new instance of `proxyBuildArgs{}` to hold the value from flags.
* call `addConfigFlag` from here (`validateConfigVersion`) to add the flag to the FlagSet
* call https://godoc.org/github.com/spf13/pflag#FlagSet.Parse on the new FlagSet, passing in `args`
* that will populate `opts.configFilename`

cmd/build.go
hannahhenderson: 🤦‍♀️

cmd/build.go
```diff
@@ -154,7 +163,56 @@ func picardImage() (string, error) {
...
+		Version string `yaml:"version"`
+	}
+	if err := mapstructure.WeakDecode(raw, &configWithVersion); err != nil {
+		return "", false
+	}
+	return configWithVersion.Version, true
+}
+
+func validateConfigVersion(args []string) error {
+	invalidConfigError := "\nYou attempted to run a local build with version '%s' of configuration.\nLocal builds do not support that version at this time.\nYou can use 'circleci config process' to pre-process your config into a version that local builds can run (see 'circleci help config process' for more information)"
```
dnephin: To make this string a bit more readable you could use golang string literals (https://golang.org/ref/spec#String_literals) to format it accross multiple strings. It's hard to show an example of this in github because github markdown uses the same character for inline code blocks, but here's a gist:  https://gist.github.com/dnephin/694496562771a68e34be9cab559ee598

PR #77, opened 2018-08-24
New publish commands for orbs

zzak:
## Create an orb

```
$ circleci-cli orb create zzak testing
 Orb created
```

## Publish the first release

```
$ echo "{version: 2.1}" | circleci-cli orb publish release - zzak testing 0.0.1
 Orb published 0.0.1
```

## Publish a dev release

```
$ echo "{version: 2.1}" | circleci-cli orb publish dev - zzak testing latest
 Orb published dev:latest
```

## Increment the minor version

```
$ echo "{version: 2.1}" | circleci-cli orb publish increment - zzak testing minor
  Orb zzak/testing bumped to 0.1.0
```

## Promote the latest tag to a minor release

```
$ circleci-cli orb publish promote zzak testing latest minor
  Orb promoted to 0.2.0
```


cmd/orb.go
```diff
@@ -288,6 +331,55 @@ func releaseOrb(cmd *cobra.Command, args []string) error {
...
+	if err := validateSegmentArg(args); err != nil {
+		return err
+	}
+
+	v2, err := api.IncrementOrb(context.Background(), Logger, args[0], args[1], args[2], args[3])
+	if err != nil {
+		return err
+	}
+
+	Logger.Infof("Orb %s bumped to %s\n", args[1]+"/"+args[2], v2)
```
marcomorain: can you do the string manipulation in the format string? `("Orb %s/%s bumped to %s\n", args[1], args[2], v2)`

No need to be creating extra strings here with `+`.

api/api.go
```diff
@@ -461,6 +511,90 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
...
 
+// IncrementOrb accepts an orb and segment to increment the orb.
+func IncrementOrb(ctx context.Context, logger *logger.Logger, configPath string, namespace string, orb string, segment string) (string, error) {
+	name := namespace + "/" + orb
+	id, err := getOrbID(ctx, logger, name)
+	if err != nil {
+		return "", err
+	}
+
+	ov, err := OrbVersion(ctx, logger, namespace, orb)
```
marcomorain: can you pull the logic to produce a new version out into a separate function

api/api.go
```diff
@@ -242,6 +255,55 @@ func OrbPublish(ctx context.Context, logger *logger.Logger,
 	return &response.PublishOrb.PublishOrbResponse, err
 }
 
+// OrbPublishID publishes a new version of an orb by id
+func OrbPublishID(ctx context.Context, logger *logger.Logger,
```
dnephin: `OrbPublishByID` ?  `OrbPublishID` sounds like it's fetching and returning an `ID`.

api/api.go
```diff
@@ -242,6 +255,55 @@ func OrbPublish(ctx context.Context, logger *logger.Logger,
...
+	configPath string, orbID string, orbVersion string) (*PublishOrbResponse, error) {
```
dnephin: Not directly related to this PR, but I see Logging as a cross-cutting concern, so I generally don't like passing it as an argument. You could pull one from the context (I can send you a link to an example) if you don't want to use a global instance.

api/api.go
```diff
@@ -461,6 +523,159 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
...
+		v2 = v.IncMinor()
+	case "patch":
+		v2 = v.IncPatch()
+	}
+
+	return v2.String(), nil
+}
+
+// IncrementOrb accepts an orb and segment to increment the orb.
+func IncrementOrb(ctx context.Context, logger *logger.Logger, configPath string, namespace string, orb string, segment string) (string, error) {
```
dnephin: IncrementOrb seems to be inconsistent with the other naming in this file. This should probably be `OrbIncrementVersion`.

I would use a  struct like `OrbIncrementVersionOptions` any time a function has more than 3 arguments.

api/api.go
```diff
@@ -461,6 +523,159 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
...
+	if err != nil {
+		return "", err
+	}
+
+	v2, err := incOrbVersion(v, segment)
+	if err != nil {
+		return "", err
+	}
+
+	response, err := OrbPublishID(ctx, logger, configPath, id, v2)
```
dnephin: I believe this is racy and could end up clobbering a previous version.

Is it possible to add the previous version to the query so that it will only update when the previous version matches?

api/api.go
```diff
@@ -461,6 +523,159 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
...
+		Orb struct {
+			Versions []struct {
+				Version string
+			}
+		}
+	}
+
+	query := `query($name: String!) {
+			    orb(name: $name) {
+			      versions(count: 1) {
```
dnephin: I don't know much about graphQL, but from this query I can't tell that it's fetching latest. I guess the api-server always sorts them by semvar?

api/api.go
```diff
@@ -461,6 +523,159 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
...
+		return "", response.ToError()
+	}
+
+	logger.Debug("Bumped %s#%s from %s by %s to %s\n.", name, id, v, segment, v2)
+
+	return v2, nil
+}
+
+// OrbVersion finds the latest published version of an orb and returns it.
+func OrbVersion(ctx context.Context, logger *logger.Logger, namespace string, orb string) (string, error) {
```
dnephin: The comment should probably clarify that it's the latest according to semvar, not just latest published date (assuming that is the case).

Maybe `OrbLatestVersion` ?

api/api.go
```diff
@@ -461,6 +523,159 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
...
+
+	if len(response.Orb.Versions) != 1 {
+		return "", fmt.Errorf("the %s orb has never published a revision", name)
+	}
+
+	return response.Orb.Versions[0].Version, nil
+}
+
+// PromoteOrb takes an orb and a development version and increments a semantic release with the given segment.
+func PromoteOrb(ctx context.Context, logger *logger.Logger, namespace string, orb string, label string, segment string) (*PromoteOrbResponse, error) {
```
dnephin: To be consistent with other names: `OrbPromote` ?

api/api.go
```diff
@@ -461,6 +523,159 @@ func CreateOrb(ctx context.Context, logger *logger.Logger, namespace string, nam
 	return orb, err
 }
 
+func incOrbVersion(orbVersion string, segment string) (string, error) {
```
dnephin: This function is not really related to the API.  Maybe add a TODO to move it to some other package?

cmd/orb.go
```diff
@@ -288,6 +330,70 @@ func releaseOrb(cmd *cobra.Command, args []string) error {
...
+	ctx := context.Background()
+	response, err := api.OrbPublish(ctx, Logger, args[0], args[1], args[2], devLabel(args[3]))
+
+	if err != nil {
+		return err
+	}
+
+	if len(response.Errors) > 0 {
+		return response.ToError()
+	}
```
dnephin: Returning errors in two different ways (err and response.Errors) is not a great interface. Could this be handled in the `api` layer and returned as `err` ?

api/api.go
zzak: I'm willing to accept that passing around a logger to every function is silly, and like to refactor that but think it's better to save that for another PR.

If you have any ideas or examples on how we should do that, I'm all ears! :bow:

api/api.go
zzak: This is indeed racy, as we're getting the latest version in a separate request before publishing a new one. If someone else where to publish a version in-between those requests... :boom: 

I'm unsure how to solve this without somehow modifying the API to perform the update atomically in a single request?

api/api.go
zzak: I'm doing the simple thing here, that `orb source` was doing.

Presumably that returns the last created revision, though this may not be the `latest` version per say.

For example, if a user has published "1.0.0" then creates a "0.9.8", whichever was created last gets picked afaict.

Without changing the return value from the server, my best guess at a better implementation of this is to fetch all versions and sort them using one of the semver libraries we have available. Though this may be a little expensive..?

api/api.go
zzak: The args -> options struct is a great pattern that I'd like to apply to a larger refactor when the time comes.

api/api.go
zzak: All of the APIs seem to use `verbNoun` for these mutations :thinking:

api/api.go
dnephin: This is what I am planning on using in some projects: https://gist.github.com/dnephin/4b8106cf5b09494d1caecdf8f74ce88e

api/api.go
dnephin: One way is to accept the current version as a parameter in the graphQL. The API server can perform an `UPDATE ... WHERE` using the existing version. If the version doesn't match (because some other process updated it), then the query will update 0 rows, and the API server can return an error.

PR #78, opened 2018-08-27
Also push a latest tag for the docker image

zzak:


.circleci/config.yml
```diff
@@ -96,9 +96,11 @@ jobs:
       - run:
           name: Build Docker Image
           command: |
-            docker build -t circleci/circleci-cli:0.1.$CIRCLE_BUILD_NUM .
             docker login -u $DOCKER_HUB_USERNAME -p $DOCKER_HUB_PASSWORD
+            docker build -t circleci/circleci-cli:0.1.$CIRCLE_BUILD_NUM .
             docker push     circleci/circleci-cli:0.1.$CIRCLE_BUILD_NUM
+            docker build -t circleci/circleci-cli:latest .
```
dnephin: `docker tag circleci/circleci-cli:0.1.$CIRCLE_BUILD_NUM circleci/circleci-cli:latest` ?

zzak: @dnephin can you also push both tags with a single command?

zzak: It seems there is an open PR to add this to `docker`, but not yet merged:
https://github.com/docker/cli/pull/458

So we'll just have to use 2 bash commands :man_shrugging:

PR #80, opened 2018-08-28
Add `switch` command for users who accidentally try upgrading twice

### LINKED ISSUE #79: Local CircleCI doesn't support 2.1 preview
I just downloaded circleci from the usual:

> curl https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh --fail --show-error | bash

It reports version:

```
$ circleci version
0.1.1568+255077c
```

Possible aside: I noted that `circleci switch` does nothing.

When I run something locally with the config-preview-sdk (i.e. `version: 2.1`) it errors with config parsing issues that indicate that it's not using the preview SDK e.g.

```
version: 2.1
...
jobs:
  deploy:
     executor:
        name: some-custom-executor
...      
```

```
* Error in config file: The schema/shape of the YAML is incorrect: json: cannot unmarshal object into Go struct field JobDescription.Executor of type string
```

I seem to be missing something, and guidance would be appreciated.

zzak:
This resolves #79. :pray: 

Added note about version `0.1.6`, because that was the first build we deployed to GH releases:
https://github.com/CircleCI-Public/circleci-cli/releases?after=v0.1.19

The old CLI was using `0.1.0` and never updated beyond that, so these docs reflect that status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,11 @@ You may first need to generate a CircleCI API Token from the [Personal API Token
 $ circleci setup
 ```
 
-If you are using this tool on `.com`. accept the provided default `CircleCI Host`.
+If you are using this tool on `circleci.com`. accept the provided default `CircleCI Host`.
 
 Server users will have to change the default value to your custom address (i.e. `circleci.my-org.com`).
+
+**Note**: Server does not yet support config processing and orbs, you will only be able to use `circleci local execute` (previously `circleci build`) for now.
 
 
 ## Validate A Build Config
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This project is the seed for CircleCI's new command-line application.
 
 ### Upgrade from existing CLI
 
-If you already have installed the `circleci` CLI previously, run the following commands:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -47,7 +46,9 @@ You may first need to generate a CircleCI API Token from the [Personal API Token
 $ circleci setup
 ```
 
-If you are using this tool on `.com`. accept the provided default `CircleCI API End Point`. If you are using it on Server, change the value to your Server address (i.e. `circleci.my-org.com`).
+If you are using this tool on `.com`. accept the provided default `CircleCI Host`.
```
ndintenfass: This needs to say `circleci.com` instead of `.com`

README.md
```diff
@@ -47,7 +46,9 @@ You may first need to generate a CircleCI API Token from the [Personal API Token
...
+
+Server users will have to change the default value to your custom address (i.e. `circleci.my-org.com`).
```
ndintenfass: This CLI will not actually work very well against the current production server, which creates some interesting challenges.

PR #81, opened 2018-08-28
Circle 12813 add namespace filter to orb list

eric-hu:
Tests pass and feature behaves as expected.  There's a bit of duplication that I'd like to refactor, but I don't want that to hold up merging this if it looks okay otherwise.  Refactors can come in a separate PR too.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -13,7 +13,6 @@ import (
 	"github.com/CircleCI-Public/circleci-cli/client"
 	"github.com/CircleCI-Public/circleci-cli/logger"
 	"github.com/Masterminds/semver"
-	"github.com/machinebox/graphql"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 	yaml "gopkg.in/yaml.v2"
@@ -753,7 +752,7 @@ func ListNamespaceOrbs(ctx context.Context, logger *logger.Logger, namespace str
 		}
 	}
 
-	request := graphql.NewRequest(`
+	query := `
 query namespaceOrbs ($namespace: String, $after: String!) {
 	registryNamespace(name: $namespace) {
 		name
@@ -775,22 +774,23 @@ query namespaceOrbs ($namespace: String, $after: String!) {
 		}
 	}
 }
-`)
+`
 
 	address, err := GraphQLServerAddress(EnvEndpointHost())
 	if err != nil {
 		return err
 	}
-	client := client.NewClient(address, logger)
+	graphQLclient := client.NewClient(address, logger)
 
 	var result namespaceOrbResponse
 	currentCursor := ""
 
 	for {
+		request := client.NewAuthorizedRequest(viper.GetString("token"), query)
 		request.Var("after", currentCursor)
 		request.Var("namespace", namespace)
-		err := client.Run(ctx, request, &result)
 
+		err := graphQLclient.Run(ctx, request, &result)
 		if err != nil {
 			return errors.Wrap(err, "GraphQL query failed")
 		}
--- a/api/api.go
+++ b/api/api.go
@@ -726,7 +726,7 @@ func OrbSource(ctx context.Context, logger *logger.Logger, namespace string, orb
 // ListNamespaceOrbs queries the API to find all orbs belonging to the given
 // namespace. Prints the orbs and their jobs and commands to the supplied
 // logger.
-func ListNamespaceOrbs(ctx context.Context, logger *logger.Logger, namespace string) error {
+func ListNamespaceOrbs(ctx context.Context, logger *logger.Logger, namespace string) ([]Orb, error) {
 	// Define a structure that matches the result of the GQL
 	// query, so that we can use mapstructure to convert from
 	// nested maps to a strongly typed struct.
@@ -775,10 +775,11 @@ query namespaceOrbs ($namespace: String, $after: String!) {
 	}
 }
 `
+	var orbs []Orb
 
 	address, err := GraphQLServerAddress(EnvEndpointHost())
 	if err != nil {
-		return err
+		return orbs, err
 	}
 	graphQLclient := client.NewClient(address, logger)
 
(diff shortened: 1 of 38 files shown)


cmd/orb.go
```diff
@@ -23,10 +24,14 @@ var orbAnnotations = map[string]string{
...
-		Short: "List orbs",
-		RunE:  listOrbs,
+		Use:         "list NAMESPACE",
+		Short:       "List orbs",
+		Args:        cobra.MaximumNArgs(1),
+		RunE:        listOrbs,
+		Annotations: make(map[string]string),
 	}
+	optionalOrbs := []string{orbAnnotations["NAMESPACE"], "(Optional)"}
+	listCommand.Annotations["NAMESPACE"] = strings.Join(optionalOrbs, " ")
```
dnephin: Doesn't this break the convention of using flags for optional args and positionals for required args?

If we want to keep this exception the convention for documenting optional positional args is on the `Use` line:

```
list [NAMESPACE]
```

As an FYI, for concating strings in golang this would be more obvious:

```go
orbAnnotations["NAMESPACE"] + " (Optional)"
```

although I would just use the `Use` line.

cmd/orb.go
```diff
@@ -239,6 +247,104 @@ query ListOrbs ($after: String!) {
 	return nil
 }
 
+func listNamespaceOrbs(namespace string) error {
```
dnephin: This seems like it belongs in the `api` package.

cmd/orb_test.go
```diff
@@ -611,6 +611,85 @@ var _ = Describe("Orb integration tests", func() {
...
+			})
+
+			It("makes a namespace query and requests all orbs on that namespace", func() {
+				By("setting up a mock server")
+
+				// These requests and responses are generated from production data,
+				// but using a 5-per-page limit instead of the 20 requested.
+				tmpBytes, err := ioutil.ReadFile(filepath.Join("testdata/gql_orb_list_with_namespace", "first_request.json"))
+				Expect(err).ShouldNot(HaveOccurred())
+				firstGqlRequest := string(tmpBytes)
```
dnephin: This test would probably be more readable with some additional test helpers.

Ex: using https://godoc.org/gotest.tools/golden#Get

```go
firstGqlRequest := golden.Get(GinkgoT(), filepath.FromSlash("gql_orb_list_with_namespace/first_request.json"))
```

`golden.Get()` handles the errors.

cmd/orb.go
eric-hu: Thanks for the suggestion! @zzak has been doing most of the CLI work and has a strong preference for optional arguments over flags. He called this out as part of this feature's when we were talking about the JIRA ticket, so I'm leaving this as is to maintain consistency.

eric-hu: Also thanks for the tip about string concatenation with `+` -- that's much easier to work with!

cmd/orb_test.go
eric-hu: `FromSlash` for Windows compatibility, very nice 👍

eric-hu:
Updated with Daniel's suggestions and to pass lint.

Moving `listNamespaceOrbs` to `api.go` required modifying `listOrbs` in `orb.go` since they were sharing a type. `listOrbs` is now the only function in `orb.go` that interacts with the GraphQL API, but I'm holding off on cleaning that up since we want to get orb functionality in front of customers sooner.

eric-hu:
Thanks for the initial review @dnephin , does this PR look better in its current state?

api/api.go
```diff
@@ -66,6 +69,51 @@ type CreateOrbResponse struct {
...
+	Jobs      map[string]struct{}
+	Executors map[string]struct{}
+}
+
+func addOrbElementsToBuffer(buf *bytes.Buffer, name string, elems map[string]struct{}) error {
+	var err error
+
+	if len(elems) > 0 {
+		_, err = buf.WriteString(fmt.Sprintf("  %s:\n", name))
+		if err != nil {
```
dnephin: It's safe to ignore the error from `bytes.Buffer.Write` methods. They can't actually error, they have an error return so that they fit the `io.Writer` interface.

With the error removed the FIXME/panic in the caller should be save to remove as well.

api/api.go
```diff
@@ -502,3 +550,102 @@ func OrbSource(ctx context.Context, logger *logger.Logger, namespace string, orb
...
+				logger.Infof("%s (%s)", edge.Node.Name, v.Version)
+
+				// Parse the orb source to print its commands, executors and jobs
+				var o Orb
+				err := yaml.Unmarshal([]byte(edge.Node.Versions[0].Source), &o)
+				if err != nil {
+					logger.Error(fmt.Sprintf("Corrupt Orb %s %s", edge.Node.Name, v.Version), err)
+					continue NamespaceOrbs
+				}
+				logger.Info(o.String())
```
dnephin: This function should probably return the `[]Orb` so that the `cmd` package can print them.

The `Orb` struct could have a `Name` and `Version` field as well.

cmd/orb.go
zzak: While this does break the convention for optional args, I think using the namespace filter will be common enough and don't see there being a use for another arg in it's position to warrant a flag.

We may want to support something like `orb list certified` which we can re-use the same positional arg in place of namespace to specify a category of orbs to search for.

api/api.go
```diff
@@ -88,41 +88,27 @@ type Orb struct {
...
-	}
-	err = addOrbElementsToBuffer(&buffer, "Jobs", orb.Jobs)
-	if err != nil {
-		panic(err)
-	}
-	err = addOrbElementsToBuffer(&buffer, "Executors", orb.Executors)
-	if err != nil {
-		panic(err)
-	}
+	addOrbElementsToBuffer(&buffer, "Commands", orb.Commands)
```
eric-hu: Nice find with Buffer.Write always returning nil errors.

PR #86, opened 2018-08-30
Pretty print JSON to fix issue w/ grep commands

### LINKED ISSUE #85: Install script does not work
Tried using the install script and it does not work, as is. I believe the issue is the due to the formatting of the `release.json` file that is downloaded. The JSON text is all on one line, which causes issues with the `grep` commands as they return the entire JSON object instead of individual lines.

I have updated the script to fix this issue and will create a PR today.

**Before:**
```
Finding latest release.
Downloading CircleCI ## Changelog\n\n1f1866a Merge pull request #84 from CircleCI-Public/use-latest-codecov-orb\n229aaf5 Use released 0.0.4 codecov-clojure orb\n\n
An error occured installing the tool.
```

**After:**
```
Finding latest release.
Downloading CircleCI v0.1.1789
```

mjcbsn22:
Fixes #85


eric-hu:
Thanks for the PR! Could you push another commit with `git commit --allow-empty -m "bump"`?

We had fork PRs disabled when you opened the PR, so tests didn't run initially. We just ran through some security checks and enabled them, so another push should get things going.

PR #90, opened 2018-08-31
diagnostic command should test API availability

zzak:
This command can now run an introspection query on the GraphQL API
to make sure that it's able to connect and get a proper response.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -154,13 +154,13 @@ func GraphQLServerAddress(endpoint string, host string) (string, error) {
 	// 1. Parse the endpoint
 	e, err := url.Parse(endpoint)
 	if err != nil {
-		return e.String(), errors.Wrapf(err, "Parsing endpoint '%s'", endpoint)
+		return "", errors.Wrapf(err, "Parsing endpoint '%s'", endpoint)
 	}
 
 	// 2. Parse the host
 	h, err := url.Parse(host)
 	if err != nil {
-		return h.String(), errors.Wrapf(err, "Parsing host '%s'", host)
+		return "", errors.Wrapf(err, "Parsing host '%s'", host)
 	}
 	if !h.IsAbs() {
 		return h.String(), fmt.Errorf("Host (%s) must be absolute URL, including scheme", host)
--- a/api/api.go
+++ b/api/api.go
@@ -26,6 +26,26 @@ type GQLResponseErrors struct {
 	}
 }
 
+// IntrospectionResponse matches the result from making an introspection query
+type IntrospectionResponse struct {
+	Schema struct {
+		MutationType struct {
+			Name string
+		}
+		QueryType struct {
+			Name string
+		}
+		Types []struct {
+			Description string
+			Fields      []struct {
+				Name string
+			}
+			Kind string
+			Name string
+		}
+	} `json:"__schema"`
+}
+
 // ConfigResponse is a structure that matches the result of the GQL
 // query, so that we can use mapstructure to convert from
 // nested maps to a strongly typed struct.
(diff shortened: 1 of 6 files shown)


api/api.go
```diff
@@ -810,3 +830,42 @@ query namespaceOrbs ($namespace: String, $after: String!) {
...
 }
+
+// IntrospectionQuery makes a query on the API asking for bits of the schema
+// This query isn't intended to get the entire schema, there are better tools for that.
+func IntrospectionQuery(ctx context.Context, logger *logger.Logger) (*IntrospectionResponse, error) {
+	var response struct {
+		IntrospectionResponse
+	}
+
+	query := `query IntrospectionQuery {
```
eric-hu: What do you think of moving queries out into their own `.graphql` files? It would make our implementation code more focused on what Go does with the response, and we could more directly reference these queries with `cat ... | circleci query -` if we want to interact with them in the terminal.

ndintenfass: We have talked about doing exactly that. I love the idea also because it would help this be the reference implementation of the API and be a great learning tool for our users.

cmd/diagnostic_test.go
```diff
@@ -90,16 +86,16 @@ token: mytoken
...
-				Eventually(session.Out).Should(gbytes.Say("GraphQL API address: https://example.com/graphql"))
+				Eventually(session.Out).Should(gbytes.Say(
+					fmt.Sprintf("GraphQL API address: %s", testServer.URL())))
 				Eventually(session.Out).Should(gbytes.Say("OK, got a token."))
 				Eventually(session).Should(gexec.Exit(0))
 			})
 		})
 
-		Describe("token, host, and old endpoint set in config file", func() {
+		Describe("old endpoint set in config file", func() {
```
eric-hu: I think this should be the opposite of what was deleted: "token, host set in config file"

cmd/diagnostic_test.go
```diff
@@ -111,69 +107,52 @@ token: mytoken
...
 				Eventually(session.Err.Contents()).Should(BeEmpty())
-				Eventually(session.Out).Should(gbytes.Say("GraphQL API address: https://example.com/graphql"))
+				Eventually(session.Out).Should(gbytes.Say(
+					fmt.Sprintf("GraphQL API address: %s", testServer.URL())))
 				Eventually(session.Out).Should(gbytes.Say("OK, got a token."))
 				Eventually(session).Should(gexec.Exit(0))
 			})
 		})
 
-		Describe("token, host, and new endpoint set in config file", func() {
```
eric-hu: The original version of this test is basically moved here, right? https://github.com/CircleCI-Public/circleci-cli/blob/32fd81e7cffa37fa16ae362351a46f0de91eb9a6/api/api_test.go#L8-L45

cmd/diagnostic_test.go
```diff
@@ -31,24 +31,23 @@ var _ = Describe("Diagnostic", func() {
...
 
-		appendPostHandler(testServer, "test", MockRequestResponse{
-			Status:   http.StatusOK,
-			Request:  query,
-			Response: mockResponse})
+		testServer.AppendHandlers(
+			ghttp.CombineHandlers(
+				ghttp.VerifyRequest("POST", "/"),
+				ghttp.VerifyContentType("application/json; charset=utf-8"),
+				ghttp.RespondWith(http.StatusOK, `{ "data": `+mockResponse+`}`),
```
eric-hu: 👍 nice

api/api.go
```diff
@@ -810,3 +830,42 @@ query namespaceOrbs ($namespace: String, $after: String!) {
...
+		    __schema {
+		      queryType { name }
+		      mutationType { name }
+		      types {
+		        ...FullType
+		      }
+		    }
+		  }
+
+		  fragment FullType on __Type {
```
eric-hu: Is defining this fragment in this query preferable to having it inline?

cmd/diagnostic_test.go
zzak: @eric-hu More or less, this test is covered here:

```
 	addr, _ = GraphQLServerAddress("graphql-unstable", "https://example.com") 
 	expected = "https://example.com/graphql-unstable" 
 	if addr != expected { 
 		t.Errorf("Expected %s, got %s", expected, addr) 
 	} 
```

cmd/diagnostic_test.go
zzak: The description here is probably confusing, but the behavior we're specifying is:

* Have anything in host file
* Endpoint receives a fully-qualified address (via `--endpoint` above)
* The `--endpoint` address is reported as the API address in the diagnostic output

eric-hu: Ahh, so the purpose of this test is to ensure the ordering is correct when both are set.  I'll add a small PR to capture this.

zzak: @eric-hu It's documented here:
https://github.com/CircleCI-Public/circleci-cli/blob/d9f59f7b1f79fe202cfc5af1d1a6c41193b9557f/api/api.go#L169-L176

As well as additional test written for the behavior here:
https://github.com/CircleCI-Public/circleci-cli/blob/d9f59f7b1f79fe202cfc5af1d1a6c41193b9557f/api/api_test.go#L28-L32

(Not including the test I added in ccbe2d9)

eric-hu: Thanks! I saw the `diagnostic_test.go` update from `ccbe2d9` and then updated my comment above since that seemed clear enough.  I'm guessing Github sent you an email notification for my original comment though?

PR #94, opened 2018-08-31
Hide user token on "circleci setup"

eric-hu:
I spent a limited amount of time looking into cleaning up the setup function since this particular file is already using an interface and there's a fake implementation for that implementation too. The setup function will be a lot nicer when the testing code is pulled out. I think that cleanup is better done after discussing with Zak since he's been thinking about how to restructure the CLI code as well.

The existing integration tests in `setup_test.go` still traverse the code in this PR, but I did a manual test with this branch to verify:

```bash
$ go run main.go setup
✔ A CircleCI token is already set. Do you want to change it: y
API token has been set.
✔ CircleCI Host: https://circleci.com
CircleCI host has been set.
Setup complete. Your configuration has been saved.
```

Current CLI behavior:
```bash
$ circleci setup
✔ A CircleCI token is already set. Do you want to change it: y
✔ CircleCI API Token: <redacted>
API token has been set.
✔ CircleCI Host: https://circleci.com
CircleCI host has been set.
Setup complete. Your configuration has been saved.
```

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -3,11 +3,13 @@ package cmd
 import (
 	"fmt"
 	"strings"
+	"syscall"
 
 	"github.com/manifoldco/promptui"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 var testing = false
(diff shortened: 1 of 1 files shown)


eric-hu:
Also it's worth calling out that `terminal.ReadPassword` looks like it's unix-specific under the hood. I'm open to suggestions that are more platform agnostic.

zzak:
@eric-hu I'd like to close this in favor of #95

PR #95, opened 2018-09-03
Use promptui Mask attribute for tokens

zzak:


eric-hu:
One thing I noticed testing this with `go run main.go setup` is that backspace sends a bell to my terminal. I get this with `circleci setup` as well, though, so it's not specific to this masking behavior.

PR #97, opened 2018-09-04
Start from 0.0.0 if no previous version is found through inc/promote

zzak:
This allows the following workflow for publishing orbs:

```
$ circleci orb create zzak test4
Orb created

$ echo "{version: 2.1}" | circleci orb publish dev - zzak test4 latest
Orb published dev:latest

$ circleci orb publish promote zzak test4 latest minor
Orb promoted to 0.1.0

$ circleci orb list zzak
zzak/test4 (0.1.0)
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 
 	"fmt"
@@ -557,7 +558,7 @@ func OrbIncrementVersion(ctx context.Context, logger *logger.Logger, configPath 
 		return nil, err
 	}
 
-	v, err := OrbLatestVersion(ctx, logger, namespace, orb)
+	v, err := OrbLatestVersionOrNew(ctx, logger, namespace, orb)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -604,14 +605,34 @@ func OrbLatestVersion(ctx context.Context, logger *logger.Logger, namespace stri
 	return response.Orb.Versions[0].Version, nil
 }
 
+// OrbLatestVersionOrNew wraps OrbLatestVersion and returns "0.0.0" if no version is found
+func OrbLatestVersionOrNew(ctx context.Context, logger *logger.Logger, namespace string, orb string) (string, error) {
+	v, err := OrbLatestVersion(ctx, logger, namespace, orb)
+	if err != nil {
+		matched, er := regexp.MatchString("has never published a revision", err.Error())
```
dnephin: Since we're expecting to handle this error case in a special way I think there should be some kind of error code to identify the error condition. String matching is brittle, primarily because it's not obvious from the API server code that the exact string text is meaningful. It's easy to change without realizing some consumer depends on the exact text.

I did a quick search for graphQL error codes, and the lucinia docs link to http://facebook.github.io/graphql/June2018/#sec-Errors

Can we setup some kind of error code in the API and handle it here? I guess right now `Errors` is a `[]string` but maybe we could make it a type?

PR #98, opened 2018-09-05
Always print Help when a flag error or args are invalid

zzak:
## Calling without required positional arguments

```
$ circleci orb publish dev
release a development version of an orb


Usage:
  circleci orb publish dev PATH NAMESPACE ORB LABEL [flags]

Args:
  PATH        The path to your orb (use "-" for STDIN)
  NAMESPACE   The namespace used for the orb (i.e. circleci)
  ORB         The name of your orb (i.e. rails)
  LABEL       Tag to use for this development version (i.e. "volatile")


Flags:
  -h, --help   help for dev

Global Flags:
      --host string    URL to your CircleCI host (default "https://circleci.com")
      --token string   your token for using CircleCI
      --verbose        Enable verbose logging.

Error: accepts 4 arg(s), received 0
exit status 255
```

## Calling with invalid flag

```
$ circleci orb publish dev --foo
release a development version of an orb


Usage:
  circleci orb publish dev PATH NAMESPACE ORB LABEL [flags]

Args:
  PATH        The path to your orb (use "-" for STDIN)
  NAMESPACE   The namespace used for the orb (i.e. circleci)
  ORB         The name of your orb (i.e. rails)
  LABEL       Tag to use for this development version (i.e. "volatile")


Flags:
  -h, --help   help for dev

Global Flags:
      --host string    URL to your CircleCI host (default "https://circleci.com")
      --token string   your token for using CircleCI
      --verbose        Enable verbose logging.

Error: unknown flag: --foo
exit status 255
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"path"
 
@@ -117,7 +118,41 @@ func MakeCommands() *cobra.Command {
 	// just the error message.
 	rootCmd.SilenceUsage = true
 
+	setFlagErrorFuncAndValidateArgs(rootCmd)
+
 	return rootCmd
+}
+
+func setFlagErrorFunc(cmd *cobra.Command, err error) error {
+	if e := cmd.Help(); e != nil {
+		return e
+	}
+	fmt.Println("")
+	return err
+}
+
+func setFlagErrorFuncAndValidateArgs(command *cobra.Command) {
+	visitAll(command, func(cmd *cobra.Command) {
+		cmd.SetFlagErrorFunc(setFlagErrorFunc)
+
+		if cmd.Args == nil {
+			return
+		}
+
+		cmdArgs := cmd.Args
+		cmd.Args = func(cccmd *cobra.Command, args []string) error {
+			if err := cmdArgs(cccmd, args); err != nil {
+				if e := cccmd.Help(); e != nil {
+					return e
+				}
+
+				fmt.Println("")
+				return err
+			}
+
+			return nil
+		}
+	})
 }
 
 func bindCobraFlagToViper(command *cobra.Command, flag string) {
@@ -147,4 +182,11 @@ func init() {
 
 func prepare() {
 	Logger = logger.NewLogger(viper.GetBool("verbose"))
+}
+
+func visitAll(root *cobra.Command, fn func(*cobra.Command)) {
+	for _, cmd := range root.Commands() {
+		visitAll(cmd, fn)
+	}
+	fn(root)
 }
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -117,9 +118,43 @@ func MakeCommands() *cobra.Command {
...
+	setFlagErrorFuncAndValidateArgs(rootCmd)
+
 	return rootCmd
 }
 
+func setFlagErrorFunc(cmd *cobra.Command, err error) error {
+	if e := cmd.Help(); e != nil {
+		return e
+	}
+	fmt.Println("")
```
marcomorain: ?

cmd/root.go
```diff
@@ -117,9 +118,43 @@ func MakeCommands() *cobra.Command {
...
+		}
+
+		cmdArgs := cmd.Args
+		cmd.Args = func(cccmd *cobra.Command, args []string) error {
+			if err := cmdArgs(cccmd, args); err != nil {
+				if e := cccmd.Help(); e != nil {
+					return e
+				}
+
+				fmt.Println("")
```
marcomorain: ?

cmd/root.go
```diff
@@ -117,9 +117,40 @@ func MakeCommands() *cobra.Command {
...
+
+		if cmd.Args == nil {
+			return
+		}
+
+		cmdArgs := cmd.Args
+		cmd.Args = func(cccmd *cobra.Command, args []string) error {
+			if err := cmdArgs(cccmd, args); err != nil {
+				if e := cccmd.Help(); e != nil {
+					return e
```
eric-hu: So if I understand this correctly, it's rewriting each command with a post-error hook, and returning the help instead of the error, if there is one. Wouldn't this discard potentially useful error messaging in `err`?

cmd/root.go
```diff
@@ -127,7 +126,6 @@ func setFlagErrorFunc(cmd *cobra.Command, err error) error {
 	if e := cmd.Help(); e != nil {
 		return e
 	}
-	fmt.Println("")
```
zzak: These were here for a reason, to separate the help text from the error message as a line-break.

cmd/root.go
zzak: If for some odd reason printing the `Help()` message returns an error we want to return that, but we could probably wrap those together.

cmd/root.go
zzak: These newlines are here to add a visual break between the help text and the error message.

Please see the PR description for an example :bow:

cmd/root.go
eric-hu: This is easily mistakable for a commit error, especially as time passes and our memory of this PR fades. A comment would communicate its intentional.

PR #109, opened 2018-09-11
CIRCLE-13449 Add --uncertified flag to orbs list which defaults to false

zzak:
This means we should only show certified orbs by default.


eric-hu:
This is ready for review.

Commit 7cc4ae6 cleans up a few tests to test the waters and see if we like this `JustBeforeEach` Ginkgo pattern. It can be reverted if tests seem less clear with it than without.

zzak:
@eric-hu I've rebased https://github.com/CircleCI-Public/circleci-cli/commit/7cc4ae66516b1ec94cd7dd2433b25b1f65a339c6 from this branch as I want to save refactor for a separate PR and keep this specific to the feature.

PR #117, opened 2018-09-20
Print out the full orb's ref after publishing

zzak:
```
Orb published my/orb@0.0.1

Orb published my/orb@dev:volatile

Orb my/orb@dev:volatile promoted to 0.1.0

Orb bar-ns/foo-orb created
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -632,10 +632,6 @@ func OrbIncrementVersion(ctx context.Context, logger *logger.Logger, configPath 
 		return nil, err
 	}
 
-	if len(response.Errors) > 0 {
-		return nil, response.ToError()
-	}
-
 	logger.Debug("Bumped %s/%s#%s from %s by %s to %s\n.", namespace, orb, id, v, segment, v2)
 
 	return response, nil
--- a/api/api.go
+++ b/api/api.go
@@ -60,23 +60,14 @@ type ConfigResponse struct {
 // The OrbPublishResponse type matches the data shape of the GQL response for
 // publishing an orb.
 type OrbPublishResponse struct {
-	Orb struct {
-		CreatedAt string
-		Version   string
-	}
-
+	Orb Orb
 	GQLResponseErrors
 }
 
 // The OrbPromoteResponse type matches the data shape of the GQL response for
 // promoting an orb.
 type OrbPromoteResponse struct {
-	Orb struct {
-		CreatedAt string
-		Version   string
-		Source    string
-	}
-
+	Orb Orb
 	GQLResponseErrors
 }
 
@@ -103,10 +94,7 @@ type WhoamiResponse struct {
 // CreateOrbResponse type matches the data shape of the GQL response for
 // creating an orb
 type CreateOrbResponse struct {
-	Orb struct {
-		ID string
-	}
-
+	Orb Orb
 	GQLResponseErrors
 }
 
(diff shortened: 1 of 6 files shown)


cmd/orb.go
```diff
@@ -217,7 +217,10 @@ func releaseOrb(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	Logger.Infof("Orb published %s", response.Orb.Version)
+	response.Orb.Namespace = args[1]
```
eric-hu: Would it make more sense for `api.OrbPublishByID` to return an `OrbVersion` instead of an `OrbPublishResponse`? Then these attribute assignments on `Orb` don't need to be duplicated in `orb.go`.

cmd/orb.go
```diff
@@ -278,7 +284,7 @@ func promoteOrb(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	Logger.Infof("Orb promoted to %s", response.Orb.Version)
+	Logger.Infof("Orb %s/%s@%s promoted to %s", args[0], args[1], devLabel(args[2]), response.Orb.Version)
```
eric-hu: I think all these string formatting changes in `orb.go` are a good opportunity to isolate API response details into `api.go` and make our data structures (Orb, OrbVersion) return just our domain concept details. In these cases, they could return an Orb reference and Orb Version reference.

cmd/orb_test.go
```diff
@@ -832,7 +832,7 @@ var _ = Describe("Orb integration tests", func() {
 				session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
 
 				Expect(err).ShouldNot(HaveOccurred())
-				Eventually(session.Out).Should(gbytes.Say("Orb promoted to 0.1.0"))
+				Eventually(session.Out).Should(gbytes.Say("Orb my/orb@dev:volatile promoted to 0.1.0"))
```
eric-hu: Nice, these are much clearer for the user.

cmd/testdata/gql_orb_list/pretty_json_output.json
```diff
@@ -2,6 +2,7 @@
   "orbs": [
     {
       "name": "circleci/codecov-clojure",
+      "version": "0.0.3",
```
zzak: This shows up because we're tagging the `HighestVersion` field in json as `"version"`, so that deserializing the JSON response from the API which is normally `orb { version }`.

Since we set this field when listing orbs, it also shows up when serializing the data back to JSON:
https://github.com/CircleCI-Public/circleci-cli/blob/b73cf303490208d69b0496954ceaea6dfbf9c659/api/api.go#L863

api/api.go
```diff
@@ -123,15 +123,16 @@ type OrbVersion struct {
...
+	ID        string `json:"-"`
+	Name      string `json:"name"`
+	Namespace string `json:"-"`
+	CreatedAt string `json:"-"`
+
+	Source string `json:"-"`
 	// Avoid "Version" since there is a "version" key in the orb source referring
 	// to the orb schema version
-	HighestVersion string              `json:"-"`
+	HighestVersion string              `json:"version"`
```
eric-hu: ✨

cmd/orb.go
```diff
@@ -236,10 +236,10 @@ func releaseOrb(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	response.Orb.Namespace = args[1]
-	response.Orb.Name = args[2]
+	response.Namespace = args[1]
+	response.Name = args[2]
 
-	Logger.Infof("Orb published %s", response.Orb.Ref())
```
eric-hu: I thought it was cleaner the way you had it before, since that was paving the way for `api.OrbPublishByID` to return an `*Orb` type instead of an `*OrbPublishRelease` type.

zzak: @eric-hu I'm happy to make further refactorings in a separate PR.

PR #120, opened 2018-09-21
[Circle-12785] Prefer 'in-band' graphql errors

bradylill:
Until we make a decision around where errors are going to live, for the users sake, create orb & namespace will prefer 'in-band' errors over native graphql errors.

### DIFF
--- a/cmd/namespace.go
+++ b/cmd/namespace.go
@@ -38,7 +38,7 @@ func createNamespace(cmd *cobra.Command, args []string) error {
 	response, err := api.CreateNamespace(ctx, Logger, args[0], args[2], strings.ToUpper(args[1]))
 
 	// Only fall back to native graphql errors if there are no in-band errors.
-	if err != nil && response == nil && len(response.Errors) == 0 {
+	if err != nil && (response == nil || len(response.Errors) == 0) {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/namespace.go
```diff
@@ -37,7 +37,8 @@ func createNamespace(cmd *cobra.Command, args []string) error {
 
 	response, err := api.CreateNamespace(ctx, Logger, args[0], args[2], strings.ToUpper(args[1]))
 
-	if err != nil {
+	// Only fall back to native graphql errors if there are no in-band errors.
+	if err != nil && response == nil && len(response.Errors) == 0 {
```
eric-hu:
```
response == nil && len(response.Errors) == 0
```
Should this be `||`?

cmd/orb.go
```diff
@@ -310,7 +310,8 @@ func createOrb(cmd *cobra.Command, args []string) error {
 
 	response, err := api.CreateOrb(ctx, Logger, args[0], args[1])
 
-	if err != nil {
+	// Only fall back to native graphql errors if there are no in-band errors.
+	if err != nil && response == nil && len(response.Errors) == 0 {
```
eric-hu: Same question re: response being nil and response.errors == 0

cmd/namespace.go
bradylill: `&&` should be correct.

I'm basically shorting out that statement so we don't trying and access `.Errors` on a `nil` `response`, `nil.Errors` is not a good thing :)

cmd/orb.go
bradylill: same response as above :)

cmd/namespace.go
bradylill: :facepalm: yep.. should be `!=`

PR #121, opened 2018-09-21
[CIRCLE-13698] Adding world readable warning to namespace, orbs, and versions create.

bradylill:
Let the user know any namespaces, orbs, or versions they create/publish are world readable.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -239,7 +239,8 @@ func releaseOrb(cmd *cobra.Command, args []string) error {
 	response.Namespace = args[1]
 	response.Name = args[2]
 
-	Logger.Infof("Orb published %s/%s@%s", args[1], args[2], response.HighestVersion)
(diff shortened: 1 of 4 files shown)


cmd/orb.go
```diff
@@ -319,8 +321,9 @@ func createOrb(cmd *cobra.Command, args []string) error {
 		return response.ToError()
 	}
 
-	Logger.Infof("Orb %s/%s created", args[0], args[1])
-	Logger.Infof("You can now register versions of %s/%s using `circleci orb publish`", args[0], args[1])
+	Logger.Infof("Orb `%s/%s` created.", args[0], args[1])
+	Logger.Info("Please not that any versions you publish of this orb are open orb and are world-readable.")
```
marcomorain: Please not

marcomorain: Are open orb.

bradylill: >< good catch

PR #123, opened 2018-09-22
Use orbVersionRef query when fetching source

zzak:


api/api.go
```diff
@@ -730,28 +730,42 @@ func OrbPromote(ctx context.Context, logger *logger.Logger, namespace string, or
 	return &response.PromoteOrb.OrbPromoteResponse.Orb, err
 }
 
+func orbVersionRef(orb string) string {
+	split := strings.Split(orb, "@")
+	if len(split) > 1 {
+		return strings.Join(split, "@")
+	}
+
+	// If no version was supplied, prepend @volatile to the reference
```
DanToml: The comment says prepend, but this is appending?

api/api.go
```diff
@@ -730,28 +730,42 @@ func OrbPromote(ctx context.Context, logger *logger.Logger, namespace string, or
 	return &response.PromoteOrb.OrbPromoteResponse.Orb, err
 }
 
+func orbVersionRef(orb string) string {
```
DanToml: This is a fairly simple function, but it might be worth adding a couple of test cases.

api/api.go
```diff
@@ -730,28 +730,42 @@ func OrbPromote(ctx context.Context, logger *logger.Logger, namespace string, or
...
+	split := strings.Split(orb, "@")
+	if len(split) > 1 {
+		return strings.Join(split, "@")
```
DanToml: is this not the same as `return orb`?

api/api.go
```diff
@@ -730,28 +730,42 @@ func OrbPromote(ctx context.Context, logger *logger.Logger, namespace string, or
 	return &response.PromoteOrb.OrbPromoteResponse.Orb, err
 }
 
+func orbVersionRef(orb string) string {
+	split := strings.Split(orb, "@")
+	if len(split) > 1 {
```
DanToml: Should it be an error if there is more than 1 `@`?

api/api.go
eric-hu: Can you add some tests for `orbVersionRef` to demonstrate the use cases?

cmd/orb.go
```diff
@@ -115,14 +115,15 @@ func newOrbCommand() *cobra.Command {
...
 		Short:       "Show the source of an orb",
 		RunE:        showSource,
-		Args:        cobra.ExactArgs(2),
+		Args:        cobra.MaximumNArgs(2),
 		Annotations: make(map[string]string),
 	}
-	sourceCommand.Annotations["NAMESPACE"] = orbAnnotations["NAMESPACE"]
-	sourceCommand.Annotations["ORB"] = orbAnnotations["ORB"]
+	sourceCommand.Annotations["ORB"] = "The name of your orb, including namespace (i.e. circleci/rails)"
+	sourceCommand.Example = `  circleci orb source circleci/python@0.1.4 # grab the source at version 0.1.4
```
eric-hu: Nice example 👍

api/api.go
```diff
@@ -730,28 +730,42 @@ func OrbPromote(ctx context.Context, logger *logger.Logger, namespace string, or
...
+	}
+
+	// If no version was supplied, prepend @volatile to the reference
+	return fmt.Sprintf("%s@%s", split[0], "volatile")
+}
+
 // OrbSource gets the source or an orb
-func OrbSource(ctx context.Context, logger *logger.Logger, namespace string, orb string) (string, error) {
-	name := namespace + "/" + orb
+func OrbSource(ctx context.Context, logger *logger.Logger, orb string) (string, error) {
```
eric-hu: `orb` would be more accurately named `orbRef`

PR #127, opened 2018-09-24
Use orb references for all commands

marcomorain:
Addresses:
CIRCLE-13835
CIRCLE-13836
CIRCLE-13837

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -13,10 +13,9 @@ import (
 )
 
 var orbAnnotations = map[string]string{
-	"PATH":      "The path to your orb (use \"-\" for STDIN)",
-	"NAMESPACE": "The namespace used for the orb (i.e. circleci)",
-	"ORB":       "The name of your orb (i.e. rails)",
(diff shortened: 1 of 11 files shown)


cmd/orb.go
```diff
@@ -56,45 +57,24 @@ func newOrbCommand() *cobra.Command {
...
-	releaseCommand.Annotations["NAMESPACE"] = orbAnnotations["NAMESPACE"]
-	releaseCommand.Annotations["ORB"] = orbAnnotations["ORB"]
-	releaseCommand.Annotations["SEMVER"] = "The semantic version used for this release (i.e. 0.3.6)"
-
-	devCommand := &cobra.Command{
-		Use:         "dev PATH NAMESPACE ORB LABEL",
-		Short:       "release a development version of an orb",
-		RunE:        devOrb,
-		Args:        cobra.ExactArgs(4),
+		Use:         "publish <path> <orb>",
```
marcomorain: using angle brackets like this is a more standard unix pattern. How do we feel about this?

marcomorain: If we like it, I'd like to change all uses of the UPPERCASE forms.

PR #131, opened 2018-09-25
Improve "orb process" help text

eric-hu:
The "config process" help text is similar, but I didn't update that since the
description would exceed what's reasonable for a short description string.
Cobra supports adding a long description string, but seems like something to
discuss before using.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -39,7 +39,7 @@ func newOrbCommand() *cobra.Command {
 
 	validateCommand := &cobra.Command{
 		Use:         "validate <path>",
-		Short:       "validate an orb.yml",
+		Short:       "Validate an orb.yml",
 		RunE:        validateOrb,
 		Args:        cobra.ExactArgs(1),
 		Annotations: make(map[string]string),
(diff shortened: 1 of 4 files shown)


cmd/orb.go
```diff
@@ -48,7 +48,7 @@ func newOrbCommand() *cobra.Command {
 
 	processCommand := &cobra.Command{
 		Use:         "process <path>",
-		Short:       "process an orb",
+		Short:       "Validates orb, then inlines other orb dependencies",
```
dnephin: This reads differently from other short help text. The others start with a lowercase and are in a different tense.

Maybe "validate an orb and inline its dependencies" ?

How does it inline dependencies? Looking at the source it seems equivalent to a validate. The only difference is that one prints the yaml, and the other doesn't. 

Why would someone use process instead of validate or source?

eric-hu: That's a good point about the tense. I'll adjust that in a follow up commit.

For the casing, the current subcommands are a mix of upper and lowercase:

```
$ circleci orb
Operate on orbs


Usage:
  circleci orb [command]

Available Commands:
  create      create an orb
  list        List orbs
  process     process an orb
  publish     publish a version of an orb
  source      Show the source of an orb
  validate    validate an orb.yml
```

At the top level, they're all capitalized. I can make them all consistent in this PR. It looks like most other commands and subcommands are capitalized in their short text.

PR #134, opened 2018-09-26
Fix release tag for homebrew and use MacOS executor to deploy

zzak:


zzak:
@CircleCI-Public/dx-internal: I'm merging this PR to create a new deploy and will follow up with a PR to test the deploy

PR #138, opened 2018-09-26
Improve error message when docker is not present or not started.

### LINKED ISSUE #130: Confusing error message when docker is not running
Without Docker running the `local execute` command fails with a rather unhelpful error.

```sh
$ circleci local execute
Downloading latest CircleCI build agent...
Error: Could not find picard image: failed to pull latest docker image: exit status 1
```

Starting Docker fixes it, but I had to read the source before I figured that out 😅

marcomorain:
Fixes #130

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -211,6 +211,22 @@ You can use 'circleci config process' to pre-process your config into a version 
 	return nil
 }
 
+func ensureDockerIsRunning() error {
+	dockerOnPath := exec.Command("command", "-v", "docker").Run() == nil
+
+	if !dockerOnPath {
+		return errors.New("could not find `docker` on the PATH; please ensure than docker is installed")
+	}
+
+	dockerRunning := exec.Command("docker", "ps").Run() == nil
+
+	if !dockerRunning {
+		return errors.New("docker is not running; please ensure that docker is running, and that `docker ps` succeeds")
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


marcomorain:
@dnephin do you have a minute to look at this issue? Fixes #130

cmd/build.go
```diff
@@ -223,6 +239,10 @@ func runExecute(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "Could not find pwd")
 	}
 
+	if err = ensureDockerIsRunning(); err != nil {
```
marcomorain: I might need to add this in more places 🤔

cmd/update.go
```diff
@@ -58,6 +58,11 @@ func updateBuildAgent(cmd *cobra.Command, args []string) error {
 
 // Still depends on a function in cmd/build.go
 func findLatestPicardSha() (string, error) {
+
+	if err := ensureDockerIsRunning(); err != nil {
```
dnephin: Maybe this could be called from a function that creates the `Command` so that it's a bit more obvious when it needs to be called?

```go
func DockerCommand(args ...string) (*exec.Cmd, error) {
    if err := ensureDockerIsRunning(); err != nil {
        return nil, err
    }
    return exec.Command("docker", args...), nil
}
```

cmd/build.go
zzak: Maybe to the proxy commands as well.

PR #140, opened 2018-09-29
Ensure consistency on output for promote/increment

zzak:
This makes it easy to grab the published version ref and copy/paste it into your project's config.


zzak:
@CircleCI-Public/dx-internal Merging as this is a text change

PR #143, opened 2018-10-02
[CIRCLE-13846 & 13847] Messaging & Unpublished orbs

bradylill:
- Listing orbs with a namespace will show any orbs that have not been published 
- Added example usage to promote and updated result wording.
- Added example usage to increment and update result wording.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1002,7 +1002,7 @@ query namespaceOrbs ($namespace: String, $after: String!) {
 					continue NamespaceOrbs
 				}
 			} else {
-				o.HighestVersion = "Unpublish"
+				o.HighestVersion = "Not published"
 				o.Versions = []OrbVersion{}
 			}
 			orbs.Orbs = append(orbs.Orbs, o)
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -1001,8 +1001,11 @@ query namespaceOrbs ($namespace: String, $after: String!) {
 					logger.Error(fmt.Sprintf("Corrupt Orb %s %s", edge.Node.Name, v.Version), err)
 					continue NamespaceOrbs
 				}
-				orbs.Orbs = append(orbs.Orbs, o)
+			} else {
+				o.HighestVersion = "Unpublish"
```
eric-hu: Nit: "Unpublish" would be clearer as "Not published" or "No published versions".

cmd/orb.go
```diff
@@ -71,7 +71,9 @@ Please note that at this time all orbs published to the registry are world-reada
 		Use:   "promote <orb> <segment>",
 		Short: "Promote a development version of an orb to a semantic release",
 		Long: `Promote a development version of an orb to a semantic release.
-Please note that at this time all orbs promoted within the registry are world-readable.`,
+Please note that at this time all orbs promoted within the registry are world-readable.
+
+Example: 'circleci orb publish promote foo/bar@dev:master major' => foo/bar@1.0.0`,
```
eric-hu: Nice 💯

api/api.go
bradylill: sure

PR #145, opened 2018-10-04
[CIRCLE-13803] Replace forked graphql client and improve error and response handling

zzak:
This PR is best viewed by commit, in particular 870de6c067928575320d937b2633bef8d8337723 and d637f12d5730f4344d6ea1b853da953d183e6bfb.


client/client.go
```diff
@@ -1,44 +1,135 @@
...
+// A Client is an HTTP client for our GraphQL endpoint.
+type Client struct {
+	endpoint   string
+	httpClient *http.Client
+	logger     *logger.Logger
+}
 
-	client := graphql.NewClient(endpoint)
+// RequestBody is used for serializing a request made up of a query and key/value pair of variables to send to the GraphQL server.
+type RequestBody struct {
```
bradylill: What do you think about merging this struct with the `Request` struct below and adding the json serialization to that?

client/client_test.go
```diff
@@ -0,0 +1,208 @@
...
+	client := NewClient(server.URL, logger.NewLogger(false))
+
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
+	defer cancel()
+
+	var responseData struct {
+		Data   map[string]interface{}
+		Errors []struct {
+			Message string
+		}
```
bradylill: 👍

client/client.go
```diff
@@ -1,44 +1,123 @@
...
+	}
+
+	res, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err := res.Body.Close()
+		if err != nil {
+			log.Fatal(err)
```
dnephin: This will `os.Exit()` if there is an error closing the body. That's probably not good. Closing the body is not really critical, especially in a CLI where the process is short lived. I would suggest either printing a warning  or ignoring it all together.

client/client.go
```diff
@@ -1,44 +1,123 @@
...
+		}
+	}()
+	var buf bytes.Buffer
+	if _, err := io.Copy(&buf, res.Body); err != nil {
+		return errors.Wrap(err, "reading body")
+	}
```
dnephin: Is the `io.Copy` necessary? I think we discussed this, I'm curious if you ran into any problems trying to remove it.

The `res.Body` is already a reader, so it can be passed to `NewDecoder` directly without buffering. If buffering is necessary for some reason https://godoc.org/bufio#NewReader would likely be a better option.

cmd/namespace.go
```diff
@@ -41,12 +41,12 @@ func createNamespace(cmd *cobra.Command, args []string) error {
 	response, err := api.CreateNamespace(ctx, Logger, namespaceName, args[2], strings.ToUpper(args[1]))
 
 	// Only fall back to native graphql errors if there are no in-band errors.
-	if err != nil && (response == nil || len(response.Errors) == 0) {
+	if err != nil && (response == nil || len(response.Data.CreateNamespace.Errors) == 0) {
```
dnephin: Aren't the errors in `response.Errors` ?

api/api.go
```diff
@@ -19,32 +19,51 @@ import (
...
+
+// Error turns a GQLErrorsCollection into an acceptable error string that can be printed to the user.
+func (errs GQLErrorsCollection) Error() string {
+	messages := []string{}
+
+	for i := range errs {
+		messages = append(messages, errs[i].Message)
 	}
+
+	return strings.Join(messages, ": ")
```
dnephin: Joining errors with a colon is a convention used to wrap an error with additional context as it passes through callers.

In this case it's being used to concat multiple errors, which seems very different. This could result in some very confusing errors. It might be better to join these errors with newlines or commas.

api/api.go
```diff
@@ -19,32 +19,51 @@ import (
...
+			Types []struct {
+				Description string
+				Fields      []struct {
+					Name string
+				}
+				Kind string
+				Name string
+			}
+		} `json:"__schema"`
+	}
```
dnephin: I believe this struct is missing an `Errors` field.

api/api.go
```diff
@@ -55,48 +74,198 @@ type ConfigResponse struct {
 	SourceYaml string
 	OutputYaml string
 
-	GQLResponseErrors
+	Errors GQLErrorsCollection
```
dnephin: This `Errors` field is in the wrong spot. It needs to be at the top level of the `BuildConfigResponse` struct, as a sibling to `Data` field.

api/api.go
```diff
@@ -55,48 +74,198 @@ type ConfigResponse struct {
...
 // The OrbPublishResponse type matches the data shape of the GQL response for
 // publishing an orb.
 type OrbPublishResponse struct {
-	Orb Orb
-	GQLResponseErrors
+	Data struct {
+		PublishOrb struct {
+			Orb Orb
+
+			Errors GQLErrorsCollection
```
dnephin: Same with this `Errors` field, and all the ones below.

api/api.go
```diff
@@ -55,48 +74,198 @@ type ConfigResponse struct {
...
+// OrbLatestVersionResponse wraps the GQL result of fetching an Orb and latest version
+type OrbLatestVersionResponse struct {
+	Data struct {
+		Orb struct {
+			Versions []struct {
+				Version string
+			}
+		}
+	}
+}
```
dnephin: missing errors

zzak:
@dnephin Thanks for your review! I've addressed most of these concerns in #150, so let's move the discussion there. In particular, some thoughts on in-band/out-of-band/runtime type errors being context specific.

PR #150, opened 2018-10-06
[CIRCLE-14155]:  Surface out-of-band GraphQL errors for `namespace create`

zzak:
This catches an error when trying to find an org that doesn't exist:

```
$ circleci namespace create asdf github asdf
Error: Must have member permission.: the organization 'asdf' under 'GITHUB' VCS-type does not exist. Did you misspell the organization or VCS?
exit status 255
```

We've also made sure the in-band errors for createNamespace mutation
are preferred over out-of-band errors and runtime errors.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -615,8 +615,14 @@ func createNamespaceWithOwnerID(ctx context.Context, logger *logger.Logger, name
 
 	err = graphQLclient.Run(ctx, request, &response)
 
-	if err != nil {
-		return nil, errors.Wrapf(err, "Unable to create namespace %s for ownerId %s", name, ownerID)
+	// Only fall back to native graphql errors if there are no in-band errors.
+	if err != nil && len(response.Data.CreateNamespace.Errors) == 0 {
+		return nil, err
+	}
+
+	// Only fall back to out-of-band errors if there are no in-band errors
+	if len(response.Errors) > 0 && len(response.Data.CreateNamespace.Errors) == 0 {
+		return nil, response.Errors
 	}
 
 	if len(response.Data.CreateNamespace.Errors) > 0 {
@@ -654,6 +660,10 @@ func getOrganization(ctx context.Context, logger *logger.Logger, organizationNam
 		return nil, errors.Wrap(err, fmt.Sprintf("Unable to find organization %s of vcs-type %s", organizationName, organizationVcs))
 	}
 
+	if len(response.Errors) > 0 {
+		return nil, response.Errors
+	}
+
 	return &response, nil
 }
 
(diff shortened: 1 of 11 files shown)


api/api.go
```diff
@@ -615,8 +615,14 @@ func createNamespaceWithOwnerID(ctx context.Context, logger *logger.Logger, name
 
 	err = graphQLclient.Run(ctx, request, &response)
 
-	if err != nil {
-		return nil, errors.Wrapf(err, "Unable to create namespace %s for ownerId %s", name, ownerID)
+	// Only fall back to native graphql errors if there are no in-band errors.
+	if err != nil && len(response.Data.CreateNamespace.Errors) == 0 {
```
dnephin: I think this type of comparison is much easier to understand when the condition list is in priority order. That way you don't need an `&&` in each case, and the cyclomatic complexity is much lower. It also removes the need for inline comments, which should not be unnecessary here. These are simple comparisons that can be made obvious by the structure of the code.

```go
switch {
case len(response.Data.CreateNamespace.Errors) > 0:
    return response.Data.CreateNamespace.Errors
case err != nil:
    return err
case len(response.Errors) > 0:
    return response.Errors
}
```

I believe this is the order that you have now, but I'm not sure why `err` has higher priority than `response.Errors`. I guess in practice it is unlikely that both of those are set.

api/api.go
```diff
@@ -615,8 +615,14 @@ func createNamespaceWithOwnerID(ctx context.Context, logger *logger.Logger, name
...
+		return nil, err
+	}
+
+	// Only fall back to out-of-band errors if there are no in-band errors
+	if len(response.Errors) > 0 && len(response.Data.CreateNamespace.Errors) == 0 {
+		return nil, response.Errors
```
dnephin: Having errors in 3 different place still seems like a very strange interface. A user having to be aware of errors from multiple places seems error prone. Why are there 2 different places to report errors in the graphQL response?

To reiterate my earlier argument, I think turning `response.Errors` into a typed `error` in `Run` would have the following advantages:
* no need to add `Errors` to every `Response` type (which is very error prone, and risks that errors are ignored)
* makes the API cleaner by consolidating all errors in a single place. Most callers will only need to check one variable (instead of 2-3). Any caller that does care about differentiating the error type can use  type checks

I can't think of any disadvantages to communicating all errors in the `error` response. Are there some I am missing?

zzak:
I'm merging this as-is in hopes that we can make a pass to return a single typed error from `client.Run` as described here later: https://github.com/CircleCI-Public/circleci-cli/pull/150#discussion_r223189386

Issue #153, opened 2018-10-10
Users should be able to specify target install directory


marcomorain:
When installing `install.sh`, the install location [is hard coded](https://github.com/CircleCI-Public/circleci-cli/blob/2743216fe2fb5cfedc28373aa8df8851c715d1eb/install.sh#L7) to `/usr/local/bin`. Users should be able to override this location.

Perhaps specifying an argument to the script, or by setting an env var before executing it.

bananaumai:
May i take this issue?

Do you think is it ok to use environment variable?

It will look like below.

```
$ DEST=/opt/bin && bash -c "$(curl -fSl https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh)"
```

marcomorain:
That sounds great @bananaumai - yes please. And using `DEST` is perfect 👌

bananaumai:
@marcomorain Thanx! I made PR.

PR #154, opened 2018-10-12
Made users can specify install target dir

bananaumai:
This PR is for https://github.com/CircleCI-Public/circleci-cli/issues/153

This pr makes users can specify install target dir though `DESTDIR`.

If you check, try below.

```
$ export DESTDIR=/path/to/install/ && bash -c "$(curl -fSl https://raw.githubusercontent.com/bananaumai/circleci-cli/change-install-dest/install.sh"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ This will install the CLI into the `/usr/local/bin` directory.
 
 If you do not have write permissions to `/usr/local/bin`, you may need to run the above command with `sudo`.
 
+Optionally, you can install in a arbitrary directory by using `DESTDIR` environment variable.
+
+```
+export DESTDIR=/path/to/install/ && bash -c "$(curl -fSl https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh)"
+```
+
 #### Homebrew
 
 ```
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@ set -o nounset
 set -o pipefail
 
 RELEASE_URL="https://api.github.com/repos/CircleCI-Public/circleci-cli/releases/latest"
-DEST="/usr/local/bin/circleci"
+DEST="${DESTDIR:-/usr/local/bin}/circleci"
 
 # Run the script in a temporary directory that we know is empty.
 SCRATCH=$(mktemp -d)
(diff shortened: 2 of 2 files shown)


install.sh
```diff
@@ -38,6 +38,6 @@ grep -i "$(uname)" tarball_urls.txt | xargs curl --retry 3 --fail --location --o
 tar zxvf circleci.tgz --strip 1
 
 echo "Installing to $DEST"
-mv circleci $DEST
-chmod +x $DEST
```
bananaumai: These changes are for fixing the shellcheck errors

README.md
```diff
@@ -29,6 +29,12 @@ This will install the CLI into the `/usr/local/bin` directory.
 
 If you do not have write permissions to `/usr/local/bin`, you may need to run the above command with `sudo`.
 
+Optionally, you can install in a arbitrary directory by using `DESTDIR` environment variable.
+
+```
+export DESTDIR=/path/to/install/ && bash -c "$(curl -fSl https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh)"
```
zzak: Does `DESTDIR=/path/to/install bash -c "$(...)"` not do the same thing?

I also wonder if we should make this a command-line flag instead of an environment variable.

marcomorain: See https://github.com/CircleCI-Public/circleci-cli/issues/153 for more context. I'm happy with an env var.

marcomorain: `DESTDIR=/path/to/install bash -c "$(...)"` should do it. Is `DESTDIR` a clear enough word?

There seems to be some prior art for this word [in `automake`]((https://www.gnu.org/software/automake/manual/html_node/DESTDIR.html).

bananaumai: Thanks for reviewing for both of two. 

@zzak I personally think it's enough to use env var, because we can do same thing by both way, env var and command-line flag, so I took easier way. (of courese, i think there's some trade off)
But, i don't mind to use command-line flag.
if you have some good reason to use command-line flag, give me some instruction.

@marcomorain 

I got it. I didn't know much about `automake`. 

I first tryed to use `DEST` actually, but `DEST` has been already used in install.sh (which is fully qualified path). My opinion is like below.

- When to use `DEST`
  - I think it's better to change `DEST` in install.sh to other name, like `EXEC_PATH`.
  - Other
- Other options: not to use DEST*. Some ideas are below.
  - DSTDIR
  - DIR
  - TRAGETDIR
  - EXECDIR

PR #155, opened 2018-10-12
Use Alpine for the Docker image

matfax:
I don't see any reason why a go compiler image is being used as base image for the cli, which is a large overhead.
Since the binary seems to be statically compiled already, Alpine seems to be more appropriate for me.
That change would also speed up the workflows of your orb-tools.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,6 @@
-FROM circleci/golang:1.10.3
+FROM alpine:3.8
 
 COPY ./dist/linux_amd64/circleci /usr/local/bin
+
+ENTRYPOINT ["circleci"]
+CMD ["--help"]


marcomorain:
Thanks for the PR. This is something that we want to merge. We are going to make sure that this change won’t break any existing usages.

zzak:
@matfax Thanks for your contribution?

We'd like to support an alpine image, but as an alternative.

Would you mind making a new file, called `Dockerfile.alpine` and make the change there?

Regarding the `ENTRYPOINT`, this won't work afaik but I'm happy to try it out and see.
https://circleci.com/docs/2.0/custom-images/#adding-an-entrypoint

PR #157, opened 2018-10-14
[CIRCLE-14157] Make better CLI error messages when there is no token / a bad token

zzak:


api/api.go
```diff
@@ -393,19 +357,15 @@ func loadYaml(path string) (string, error) {
 }
 
 // WhoamiQuery returns the result of querying the `/me` endpoint of the API
-func WhoamiQuery(ctx context.Context, logger *logger.Logger) (*WhoamiResponse, error) {
+func WhoamiQuery(ctx context.Context, cfg *settings.Config) (*WhoamiResponse, error) {
```
dnephin: I see a few (maybe all) of these `api` functions are accepting a `*settings.Config`, but they only use the client. Accepting a large structure or interface than is necessary for the function generally makes it harder to refactor later.

Could these functions accept just a `Client` ?

client/client.go
```diff
@@ -16,24 +16,49 @@ type Client struct {
...
+		Header:    make(map[string][]string),
+	}
+
+	request.Header.Set("Authorization", cl.token)
+	request.Header.Set("User-Agent", version.UserAgent())
+	return request, nil
+}
+
+// NewUnauthorizedRequest returns a new GraphQL request without any authorization header.
+func (cl *Client) NewUnauthorizedRequest(query string) (*Request, error) {
```
dnephin: Nice! I like these (`NewAuthorizedRequest`, `NewUnauthorizedRequest`) constructors.  It seems like they don't make use of `*Client` much. It might make sense to make these package-level functions instead of methods on `Client`, and pass `token` as an arg to `NewAuthorizedRequest`

In the future you could use the [function options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) to support setting variables and headers as well.

client/client.go
```diff
@@ -16,24 +16,49 @@ type Client struct {
...
+	request := &Request{
+		Query:     query,
+		Variables: make(map[string]interface{}),
+		Header:    make(map[string][]string),
+	}
+
+	request.Header.Set("User-Agent", version.UserAgent())
+	return request, nil
```
dnephin: This never returns an error, so I think you could probably remove the `error` return

cmd/build.go
```diff
@@ -17,6 +17,11 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+type buildOptions struct {
+	*settings.Config
```
dnephin: I always prefer to use a field (ex: `config *settings.Config`) over an embedded struct. In some cases it doesn't make a difference, but when accessing fields I find it is easier to understand where values come from when it is a field. 

I generally will only use embedded structs as a way to satisfy interfaces.

cmd/build.go
```diff
@@ -39,50 +44,63 @@ func addConfigFlag(filename *string, flagset *pflag.FlagSet) {
...
-		Use:                "execute",
-		Short:              "Run a job in a container on the local machine",
-		RunE:               runExecute,
+		Use:   "execute",
+		Short: "Run a job in a container on the local machine",
+		PreRun: func(cmd *cobra.Command, args []string) {
+			opts.args = args
+
+			if err := opts.Setup(); err != nil {
+				panic(err)
```
dnephin: I believe you can use `PreRunE` here instead of panic

settings/settings.go
```diff
@@ -1,11 +1,149 @@
...
+
+// Setup will initialize a logger and http client for the CLI
+func (cfg *Config) Setup() error {
+	cfg.Logger = cfg.NewLogger()
+	address, err := GraphQLServerAddress(cfg.Endpoint, cfg.Host)
+	if err != nil {
+		return err
+	}
+	cfg.Address = address
+	cfg.Client = cfg.NewClient()
```
dnephin: I think this `Setup` can be avoided by calling `cfg.NewClient()` when you need a client, instead of assigning it to a field on `Config`. The same can be done for `Logger`.

PR #158, opened 2018-10-14
feature(install): Easy DEST override.

cans:
This tiny commit makes it easy to override the installation
directory of the circleci cli tool.

That tool being a statically linked, easy to move, binary it
is a shame to require root privileges to install it when it
so easy to not to.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 set -o errexit
-set -o nounset
 set -o pipefail
 
 RELEASE_URL="https://api.github.com/repos/CircleCI-Public/circleci-cli/releases/latest"
-DEST="/usr/local/bin/circleci"
+DEST="${CIRCLECI_CLI_DEST:-/usr/local/bin/circleci}"
+
+set -o nounset
 
 # Run the script in a temporary directory that we know is empty.
 SCRATCH=$(mktemp -d)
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 set -o errexit
-set -o nounset
 set -o pipefail
 
 RELEASE_URL="https://api.github.com/repos/CircleCI-Public/circleci-cli/releases/latest"
-DEST="/usr/local/bin/circleci"
+DEST="${CIRCLECI_CLI_DEST:-/usr/local/bin/circleci}"
+
+set -o nounset
```
cans: Could be nicer to put all `set` commands back together here. But since `CIRCLECI_CLI_DEST` can purposedly be undefined this has to come after `DEST` defintion.

install.sh
```diff
@@ -38,6 +39,6 @@ grep -i "$(uname)" tarball_urls.txt | xargs curl --retry 3 --fail --location --o
 tar zxvf circleci.tgz --strip 1
 
 echo "Installing to $DEST"
-mv circleci $DEST
-chmod +x $DEST
-command -v circleci
+mv circleci "$DEST"
+chmod +x "$DEST"
```
cans: The two changes above are to make CI checks pass (see first, failed, PR build).

marcomorain:
Related to 
https://github.com/CircleCI-Public/circleci-cli/issues/153

marcomorain:
Thanks for the PR. I’ll take a look at this on Monday. 

I agree that requiring root is a shame. The current install script was written as a quick help to get something shipped as soon as we could. There are a lot of ways we can improve it.

install.sh
dnephin: Since you are setting a default value `nounset` will not fail on this line. It should be safe to move it back.

marcomorain:
LGTM, merged with some edits in https://github.com/CircleCI-Public/circleci-cli/commit/0f984537c02b9621221ee70266861e2abb016f24

PR #164, opened 2018-10-17
[CIRCLE-13833] When passing stdin to to `orb validate` the message is confusing

zzak:
Assuming the user can easily see the path they passed as an argument, re-printing it to the user can be confusing especially if its STDIN.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -255,7 +255,11 @@ func validateOrb(opts orbOptions) error {
 		return err
 	}
 
-	opts.log.Infof("Orb input is valid.")
+	if opts.args[0] == "-" {
+		opts.log.Infof("Orb input is valid.")
+	} else {
+		opts.log.Infof("Orb at `%s` is valid.", opts.args[0])
+	}
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


cmd/orb.go
```diff
@@ -255,7 +255,8 @@ func validateOrb(opts orbOptions) error {
 		return err
 	}
 
-	opts.log.Infof("Orb at `%s` is valid.", opts.args[0])
+	opts.log.Infof("Orb input is valid.")
```
ndintenfass: Seems a bit of a shame to actually lose the file info, especially if using this inside a build. Not a reason to hold this up, but pointing out that there was some usefulness to seeing the path in the output. If we kept it I guess we'd have to special-case "-".

marcomorain: Yeah, I would like top see `orb at <path> is valid` or `orb input is valid` here, depending on whether input is from a file or stdin.

marcomorain:
Does `config validate` have the same issue?

zzak:
@ndintenfass @marcomorain Thanks for your feedback! I've updated the behavior to match the following:

```
$ go run main.go config validate
Config file at .circleci/config.yml is valid.

$ cat .circleci/config.yml | go run main.go config validate -
Config input is valid.
```

ndintenfass:
This seems good from my end. Didn't actually run this state of the code, but everything seems to be fairly straight-forward and a clear improvement over the status quo.

PR #171, opened 2018-10-19
[CIRCLE-12843]: Update should be a single command

zzak:
This change incorporates the following:

* `circleci update` should install the latest CLI for the user
* Add a flag to `circleci update --check` that does the same thing as `circleci check`
* Hide `update check`, `update install`, and `update build-agent`
* Keep those commands for compatibility

## Tasks

- [x] Add a test for new behavior
- [x] Add compiler hook to skip update command when building
- [x] Send homebrew a PR to use that hook


ndintenfass:
What if update didn’t require confirmation ever? Can’t think of the reason I’d want that.

cmd/update.go
```diff
@@ -69,6 +86,13 @@ func newUpdateCommand(config *settings.Config) *cobra.Command {
 		},
 	})
 
+	update.PersistentFlags().BoolVar(&opts.noTTY, "no-tty", false, "Enable non-interactive mode for clients without a terminal")
+
```
dnephin: Since we're hiding `update check` maybe it makes sense to add a flag to do this and set `dryRun`? `update --check` ?

bear:
I would suggest not using `--no-tty` to suppress the prompt as I think that `tty` implies that the tool would not send *anything* to the output

Other tools in this area (specifically `apt`) use `--assume-yes` or `-y` or `-q` 
see http://manpages.ubuntu.com/manpages/cosmic/man8/apt-get.8.html

cmd/update.go
zzak: I've added `--check` in cf51a40.

zzak: I've added `--check` in cf51a40.

zzak: I've added `--check` in cf51a40.

PR #178, opened 2018-10-23
Ensure token is never printed in help text

zzak:
This was an oversight by me where trying to set the default value for a flag with the value from the config instance which is already loaded with user's config prior to flag parsing.

Instead just stick in a var and compare the value after parsing flags.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -18,6 +18,12 @@ var defaultHost = "https://circleci.com"
 // it should be set once when Execute is first called
 var rootCmd *cobra.Command
 
+// config is used internally for preparing CLI and passed to sub-commands
+var config *settings.Config
+
+// token stores the value passed in through the flag --token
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -18,6 +18,12 @@ var defaultHost = "https://circleci.com"
 // it should be set once when Execute is first called
 var rootCmd *cobra.Command
 
+// config is used internally for preparing CLI and passed to sub-commands
+var config *settings.Config
```
zzak: In hind-sight `config` is probably not a great name because it creates potential confusion with the name of the argument passed to individual commands.

I'd like to rename this to `rootOptions` in a follow-up PR.

zzak: https://github.com/CircleCI-Public/circleci-cli/pull/179

PR #179, opened 2018-10-23
Improve naming with rootOptions and rootTokenFromFlag

zzak:


zzak:
@CircleCI-Public/dx-internal Merging this, happy for post-review comments

PR #180, opened 2018-10-23
Allow execute to parse flags again.

### LINKED ISSUE #173: No help for `circleci local execute`
On version 0.1.3469 (the latest as of this posting):

```
$ circleci local execute --help
Usage of configFlags:
  -c, --config string   config file (default ".circleci/config.yml")
Error: Unable to parse flags
```

### LINKED ISSUE #174: local execute --volumes argument is not implemented
From `build.go`:

```
flags.StringSliceVarP(&args.taskInfo.Volumes, "volume", "v", nil, "Volume bind-mounting")
```

This setting doesn't seem to be read anywhere.

felicianotech:
Fixes #173
Fixes #174

This PR does two things.

1. Renabled flag parsing for the `circleci local execute` command.
2. Adds help text to show people that they can still add their Docker flags by appending a double dash (--) which is a Unix standard that Cobra supports.

With this PR, both will work:

` circleci local execute --help`
` circleci local execute -- -e INJECT_ENVAR=into-docker`

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -54,8 +54,12 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 	}
 
 	buildCommand := &cobra.Command{
-		Use:   "execute",
+		Use:   "execute [flags] [-- Docker specific flags]",
 		Short: "Run a job in a container on the local machine",
+		Long: `Execute will run a CircleCI job on the local machine via 'docker run'. 
+		Docker specific flags can be passed by appending '--' and then the flags. For 
+		example:
+			circleci local execute -- -e INJECTED_ENVAR=value`,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			opts.args = args
 			opts.log = logger.NewLogger(config.Debug)
@@ -64,7 +68,6 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 		RunE: func(_ *cobra.Command, _ []string) error {
 			return runExecute(opts)
 		},
-		DisableFlagParsing: true,
 	}
 
 	// Used as a convenience work-around when DisableFlagParsing is enabled


cmd/build.go
```diff
@@ -64,7 +68,6 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 		RunE: func(_ *cobra.Command, _ []string) error {
 			return runExecute(opts)
 		},
-		DisableFlagParsing: true,
```
dnephin: Disabling flags is necessary here. There's a short comment about this above the flags, but I guess we should add one above this line as well. The flags need to be handled by the process in the container, so we need them to be available from `args`. When flags are parsed by a command they are not provided from `args []string`.

We should add an inline comment about why this is disabled. We want these flags to exist on the command so that the `help` command works properly.

cmd/build.go
```diff
@@ -54,8 +54,12 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 	}
 
 	buildCommand := &cobra.Command{
-		Use:   "execute",
+		Use:   "execute [flags] [-- Docker specific flags]",
```
dnephin: I don't think calling these "docker specific" is correct. It should be the flags that are specified on this command.

felicianotech: But they are. Any flags here would be what's available with you run `docker run`. They are passed to the Docker CLI.

cmd/build.go
felicianotech: Did you run my PR? This was done because without it, any flags that the user would want to pass to Docker won't work because Cobra would try to interpret it as a Cobra flag for the CircleCI CLI. It's my understanding this is why this line (67) was added.

The Cobra flags can be separated from the Docker flags by a double hyphen. Anything before the double hyphen gets interpreted by Cobra as normal, and anything after gets passed to Docker, which was the reason for this line in the first place.

dnephin:
> Some users want to be able to pass their own specific flags to Docker as well. 

This is not something we should support. At worst extra flags will cause the command to fail. At best it's producing a different environment from what exists in cloud. I don't think this is what #174 is asking for. The `local execute` command itself has a `--volumes` flag.

The implementation of `local execute` may change in the future so that the flags are handled by this command. Attempting to support arbitrary docker flags now will lead users down a path that we may break in the future.

> without having to disable flag parsing which is what we do now.

This is not why flag parsing is disabled. The only reason flags were added here is to act as documentation. They are disabled because they are not used, and enabling them would hide them from the process in the container (which is what needs to see the flags).

I guess we did not document this well enough.

felicianotech:
So we don't want users to pass flags to Docker? Hmm okay. Closing this then.

PR #186, opened 2018-10-26
Show help for "local execute"

eric-hu:
We use `DisableFlagParsing: true` for `circleci local execute`, so add manual
checks of the user-provided flags for `-h` or `--help`.

Still needed:
[ ] Tests


eric-hu:
I still want to add tests for the behavior added, but submitted this as a PR to verify with @dnephin : is is this the approach you were talking about in the JIRA ticket?

cmd/build.go
```diff
@@ -61,7 +61,13 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 			opts.log = logger.NewLogger(config.Debug)
 			opts.cl = client.NewClient(config.Host, config.Endpoint, config.Token)
 		},
-		RunE: func(_ *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, flags []string) error {
+			for _, f := range flags {
+				if f == "--help" || f == "-h" {
```
dnephin: I think this might work, but wasn't exactly what I was thinking. 

This might be a little more resilient since it uses the same parsing mechanism as everything else:

```go
func(cmd *cobra.Command, args []string) error {
    if err := cmd.Flags().Parse(); err == pflag.ErrHelp {
        return cmd.Help()
    }
    return runExecute(opts)
}
```

I haven't tried this, but I believe it should work.

eric-hu: `Parse` expects one parameter `arguments []string`, so I had to add in `args` to get this to compile:

```go
func(cmd *cobra.Command, args []string) error {
	if err := cmd.Flags().Parse(args); err == pflag.ErrHelp {
		return cmd.Help()
	}
	return runExecute(opts)
},
```

This didn't appear to work, though:

```
$ go run main.go local execute --help
Usage of configFlags:
  -c, --config string   config file (default ".circleci/config.yml")
Error: Unable to parse flags
exit status 255
```

I walked through the above code with a debugger, and `err == pflag.ErrHelp` evaluates to false. Looking at the pflag code, it looks like the return value is set to `pflag.ErrHelp` when `-help` was set but not defined.

I'll write tests against the current code in this branch, but I'm still open to suggestions to how to get this working in a way that's consistent with the parsing used elsewhere!

eric-hu:
@dnephin how does the PR look now after these changes? I moved the help flag check into the beginning of `runExecute`, since that made testing a lot easier.

I also saw your note in the [JIRA ticket](https://circleci.atlassian.net/browse/CIRCLE-14447) about using the picard docker image for help, but I spoke to that in [this commit message](https://github.com/CircleCI-Public/circleci-cli/pull/186/commits/dc29d7fa8983403e13e823c7e0db050051524704).

PR #188, opened 2018-10-30
Switch CircleCI badge to shield style.

felicianotech:


zzak:
Thanks @felicianotech!

PR #194, opened 2018-11-06
[CIRCLE-13826]: Periodic update check for user

zzak:


update/update.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		return false, errors.Wrap(err, "error finding latest release")
+	}
+
+	return found, nil
+}
+
+// NewerVersionAvailable will tell us if the current version is the same as the latest version found from the GitHub releases API.
+func (options *Options) NewerVersionAvailable() bool {
```
dnephin: The name `NewerVersionAvailable` suggests to me that a true value would indicate that there are  updates available, however that is does not match the implementation.

Maybe `IsUpToDate` or `IsLatestVersion` ?

zzak: Went with `IsLatestVersion`, good catch :+1:

zzak:
Holding off on merging this until we add some tests to it.

zzak:
Holding off on merging this until we add some tests to it.

PR #201, opened 2018-11-14
Use the shellcheck orb

marcomorain:


zzak:
@marcomorain Looks like this failed because it's checking stuff in `./vendor`

marcomorain:
@zzak thanks for fixing my orb 💯 

updated with the new version now.

PR #204, opened 2018-11-16
Update README.md

antzshrek:


README.md
```diff
@@ -75,7 +75,9 @@ You may first need to generate a CircleCI API Token from the [Personal API Token
 ```
 $ circleci setup
 ```
-You should be prompted to enter the _CircleCI API Token_ you generated from Personal API Token tab](https://circleci.com/account/api)
+
+You should be prompted to enter the _CircleCI API Token_ you generated from the [Personal API Token tab](https://circleci.com/account/api)
+
```
zzak: Actually looks like this part about the "Personal API Token tab" is duplicated from the line above it here (line 74).

How about changing the first line in this section from:

> You may first need to generate a CircleCI API Token from the Personal API Token tab.

To:

> After installing the latest version of our CLI, you must run setup to configure the tool.

PR #216, opened 2018-11-28
Allow slashes in dev labels

johnswanson:
The server now allows the dev label to contain a `/` character, so the CLI
should allow it as well.


zzak:
Merged #214 instead

PR #229, opened 2018-12-14
Add slack notifications

zzak:


api/api.go
```diff
@@ -18,8 +18,6 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-// GQLErrorsCollection is a slice of errors returned by the GraphQL server.
-// Each error is made up of a GQLResponseError type.
```
yundt: Just making sure, as it's not connected to the main PR's topic: was this removal intentional?

.circleci/config.yml
```diff
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   codecov: circleci/codecov-clojure@0.0.4
   shellcheck: circleci/shellcheck@1.2.0
+  slack-orb: circleci/slack@dev:volatile
```
yundt: If we want to rely on those slack notifications in our workflow, maybe it would be safer to pin the version instead of using `volatile`?

Also, a nit: we don't use `-orb` for other aliases; we could change it to just `slack`, and then rename the `slack` job below to something like `slack-notify-on-failure`?

zzak: We're waiting on a release of the slack orb, that's backed up due to the master branch failing before we can change from `volatile`. (Actually the PR I wrote was never deployed because of this).

Regarding the naming, good idea; I will change this to what you suggested.

api/api.go
zzak: Correct, this was removed in 43a2846 so we could see an actual notification in slack

PR #230, opened 2018-12-17
[CIRCLE-14938]: Expose orb usage stats

zzak:
Couple of things to do, but this is ready for review:

- [x] Go back and rewrite f8b3f68 commit message
- [x] Add a test for sorting

_**edit**_: This is now ready for review :pray: :bow:

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -220,6 +220,12 @@ type OrbWithData struct {
 		Source  string `json:"source"`
 	} `json:"versions"`
 
+	UsageStats struct {
+		Last30DaysBuildCount        int
+		Last30DaysProjectCount      int
+		Last30DaysOrganizationCount int
+	} `json:"-"`
+
 	// These fields are printing manually when --details flag is added so hidden from JSON output.
 	Commands  map[string]OrbElement `json:"-"`
 	Jobs      map[string]OrbElement `json:"-"`
@@ -274,6 +280,12 @@ type Orb struct {
 
 	Source         string
 	HighestVersion string `json:"version"`
+
+	UsageStats struct {
+		Last30DaysBuildCount        int
+		Last30DaysProjectCount      int
+		Last30DaysOrganizationCount int
+	}
 
 	Commands  map[string]OrbElement
 	Jobs      map[string]OrbElement
@@ -858,6 +870,11 @@ func OrbInfo(opts Options, orbRef string) (*OrbVersion, error) {
                                     id
                                     createdAt
                                     name
+	                            usageStats {
+		                        last30DaysBuildCount,
+		                        last30DaysProjectCount,
+		                        last30DaysOrganizationCount
+	                            }
                                     versions {
                                         createdAt
                                         version
@@ -910,6 +927,11 @@ query ListOrbs ($after: String!, $certifiedOnly: Boolean!) {
 		cursor
 	  node {
 	    name
+	    usageStats {
+		last30DaysBuildCount,
+		last30DaysProjectCount,
+		last30DaysOrganizationCount
+	    }
 		  versions(count: 1) {
 			version,
 			source
@@ -921,7 +943,7 @@ query ListOrbs ($after: String!, $certifiedOnly: Boolean!) {
     }
   }
 }
-	`
+`
 
 	var orbs OrbsForListing
 
@@ -986,6 +1008,11 @@ query namespaceOrbs ($namespace: String, $after: String!) {
 						version
 					}
 					name
+	                                usageStats {
+		                           last30DaysBuildCount,
+		                           last30DaysProjectCount,
+		                           last30DaysOrganizationCount
+	                               }
 				}
 			}
 			totalCount
--- a/api/api.go
+++ b/api/api.go
@@ -220,7 +220,7 @@ type OrbWithData struct {
 		Source  string `json:"source"`
 	} `json:"versions"`
 
-	UsageStats struct {
+	Statistics struct {
 		Last30DaysBuildCount        int
 		Last30DaysProjectCount      int
 		Last30DaysOrganizationCount int
@@ -281,7 +281,7 @@ type Orb struct {
 	Source         string
 	HighestVersion string `json:"version"`
 
-	UsageStats struct {
+	Statistics struct {
 		Last30DaysBuildCount        int
 		Last30DaysProjectCount      int
 		Last30DaysOrganizationCount int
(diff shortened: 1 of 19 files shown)


cmd/orb.go
```diff
@@ -366,17 +369,38 @@ func logOrbs(orbCollection *api.OrbsForListing) error {
...
 
-	orbs, err := api.ListOrbs(opts.apiOpts, orbListUncertified)
+	orbs, err := api.ListOrbs(opts.apiOpts, opts.listUncertified)
 	if err != nil {
 		return errors.Wrapf(err, "Failed to list orbs")
 	}
 
-	return logOrbs(orbs)
+	if opts.sortBy != "" {
+		if err := validateSortFlag(opts.sortBy); err != nil {
```
eric-hu: Nit: The validation check would return a little faster if it ran before the API request `api.ListOrbs(...)`

cmd/orb.go
```diff
@@ -366,17 +369,38 @@ func logOrbs(orbCollection *api.OrbsForListing) error {
...
+var validSortFlag = map[string]bool{
+	"builds":   true,
+	"projects": true,
+	"orgs":     true}
+
+func validateSortFlag(sort string) error {
+	if _, valid := validSortFlag[sort]; valid {
+		return nil
+	}
+	// TODO(zzak): we could probably reuse the map above to print the valid values
```
eric-hu: There's a few approaches in this [SO thread](https://stackoverflow.com/a/36387643/640517), but I liked this one for its simplicity:

`keys := reflect.ValueOf(abc).MapKeys()`

api/api.go
```diff
@@ -211,19 +213,60 @@ type OrbsForListing struct {
...
+	case "orgs":
+		sort.Slice(orbs.Orbs, func(i, j int) bool {
+			return orbs.Orbs[i].Statistics.Last30DaysOrganizationCount > orbs.Orbs[j].Statistics.Last30DaysOrganizationCount
+		})
+	}
+}
+
+// OrbBase represents the minimum fields we wish to serialize for orbs.
+// This type can be embedded for extending orbs with more data. e.g. OrbWithData
+type OrbBase struct {
```
eric-hu: 👏 clear naming

cmd/build.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
```
eric-hu: Just curious about this change: did something require this import be explicitly named?

cmd/diagnostic_test.go
```diff
@@ -171,26 +198,114 @@ token: mytoken
...
+		    fields(includeDeprecated: true) {
+		      name
+		    }
+		  }`
+
+			request := client.NewUnauthorizedRequest(query)
+			expected, err := request.Encode()
+			Expect(err).ShouldNot(HaveOccurred())
+
+			tmpBytes, err := ioutil.ReadFile(filepath.Join("testdata/diagnostic", "response.json"))
```
eric-hu: Daniel pointed me to [golden.get](https://github.com/gotestyourself/gotest.tools/blob/b2aa35cc89847bde6179170bed441c491b3f0818/golden/golden.go#L25-L33) to read test files. It doesn't return an error and makes an assertion that the read was error-free.

cmd/diagnostic_test.go
```diff
@@ -171,26 +198,114 @@ token: mytoken
...
-				Eventually(session.Out).Should(gbytes.Say(
-					fmt.Sprintf("API host: %s", defaultHost)))
-				Eventually(session.Out).Should(gbytes.Say(
-					fmt.Sprintf("API endpoint: %s", testServer.URL())))
-				Eventually(session.Out).Should(gbytes.Say("OK, got a token."))
-				Eventually(session.Out).Should(gbytes.Say("Hello, zzak."))
-				Eventually(session).Should(gexec.Exit(0))
-			})
+		BeforeEach(func() {
+			var err error
```
eric-hu: Just a general comment, not specific to this PR: we might want this behavior for all of the CLI integration tests. I remember hitting a development issue with my CLI config file interfering with the behavior of the test suite, causing it to hit production.

cmd/build.go
zzak: This happened when `goimports` ran on one of my machines :man_shrugging:

cmd/diagnostic_test.go
zzak: This was a result of test re-use, but I think we can replace these `ioutil.ReadFile()` calls in one fell swoop. (On a separate PR). :pray:

cmd/orb.go
zzak: Let's fix this in a follow up PR :bow:

cmd/orb_test.go
```diff
@@ -1150,43 +1228,64 @@ var _ = Describe("Orb integration tests", func() {
 				)
 				By("setting up a mock server")
 
-				tmpBytes := golden.Get(GinkgoT(), filepath.FromSlash("gql_orb_list_uncertified/first_request.json"))
```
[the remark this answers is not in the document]
zzak: I have a plan to address this, for now rewriting the queries and calling the request builder is appropriate until we have a pattern for storing queries. This gist of this plan would be we'd only have to replace a single local variable with the designated query once it's available. So we can reuse the same query in our API code and tests, without polluting the repo with extra files or our tests with IO code.

cmd/diagnostic_test.go
eric-hu: (Clarification/Follow-up from Slack): It looks like this test setup is creating a temporary config file and configuring the CLI via `command.Env` to use that instead of the default config path. I think we may want this for all tests that make a GraphQL request.

I *think* the weird behavior I saw (months ago) was in tests without `command.Env`, a setting in my `~/.circleci/cli.yml` was overriding the CLI flag we use to set the mock server endpoint. It was either `hosts` or `endpoint` having a URL in my config. I think that usage is now deprecated, but this could be an issue for current CircleCI users who might want to contribute to the CLI.

zzak: Thank you for the clarification.

For tests that write to config, we use an isolated environment to ensure the user settings aren't affected. We use this pattern in a few other places that I thought were causing similar issues to the one you reported -- but am unaware if this is still a problem.

I agree that probably all tests should run in a clean environment, but that is probably a separate conversation we could have outside this PR.

PR #231, opened 2018-12-18
[CIRCLE-14881]: `orb list <namespace>` should report if ns not found

zzak:


cmd/orb_test.go
```diff
@@ -1686,6 +1686,75 @@ query namespaceOrbs ($namespace: String, $after: String!) {
...
+
+				appendPostHandler(testServer, "",
+					MockRequestResponse{
+						Status:   http.StatusOK,
+						Request:  encodedRequest.String(),
+						Response: mockResponse,
+					})
+			})
+
+			It("makes a namespace query and requests all orbs on that namespace", func() {
```
kimh: This needs a fix.

PR #239, opened 2018-12-25
[CIRCLE-15461] Show orb usage stats in `orb list --details`

zzak:


cmd/orb.go
```diff
@@ -308,13 +308,54 @@ func addOrbElementsToBuffer(buf *bytes.Buffer, name string, namedOrbElements map
...
+	}
+	sort.Strings(keys)
+
+	for _, key := range keys {
+		value := data[key]
+		_, err = buf.WriteString(fmt.Sprintf("    - %s: %d\n", key, value))
+	}
+
+	// This will never occur. The docs for bytes.Buffer.WriteString says err
+	// will always be nil. The linter still expects this error to be checked.
```
yundt: Is there a way to disable the linter for a particular case like this, instead of having to write a long comment and effectively 3 lines of dead code?

cmd/orb.go
```diff
@@ -308,13 +308,54 @@ func addOrbElementsToBuffer(buf *bytes.Buffer, name string, namedOrbElements map
...
+		panic(err)
+	}
+
+	if err = json.Unmarshal(encoded, &data); err != nil {
+		panic(err)
+	}
+
+	_, err = buf.WriteString(fmt.Sprintf("  %s:\n", name))
+
+	// Sort the keys so we always get the same results even after the round-trip
```
yundt: Could you explain why we need to marshal and unmarshal again? I don't get the comment that `we don't care about the fields`, or that we need to sort the keys later anyway.

cmd/orb.go
zzak: @yundt That's a fair point, we should be able to remove these lines with a single comment `// nolint: errcheck`

zzak: 02b2b93 and 4d209a3 address this.

cmd/orb.go
zzak: Usage stats that come in from the API we map to a type here:
https://github.com/CircleCI-Public/circleci-cli/blob/02b2b938a58829146055c22775169e98eb167a24/api/api.go#L248

However, we can't iterate over a struct so easily in Go -- so we round-trip it with JSON to turn it into a map structure. Rather than calling each field by hand, this makes it easy to enumerate the fields and print them -- so if we change the name or add new fields we don't have to update our formatting code each time.

This creates a nice separation between our formatting code and the API mapping logic.

yundt: Thanks for the explanation! I guess once we start doing this in other places, it'd be nice to extract this logic into a separate function, at least.

Out of curiosity I also just googled [`go struct to map`](https://www.google.com/search?q=go+struct+to+map&oq=go++struct+to+map&aqs=chrome.0.0l6.3748j0j7&sourceid=chrome&ie=UTF-8), and as expected we're not the only one with this problem, but couldn't find anything obviously better than this roundtrip.

zzak: [mapstructure](https://github.com/mitchellh/mapstructure) is another common library for this, I believe we've used elsewhere.

PR #246, opened 2018-12-29
Token validation should be command responsibility not HTTP client

zzak:
This change moves the token validation out of our GraphQL/HTTP client
and into the cmd package itself.

Commands can opt-in to validateToken pre-hook function.

Refactor `orb` command to use a single persistent pre-hook for setting
up the client and options.

Combined client.NewUnauthorizedRequest and client.NewAuthorizedRequest
to a single function with a request.SetToken helper.

## Motivation

The URL to create an API token may not always be the same, as is the case with Server installs.

Moving this logic out of the HTTP client gives us a path forward to solving that.

Happy to wait on merging this until after #245 .

---

7. Make sure your code lints (`make lint`). Tip: `make dev` will install `gometalinter`.


cmd/diagnostic_test.go
```diff
@@ -88,34 +81,13 @@ var _ = Describe("Diagnostic", func() {
...
-			)
-			Expect(err).ToNot(HaveOccurred())
-		})
-
 		Describe("token set in config file", func() {
 			BeforeEach(func() {
-				_, err := config.Write([]byte(`token: mytoken`))
-				Expect(err).ToNot(HaveOccurred())
-				Expect(config.Close()).To(Succeed())
+				tempSettings.writeToConfigAndClose([]byte(`token: mytoken`))
```
eric-hu: 👏 💯

data/data.go
```diff
@@ -0,0 +1,39 @@
+package data
+
+import (
+	"github.com/go-yaml/yaml"
+	packr "github.com/gobuffalo/packr/v2"
+)
+
+// YML maps the YAML found in _data/data.yml
+// Be sure to update this type when you modify the structure of that file!
+type YML struct {
```
eric-hu: Nit: The YML type is only related to yaml in that it uses yaml as a storage medium. Something like `StaticData` or `CircleCIInstallationData` would be more self-describing (though I don't love the names I came up with either).

cmd/query_test.go
```diff
@@ -16,36 +14,31 @@ import (
...
 		server = ghttp.NewServer()
 
-		var err error
-		tempHome, err = ioutil.TempDir("", "circleci-cli-test-")
-		Expect(err).ToNot(HaveOccurred())
+		tempSettings = withTempSettings()
 
 		token = "mytoken"
-		command = exec.Command(pathCLI, "query", "-",
+		command = commandWithHome(pathCLI, tempSettings.home,
```
eric-hu: I missed the PR when this was introduced, but still nice helper function. 👏

zzak: This was refactored here:
https://github.com/CircleCI-Public/circleci-cli/pull/248/commits/8937c9b09af515e5501cfdcc4c22381c2ec670bb

data/data.go
zzak: I can't think of a better name either, but went with `YML` because `DataYML` was redundant and failed the linter. Happy to change it later, it's doubtful anyone will depend on this tiny package outside of us (in spite it being public and exported type).

PR #247, opened 2018-12-31
Use packr to bundle static YAML for link data into the CLI

zzak:
This allows us to put all our URLs in a single place: `_data/data.yml`.

Probably the next use-case for this is to pull out GraphQL queries from embedded strings in the source code to external files with `.graphql` extension (or similar) for better editor support and reference usage.

---

7. Make sure your code lints (`make lint`). Tip: `make dev` will install `gometalinter`.


data/data.go
```diff
@@ -0,0 +1,39 @@
+package data
+
+import (
+	packr "github.com/gobuffalo/packr/v2"
+	yaml "gopkg.in/yaml.v2"
```
dnephin: `gopkg.in` uses fallbacks to determine the git repo for a given URL. There has been at least one case in the past where a user moved their github repo and someone else set up their own fork as the target for the `gopkg.in` url. To avoid this security problem I would suggest using `github.com/go-yaml/yaml` as the import url. Now that go modules exist there is no reason to use `gopkg.in` urls.

Issue #256, opened 2019-01-05
v0.1.5072+f930292: Bad deploy


zzak:
:speaker: I've deleted this release `v0.1.5072` as it contained a bug with how we package static files in the binary.

## How to know you're affected

If you run any `circleci` command you should see the following error:

```
panic: stat /_data/data.yml: no such file or directory
```

## Impacted versions

* `0.1.5017` was released about 3 hours before writing this.
* Docker image `circleci/circleci-cli:latest` and `circleci/circleci-cli:0.1.5072`

The tag and release has been removed from GitHub but docker tags haven't been changed.

Since we run a quick integration test on `-alpine` variant docker images, they were unaffected due to their deploy failing.

## How do I fix this issue

Once a new deploy is available I will close this issue with a note how to upgrade to the latest version.

For now, you can re-install `circleci` by downloading the appropriate binary on the GitHub releases page: https://github.com/CircleCI-Public/circleci-cli/releases/tag/v0.1.5060

For docker users, you can pin the image to the last known working version: `circleci/circleci-cli:0.1.5060` or the last known working SHA: `b854393e42e5e77d5776babc12b5239444af1b92e876b20a48f21f7c796880af`.

zzak:
This should be fixed by `v0.1.5124`:
https://github.com/CircleCI-Public/circleci-cli/releases/tag/v0.1.5124

Impacted users will have to reinstall the CLI through our script:

```
$ curl -fLSs https://circle.ci/cli | bash
```

Docker image users should be able to use `circleci/circleci-cli:0.1.5124` or `:latest`.

Here's the deploy job for detailed info:
https://circleci.com/gh/CircleCI-Public/circleci-cli/5124

PR #257, opened 2019-01-05
Ensure packr builds static assets before goreleaser runs

zzak:
This should take care of #256.


zzak:
@CircleCI-Public/dx-clients Merging this as it's blocking deploy on master.

PR #258, opened 2019-01-05
Extract common steps from deploy/deploy-test to ensure equality

zzak:
I had missed updating parts of `deploy` job which is a great reason to pull these into inline commands.

cc #256


zzak:
@CircleCI-Public/dx-clients Again, merging this to unblock deploy

PR #260, opened 2019-01-07
[CIRCLE-15472] Add confirm UI to orb create

zzak:
Since it's not possible to rename an orb, we want to ensure users are
aware of that fact and given a chance to prevent misnaming their orbs.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -11,6 +11,7 @@ import (
 	"github.com/CircleCI-Public/circleci-cli/client"
 	"github.com/CircleCI-Public/circleci-cli/references"
 	"github.com/CircleCI-Public/circleci-cli/settings"
+	"github.com/manifoldco/promptui"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


cmd/orb.go
```diff
@@ -567,15 +611,30 @@ func createOrb(opts orbOptions) error {
...
-	if err != nil {
-		return err
+You will not be able to change the name of this orb.
+
+If you change your mind about the name, you will have to create a new orb with the new name.
+
+`, namespace, orb)
+	}
+
+	confirm := fmt.Sprintf("Please confirm the name of this orb: `%s/%s`.", namespace, orb)
```
yundt: screen shot 2019-01-08 at 13 56 39
screen shot 2019-01-08 at 13 59 49

I'm not a fan of those `.?` and `.:` - I'd remove the `.`, maybe? Also, double `:` in the second screenshot seems a bit weird - maybe always keeping a question mark at the end would work, too?

PR #265, opened 2019-01-15
Document packr2 requirement

eric-hu:
Documenting something I fumbled with while trying to make a small PR earlier today.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -30,10 +30,13 @@ $ cd circleci-cli
 
 If you cloned the repo inside of your `$GOPATH`, you can use `GO111MODULE=on` in order to use Go modules. We recommend cloning the repo outside of `$GOPATH` as you would any other source code project, for example `~/code/circleci-cli`.
 
-The Makefile also assumes you have the `packr2` binary installed. You can install this with:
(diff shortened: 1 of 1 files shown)


zzak:
I would personally rather install [a release](https://github.com/gobuffalo/packr/releases/tag/v2.0.0-rc.14), even if it's in RC. The binary is only 3MB and using `go get` can take a very long time due to the number of dependencies they have -- also there's a chance it messes with your `go.mod` file :man_facepalming:

eric-hu:
Updated to point to the binary release. It threw me off that the `packr2` folder in that repo has both library and binary install instructions. Even the binary instructions in there use `go get`, which fetches a bunch of libraries as you pointed out.

zzak:
@eric-hu Also [2.0.0](https://github.com/gobuffalo/packr/releases/tag/v2.0.0) was finally released, so we can just use that.

eric-hu:
Yep, that's what the instructions are now pointing to. I only included macOS instructions since I wasn't sure what a good directory would be for unzipping to on Linux (or if the `tar` flags might be slightly different).

eric-hu:
Came here to close this after `make dev` took care of this problem. Thanks for taking care of this @zzak !

PR #267, opened 2019-01-27
[CIRCLE-15483]: Catch all 400s from GitHub release api in `update`

zzak:
We want to make sure we cover all bases and catch every 4** error
GitHub might return to us and display a verbose message for how to
treat these kinds of errors.

Since the majority of the the time this error occurs when the user
hits the API without setting a GITHUB_TOKEN, and results in a rate
limit for their IP address; these directions should help solve this problem.

### DIFF
--- a/cmd/update_test.go
+++ b/cmd/update_test.go
@@ -165,6 +165,10 @@ var _ = Describe("Update", func() {
 
 			// TODO: This should exit with error status 1, since 255 is a
 			// special error status for: "exit status outside of range".
+			//
+			// However this may be difficult to change, since all commands that return
+			// an error after executing cause the program to exit with a non-zero code:
+			// https://github.com/CircleCI-Public/circleci-cli/blob/5896baa95dad1b66f9c4a5b0a14571717c92aa55/cmd/root.go#L38
 			stderr := session.Wait().Err.Contents()
 			Expect(string(stderr)).To(ContainSubstring(`Error: Failed to query the GitHub API for updates.
 
@@ -173,13 +177,13 @@ This is most likely due to GitHub rate-limiting on unauthenticated requests.
 To have the circleci-cli make authenticated requests please:
 
   1. Generate a token at https://github.com/settings/tokens
-  2. Set the token by either adding it to your ~/.gitconfig
+  2. Set the token by either adding it to your ~/.gitconfig or
      setting the GITHUB_TOKEN environment variable.
 
 Instructions for generating a token can be found at:
 https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/
 
-We call the github releases API to look for new releases.
+We call the GitHub releases API to look for new releases.
 More information about that API can be found here: https://developer.github.com/v3/repos/releases/`))
 		})
 	})
--- a/update/update.go
+++ b/update/update.go
@@ -148,7 +148,7 @@ func latestRelease(opts *Options) error {
 		ghErr, ok := err.(*github.ErrorResponse)
 
 		if ok {
-			if ghErr.Response.StatusCode >= 400 || ghErr.Response.StatusCode < 500 {
+			if ghErr.Response.StatusCode >= 400 && ghErr.Response.StatusCode < 500 {
 				return errors.Wrap(err, `Failed to query the GitHub API for updates.
 
 This is most likely due to GitHub rate-limiting on unauthenticated requests.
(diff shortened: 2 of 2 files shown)


cmd/update_test.go
```diff
@@ -141,4 +141,46 @@ var _ = Describe("Update", func() {
 			Eventually(session).Should(gexec.Exit(0))
 		})
 	})
+
+	Describe("When Github returns a 403 error", func() {
+		BeforeEach(func() {
+			command = exec.Command(pathCLI,
+				"update", "check",
+				"--github-api", tempSettings.TestServer.URL(),
```
yundt: Just to make sure I understand this correctly - `--github-api` is a hidden flag to point the CLI to a test, fake instance of "GitHub API"?

cmd/update_test.go
```diff
@@ -141,4 +141,46 @@ var _ = Describe("Update", func() {
...
+			)
+		})
+
+		It("should print a helpful error message & exit 255", func() {
+			session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+			Expect(err).ShouldNot(HaveOccurred())
+			Eventually(session).Should(gexec.Exit(255))
+
+			// TODO: This should exit with error status 1, since 255 is a
+			// special error status for: "exit status outside of range".
```
yundt: What is stopping us from following through with this `TODO`? If there's a good reason to postpone it, it would be useful to put an explanation in the comment here as well.

cmd/update_test.go
```diff
@@ -141,4 +141,46 @@ var _ = Describe("Update", func() {
...
+			stderr := session.Wait().Err.Contents()
+			Expect(string(stderr)).To(ContainSubstring(`Error: Failed to query the GitHub API for updates.
+
+This is most likely due to GitHub rate-limiting on unauthenticated requests.
+
+To have the circleci-cli make authenticated requests please:
+
+  1. Generate a token at https://github.com/settings/tokens
+  2. Set the token by either adding it to your ~/.gitconfig
+     setting the GITHUB_TOKEN environment variable.
```
yundt: Missing an `or` in `adding it to your ~/.gitconfig [or] setting the GITHUB_TOKEN environment variable`

cmd/update_test.go
```diff
@@ -141,4 +141,46 @@ var _ = Describe("Update", func() {
...
+
+Instructions for generating a token can be found at:
+https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/
+
+We call the github releases API to look for new releases.
```
yundt: `github` -> `GitHub` to be consistent with previous usages in this message.

update/update.go
```diff
@@ -145,11 +145,34 @@ func latestRelease(opts *Options) error {
...
+Instructions for generating a token can be found at:
+https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/
+
+We call the github releases API to look for new releases.
+More information about that API can be found here: https://developer.github.com/v3/repos/releases/
+`)
+			}
+
+			if ghErr.Response.StatusCode == http.StatusUnauthorized {
+				return errors.Wrap(err, "Your Github token is invalid. Check the [github] section in ~/.gitconfig\n")
```
yundt: `Github` -> `GitHub`

cmd/update_test.go
zzak: Correct, `--github-api` is a global flag which maps to the `GitHubAPI` field of the rootOptions:
https://github.com/CircleCI-Public/circleci-cli/blob/5896baa95dad1b66f9c4a5b0a14571717c92aa55/cmd/root.go#L79

The flag is created here, but hidden from the user:
https://github.com/CircleCI-Public/circleci-cli/blob/5896baa95dad1b66f9c4a5b0a14571717c92aa55/cmd/root.go#L137

cmd/update_test.go
zzak: I extended the comment to add some more context, but the problem is essentially all commands that return an error get the same status code: https://github.com/CircleCI-Public/circleci-cli/blob/5896baa95dad1b66f9c4a5b0a14571717c92aa55/cmd/root.go#L38

update/update.go
```diff
@@ -145,11 +145,34 @@ func latestRelease(opts *Options) error {
 	opts.Found = found
 
 	if err != nil {
-		if errResponse, ok := err.(*github.ErrorResponse); ok && errResponse.Response.StatusCode == http.StatusUnauthorized {
-			return errors.Wrap(err, "Your Github token is invalid. Check the [github] section in ~/.gitconfig\n")
-		}
+		ghErr, ok := err.(*github.ErrorResponse)
+
+		if ok {
+			if ghErr.Response.StatusCode >= 400 || ghErr.Response.StatusCode < 500 {
```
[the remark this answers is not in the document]
zzak: Nice catch! :bow:

cmd/update_test.go
eric-hu: I'm just guessing here, but is `-1` syntactic sugar for 255? If so, I wonder if there's any time we actually want to return 255 instead of 1.

PR #268, opened 2019-01-30
Assume all errors returned from github release API are client related

zzak:
The code to cast err into a `github.ErrorResponse` seems to have never worked, therefor in the interest of shipping, let's catch all errors and display the rate-limit message as it's the most likely cause.

If anyone has any ideas how to make this code actually work, please let me know!

It was originally added in #88 but in testing doesn't appear to actually get executed. :pray:

## Steps to test

- Make sure `GITHUB_TOKEN` is unset
- `go run main.go update check` a bunch of times until your ip gets rate-limited

### DIFF
--- a/update/update.go
+++ b/update/update.go
@@ -3,14 +3,12 @@ package update
 import (
 	"encoding/json"
 	"fmt"
-	"net/http"
 	"os/exec"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


zzak:
@eric-hu Good call out, I've added steps to test to the PR description

eric-hu:
🙇

PR #272, opened 2019-02-11
[CIRCLE-15952]: Replace CI packr install with vendored binary

zzak:
We have to do this in the case of homebrew, installing go dependencies during their build isn't acceptable.

See also: Homebrew/homebrew-core#36546

### DIFF
--- a/.circleci/install-packr.sh
+++ b/.circleci/install-packr.sh
@@ -48,7 +48,7 @@ ARCH="$(get_arch_base)_$(get_arch_type)"
 
 PACKR_RELEASE_URL=$(grep -i "$ARCH" tarball_urls.txt)
 
-echo "Fetching packr for $ARCH at $PACKR_RELEASE_URL"
+echo "Fetching packr from $PACKR_RELEASE_URL"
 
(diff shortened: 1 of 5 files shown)


zzak:
@eric-hu Thanks for the call out about other OS, was not thinking :rofl:

This should be working now, please give it another shot!

.circleci/install-packr.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+RELEASE_URL="https://api.github.com/repos/gobuffalo/packr/releases/latest"
```
royvandewater: I wonder if we might want to lock this down to a particular version so the build will be deterministic on CI. I think the fact that we log out the version number used (it's in the curl output) is good enough for now, cause we can always go back and look at the most recent passing build if something starts failing.

zzak: @royvandewater That's a great point, after talking with @eric-hu we came up with fab85b2f5d36b0eae8431f3dc802d0a435d7ce5d

royvandewater: Sweet!

Issue #274, opened 2019-02-15
Fix install.sh on OSX <10.11


patcon:
- [x] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/master/CONTRIBUTING.md).
- [x] I have checked for similar issues and haven't found anything relevant.
- [x] This is not a security issue (which should be reported here: https://circleci.com/security/)

First off, thanks for maintaining this tool! :)

**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
`curl -fLSs https://circle.ci/cli | bash` fails on osx

Can you provide an example?
```
$ curl -fLSs https://circle.ci/cli | bash
Installing CircleCI CLI
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

**What is the expected behavior?**
Install should work.

This should fix it:
https://unix.stackexchange.com/questions/30091/fix-or-alternative-for-mktemp-in-os-x

**Which version of the CLI and OS are you using? Did this work in previous versions?**
osx 10.10 (newest cli)

Please provide the output of `circleci version` and `circleci diagnostic`.
```
# after install
$ circleci version
0.1.5389+4379bab
```

**If you have any questions, feel free to ping us at @CircleCI-Public/dx-clients.**

zzak:
Hello @patcon,

Thank you for your report.

While I don't have a mac to test this on, some other people here should.

In the mean time, you might want to try `brew install circleci` as an alternative.

Sorry for the trouble!

subuta:
In my case, `curl` command for getting latest release of Github failed with `API rate limit exceeded` error. but not shown at install command (`curl -fLSs https://circle.ci/cli | bash`) execution because of `--silent` option.

Fixed by adding `-H "Authorization: token ${your_gh_personal_access_token_goes_here}"` to curl command(at 25 of `install.sh`).

https://github.com/CircleCI-Public/circleci-cli/blob/master/install.sh#L25

eric-hu:
@patcon I was unable to reproduce this with macOS 10.12.6 (Sierra):

```
$ curl -fLSs https://circle.ci/cli | bash
Installing CircleCI CLI
Finding latest release.
Downloading CircleCI v0.1.5389
Installing to /usr/local/bin
/usr/local/bin/circleci
```

[...]

PR #277, opened 2019-02-28
Fix install script failure for macOS < 10.11

eric-hu:
MacOS versions prior to 10.11 included a version of mktemp that did not set
a default template. This should be compatible with both BSD and GNU
versions of mktemp.

Credit to @patcon for the solution

Addresses: https://github.com/CircleCI-Public/circleci-cli/issues/274


eric-hu:
Pinging @yundt or @royvandewater for a quick review since this is a 1 line change.

eric-hu:
This failed on deploy to Snap. Linux Snap users won't have the latest version, but will otherwise be unaffected. I don't think it's worth rolling back, as it seems like a case of expired credentials. Other merges to master will fail on this job too.

PR #279, opened 2019-03-05
Snap login lasts 1yr not 3.

felicianotech:
The Snapcraft Team got back to me. Even though I specified a 3 year refresh period, for the publish permission, we're still capped at 1yr. Updating the config comment/note.


eric-hu:
@felicianotech is this closed because the info in the original description isn't accurate anymore?

felicianotech:
Yeah I think I misunderstood. This PR shouldn't be needed.

PR #294, opened 2019-04-10
[CIRCLE-17397] Bugfix setup diagnostic check

iynere:
- [X] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/master/CONTRIBUTING.md).

## problem

[we've hardcoded a requirement that users have a "name" set in their VCS](https://github.com/CircleCI-Public/circleci-cli/blob/82d3083a4bdc5b845246453b2bf37dd2496dc60d/cmd/setup.go#L212), into the `setupDiagnosticCheck` function, which runs automatically at the end of `circleci setup` for all users

for anyone without a name in their VCS, this will result in a misleading error message suggesting that something is wrong with their token or CLI configuration or CircleCI account setup, when in reality everything's fine

## solution

- change error messaging from 'unable to query graphql api for your username' (not accurate, since it's actually looking for a user's _name_, as opposed to username) to a more general graphql api error message
- include specific response for condition in which token is valid but name is blank (just say 'hello' instead of, e.g., 'hello `<name>`')

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -211,7 +211,7 @@ func setupDiagnosticCheck(opts setupOptions) {
 
 	if err != nil {
 		fmt.Println("\nUnable to make a query against the GraphQL API, please check your settings.")
-	// If user does not have a name set in their VCS, return their login (VCS handle) instead
+		// If user does not have a name set in their VCS, return their login (VCS handle) instead
 	} else if responseWho.Me.Name == "" {
 		fmt.Printf("Hello, %s.\n", responseWho.Me.Login)
 	} else {
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -206,11 +206,11 @@ func setupDiagnosticCheck(opts setupOptions) {
 		fmt.Println("Ok.")
 	}
 
-	fmt.Printf("Trying to query your username given the provided token... ")
+	fmt.Printf("Trying to query our API for your profile name... ")
 	responseWho, err := api.WhoamiQuery(opts.cl)
 
 	if err != nil {
-		fmt.Println("\nUnable to make a query against the GraphQL API, please check your settings.")
+		fmt.Println("\nUnable to query our API for your profile name, please check your settings.")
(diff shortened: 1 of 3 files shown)


cmd/setup.go
```diff
@@ -209,8 +209,11 @@ func setupDiagnosticCheck(opts setupOptions) {
 	fmt.Printf("Trying to query your username given the provided token... ")
 	responseWho, err := api.WhoamiQuery(opts.cl)
 
-	if err != nil || responseWho.Me.Name == "" {
-		fmt.Println("\nUnable to query the GraphQL API for your username, please check your settings.")
```
zzak: Nit: this message is the same exact message as above:
https://github.com/CircleCI-Public/circleci-cli/pull/294/files#diff-0b2d181107099e093892167545faf375R204

Although they could fail for the exact same reasons.

cmd/setup.go
```diff
@@ -209,8 +209,11 @@ func setupDiagnosticCheck(opts setupOptions) {
 	fmt.Printf("Trying to query your username given the provided token... ")
```
zzak: Let's say "Trying to query our API for your profile name..." (or similar)

cmd/setup.go
```diff
@@ -209,8 +209,11 @@ func setupDiagnosticCheck(opts setupOptions) {
...
 	responseWho, err := api.WhoamiQuery(opts.cl)
 
-	if err != nil || responseWho.Me.Name == "" {
-		fmt.Println("\nUnable to query the GraphQL API for your username, please check your settings.")
+	if err != nil {
+		fmt.Println("\nUnable to make a query against the GraphQL API, please check your settings.")
+		// If user does not have a name set in their VCS, let's just say hi :)
```
zzak: :+1: :heart: Great comment!

cmd/setup.go
iynere: ya i copy & pasted, i thought it was good / general @zzak you think they should read differently?

PR #304, opened 2019-05-16
Use Print instead of Printf for printing config

johnswanson:
If the config contains things like `%m` then using `fmt.Printf` will render this as `%!m(MISSING)`.


eric-hu:
@johnswanson Was this closed after a conversation in Slack? Just wondering in what context `%m` came up.

PR #309, opened 2019-06-17
[CIRCLE-18591] fix `orb info` releases/versions output

iynere:

### problem

we don't currently specify a `count` value when we pull versions from GraphQL API, so we get the default response (most recent 10 releases). thus, we're outputting incorrect information, both for `Total-revisions` and `First-release`:

https://github.com/CircleCI-Public/circleci-cli/blob/master/cmd/orb.go#L646-L688

e.g., `circleci orb info circleci/orb-tools` outputs:

```
Latest: circleci/orb-tools@8.20.1
Last-updated: 2019-06-12T18:56:12.409Z
Created: 2018-09-24T07:55:19.386Z
First-release: 8.13.0 @ 2019-05-24T19:14:38.845Z
Total-revisions: 10
```

but `First-release` is wrong & `Total-revisions` is wrong

### changes

- bump `count` to 200 (arbitrarily large `n`) so we don't truncate our list of releases
- remove `First-release` entirely, since we already output `Created:`—this is inline with the UI/UX in the orb registry, & arguably first release date is not in itself particularly useful or meaningful

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -657,7 +657,7 @@ func orbInfo(opts orbOptions) error {
 		fmt.Printf("Latest: %s@%s\n", info.Orb.Name, info.Orb.HighestVersion)
 		fmt.Printf("Last-updated: %s\n", info.Orb.Versions[0].CreatedAt)
 		fmt.Printf("Created: %s\n", info.Orb.CreatedAt)
-		firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
+		// firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
 
 		fmt.Printf("Total-revisions: %d\n", len(info.Orb.Versions))
 	} else {
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -658,7 +658,6 @@ func orbInfo(opts orbOptions) error {
 		fmt.Printf("Last-updated: %s\n", info.Orb.Versions[0].CreatedAt)
 		fmt.Printf("Created: %s\n", info.Orb.CreatedAt)
 		firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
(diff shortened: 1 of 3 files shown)


cmd/orb.go
```diff
@@ -657,8 +657,7 @@ func orbInfo(opts orbOptions) error {
 		fmt.Printf("Latest: %s@%s\n", info.Orb.Name, info.Orb.HighestVersion)
 		fmt.Printf("Last-updated: %s\n", info.Orb.Versions[0].CreatedAt)
 		fmt.Printf("Created: %s\n", info.Orb.CreatedAt)
-		firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
-		fmt.Printf("First-release: %s @ %s\n", firstRelease.Version, firstRelease.CreatedAt)
+		// firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
 
 		fmt.Printf("Total-revisions: %d\n", len(info.Orb.Versions))
```
marcomorain: If I understand the bug correctly, then this line of output is also incorrect, right? The `len(info.Orb.Versions)` shows the number that were downloaded, not the numer in total?

cmd/orb.go
```diff
@@ -657,8 +657,7 @@ func orbInfo(opts orbOptions) error {
 		fmt.Printf("Latest: %s@%s\n", info.Orb.Name, info.Orb.HighestVersion)
 		fmt.Printf("Last-updated: %s\n", info.Orb.Versions[0].CreatedAt)
 		fmt.Printf("Created: %s\n", info.Orb.CreatedAt)
-		firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
-		fmt.Printf("First-release: %s @ %s\n", firstRelease.Version, firstRelease.CreatedAt)
+		// firstRelease := info.Orb.Versions[len(info.Orb.Versions)-1]
```
marcomorain: Please delete this line rather than commenting it out.

PR #320, opened 2019-08-12
Add orb diff command.

### LINKED ISSUE #312: [CIRCLE-14348] `circleci orb diff`

**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

lack of way to easily compare orb versions

Can you provide an example?

any orb

**What is the expected behavior?**

something like `circleci orb diff ORB-NAME SemVer1 SemVer2` that outputs code differences between orb versions

felicianotech:
Fixes #312.

Allows diff'ing the source between two different orb versions.

This PR doesn't contain a new test since I don't really understand what's going on in `cmd/orb_test.go`. If someone would like to help me with that or provide the test, that would be :+1:.


ndintenfass:
@felicianotech can you please post some screen shots of the output this produces?

felicianotech:
@ndintenfass Funny you wrote that. I'm trying to find a decent screen recorder for this very reason right now. I'll update in a bit.

felicianotech:
Updated OP with a screencast.

ndintenfass:
Personally, I don't think adding generic diff into the CLI makes a lot of sense - it's not very actionable. The original card talked about a semantic diff that would give you enough info to make decisions about semver level and such. As-is, this isn't much more valuable than using whatever diff tool you like, scripted with pulling two different orb sources via the CLI.

felicianotech:
@ndintenfass Your original card. Rose's card and Issue didn't.

Right now there's no way to see the actual difference. Yes, there's a way to string some commands to script this, but that logic can be applied to many of the CLI commands. The hope was to make it a bit easier.

I think your card is definitely a better approach, but I was stair stepping. Right now we have nothing.

zzak:
@felicianotech @ndintenfass I'm happy to merge this command as a WIP towards whatever goal you have in mind for this command -- but I think it should be marked as hidden until we flesh out the details.

ndintenfass:
Once the shape of this command launches you can’t just change it later. My point is that anyone who would know what to do with a diff like this wouldn’t be intimidated by some scripting. Generic string diff doesn’t feel like enough value add to occupy this command space to me.

PR #342, opened 2019-10-09
Added Meaning of some words to understand better.

MdHanif01:
Add meaning of some words which can help users understand better and easier.


gmemstr:
Given the widespread nature of linters in software dev, I don't think this is necessary to explain, at least not in our README. If anyone feels differently, feel free to lay out your case :)

PR #352, opened 2019-11-28
Make the install script a little bit safer when being piped

glenjamin:
- (ish, but minor) This is not a security issue (which should be reported here: https://circleci.com/security/)

If the connection cut out midway through, the script would partially execute.

Wrapping in a function ensures that nothing will be executed unless the file is fully downloaded

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,66 +10,72 @@
 
 set -o errexit
 
-echo "Starting installation."
-
-# GitHub's URL for the latest release, will redirect.
-GITHUB_BASE_URL="https://github.com/CircleCI-Public/circleci-cli"
-LATEST_URL="${GITHUB_BASE_URL}/releases/latest/"
-DESTDIR="${DESTDIR:-/usr/local/bin}"
-
-if [ -z "$VERSION" ]; then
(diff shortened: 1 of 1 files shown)


glenjamin:
This shouldn't make a difference to trap - I did a local test of making a command fail and it printed the error as before.

The main difference now is that when you run the script as `curl <url> | bash`, if the download aborts halfway through, the part of the script downloaded so far would still be passed to `bash`.

Wrapping in a function like this ensures that instead of executing half of the commands, we only define half of a function - nothing will be executed.

am-bui:
@glenjamin have you had a chance to consider @taxonomic-blackfish 's suggestions? We'd like to pick this pr back up but would also like to make sure the logic is up to date and applicable.

christian-stephen:
Perhaps this is platform/shell specific, but this seems to introduce a bug because of a name collision with the `install` function and the system's `install` command. When I pipe the script into `bash`, it loops indefinitely as the `install` function recursively calls itself:
```
Starting installation.
Installing CircleCI CLI v0.1.26646
Installing to /usr/local/bin
Starting installation.
Installing CircleCI CLI v0.1.26646
Installing to /usr/local/bin
Starting installation.
Installing CircleCI CLI v0.1.26646
^C⏎
```

PR #357, opened 2020-01-10
Delegate the entire 'tests' subcommand to circleci-agent

dnephin:
The `circleci tests` commands are intended to be run from inside a CircleCI build, so the official source for these commands is the agent.

We made some changes recently that were not reflected in the CLI because it reimplemented some of the commands.

This change removes the duplicate implementation and calls out to the agent.


go.mod
```diff
@@ -3,47 +3,37 @@ module github.com/CircleCI-Public/circleci-cli
 require (
```
dnephin: The changes in this file are a result of me running `go mod tidy` because I removed the doublestar dependency. It looks like `go mod tidy` is not being run in CI to verify it is updated on every change, so there were more changes than I expected.

Someone may want to add https://circleci.com/orbs/registry/orb/gotest/tools#commands-mod-tidy-check to the CI config

cmd/tests.go
```diff
@@ -1,95 +1,19 @@
...
-		Use:   "glob",
-		Short: "Glob files using pattern",
-		PreRun: func(cmd *cobra.Command, args []string) {
-			opts.args = args
-		},
-		RunE: func(_ *cobra.Command, _ []string) error {
-			return globRun(opts)
+		Use:                "tests",
+		Short:              "Collect and split tests so they may be run in parallel.",
+		DisableFlagParsing: true,
```
dnephin: This flag ensures that when `circleci tests --help` is called, the `--help` flag is passed to the agent instead of being handled by the CLI. That way the help text includes the subcommands.

dnephin:
The build required 2 fixes:
* Call `apt-get update` before trying to install the package, to update the package index. It was out of date, which is why it couldn't install things
* Update to use go1.12. `x/sys` no longer works with go1.11, and I couldn't find an appropriate older version to use. See https://github.com/golang/go/issues/34742

dnephin:
@lokst  Are you the right person to ping on this? I want to make sure there are no concerns about updating the Go version.

lokst:
@dnephin The X Team should be the right people to ping on this, as they are the codeowners

go.mod
```diff
@@ -3,47 +3,37 @@ module github.com/CircleCI-Public/circleci-cli
...
-	golang.org/x/sys v0.0.0-20181228144115-9a3f9b0469bb
-	golang.org/x/text v0.3.0
+	golang.org/x/sys v0.0.0-20181228144115-9a3f9b0469bb // indirect
 	golang.org/x/tools v0.0.0-20181221235234-d00ac6d27372 // indirect
-	google.golang.org/appengine v1.2.0
 	gopkg.in/yaml.v2 v2.2.2
 	gotest.tools v2.1.0+incompatible
 )
+
+go 1.13
```
zzak: Does it matter that we locked modules with 1.13 but using 1.12 in CI?

dnephin: Good catch! This is not exactly a lock, it's a "minimum required version" (I believe). In practice I haven't seen this do much, but for consistency I will change it to 1.12.

PR #362, opened 2020-02-07
[CIRCLE-24198] Add commands for managing contexts

marcomorain:
Add a new command, `context`:

```
$ circleci context
Contexts provide a mechanism for securing and sharing environment variables across projects. The environment variables are defined as name/value pairs and are injected at runtime.

Usage:
  circleci context [command]

Available Commands:
  list        List contexts
  remove      Remove a secret from the named context
  show        Show a context
  store       Store an new secret in the named context. The value is read from stdin.
```

Currently, the commands enable users to list contexts, show contexts, and to add and
remove variables from contexts.

Not yet supported / things to do:

- [x] write tests
- [x] create context
- [x] delete context
- [ ] CRUD on context groups

### DIFF
--- a/api/context.go
+++ b/api/context.go
@@ -1,0 +1,288 @@
+// Go functions that expose the Context-related calls in the GraphQL API.
+package api
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/CircleCI-Public/circleci-cli/client"
+	"github.com/pkg/errors"
+)
+
+type Resource struct {
+	Variable       string
+	CreatedAt      string
+	TruncatedValue string
+}
+
+type CircleCIContext struct {
+	ID        string
+	Name      string
+	CreatedAt string
+	Groups    struct {
+	}
+	Resources []Resource
+}
+
+type ContextsQueryResponse struct {
+	Organization struct {
+		Id       string
+		Contexts struct {
+			Edges []struct {
+				Node CircleCIContext
+			}
+		}
+	}
+}
+
+func improveVcsTypeError(err error) error {
+	if responseErrors, ok := err.(client.ResponseErrorsCollection); ok {
+		if len(responseErrors) > 0 {
+			details := responseErrors[0].Extensions
+			if details.EnumType == "VCSType" {
+				allowedValues := strings.ToLower(strings.Join(details.AllowedValues[:], ", "))
+				return fmt.Errorf("Invalid vcs-type '%s' provided, expected one of %s", strings.ToLower(details.Value), allowedValues)
+			}
+		}
+	}
+	return err
+}
+
+func CreateContext(cl *client.Client, vcsType, orgName, contextName string) error {
+
+	org, err := getOrganization(cl, orgName, vcsType)
+
+	if err != nil {
+		return err
+	}
+
+	query := `
+	mutation CreateContext($input: CreateContextInput!) {
+		createContext(input: $input) {
+		  ...CreateButton
+		}
+	  }
+
+	  fragment CreateButton on CreateContextPayload {
+		error {
+		  type
+		}
+	  }
+
+	`
+
+	var input struct {
+		OwnerId     string `json:"ownerId"`
+		OwnerType   string `json:"ownerType"`
+		ContextName string `json:"contextName"`
+	}
+
+	input.OwnerId = org.Organization.ID
+	input.OwnerType = "ORGANIZATION"
+	input.ContextName = contextName
+
+	request := client.NewRequest(query)
+	request.SetToken(cl.Token)
+	request.Var("input", input)
+
+	var response struct {
+		CreateContext struct {
+			Error struct {
+				Type string
+			}
+		}
+	}
+
+	if err = cl.Run(request, &response); err != nil {
+		return improveVcsTypeError(err)
+	}
+
+	if response.CreateContext.Error.Type != "" {
+		return fmt.Errorf("Error creating context: %s", response.CreateContext.Error.Type)
+	}
+
+	return nil
+}
+
+func ListContexts(cl *client.Client, orgName, vcsType string) (*ContextsQueryResponse, error) {
+
+	query := `
+	query ContextsQuery($orgName: String!, $vcsType: VCSType!) {
+		organization(name: $orgName, vcsType: $vcsType) {
+			id
+			contexts {
+				edges {
+					node {
+						...Context
+					}
+				}
+			}
+		}
+	}
+
+	fragment Context on Context {
+		id
+		name
+		createdAt
+		groups {
+			edges {
+				node {
+					...SecurityGroups
+				}
+			}
+		}
+		resources {
+			...EnvVars
+		}
+	}
+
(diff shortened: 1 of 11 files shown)


cmd/context.go
```diff
@@ -0,0 +1,225 @@
...
+	deleteContextCommand := &cobra.Command{
+		Short: "Delete the named context",
+		Use:   "delete <context name>",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return deleteContext(client, organization, args[0])
+		},
+		Args: cobra.ExactArgs(1),
+	}
+
+	command.PersistentFlags().StringVar(&organization.name, "organization", organization.name, "The organization to operate on.")
```
aengelberg: How does `organization.provider` get set? Or does it not matter for the contexts api?

cmd/context.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	removeCommand := &cobra.Command{
+		Short: "Remove a secret from the named context",
+		Use:   "remove <context name> <variable name>",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return removeEnvVar(client, organization, args[0], args[1])
+		},
+		Args: cobra.ExactArgs(2),
+	}
```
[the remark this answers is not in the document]
marcomorain: Yes, happy to change the names to be more clear.

cmd/context.go
```diff
@@ -0,0 +1,225 @@
...
+	for _, context := range contexts.Organization.Contexts.Edges {
+
+		table.Append([]string{
+			string(org.provider),
+			org.name,
+			context.Node.Name,
+			context.Node.CreatedAt,
+		})
+	}
+	table.Render()
```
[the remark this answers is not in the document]
marcomorain: Yes, I think that is valuable. I'll leave that for a second pass through this feature.

git/git.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type VcsType string
+
+const (
+	GitHub    VcsType = "GITHUB"
+	BitBucket VcsType = "BITBUCKET"
+)
+
+func InferOrganizationFromGitRemotes() (VcsType, string, error) {
```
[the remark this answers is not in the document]
marcomorain: I've decided to make the code explicit, matching the pattern from `circleci namespace create`, and I've removed this `git.go` file.

client/client.go
```diff
@@ -22,13 +22,13 @@ type Client struct {
 	Endpoint   string
 	Host       string
 	Token      string
-	httpClient *http.Client
+	HttpClient *http.Client
```
glenjamin: You could use the address from the server rather making this public, but this seems fine too

cmd/context.go
```diff
@@ -0,0 +1,239 @@
...
+func newContextCommand(config *settings.Config) *cobra.Command {
+
+	var cl *client.Client
+
+	initClient := func(cmd *cobra.Command, args []string) {
+		cl = client.NewClient(config.Host, config.Endpoint, config.Token, config.Debug)
+	}
+
+	command := &cobra.Command{
+		Use:   "context",
```
glenjamin: Worth marking this hidden until we’ve done a few iterations so we can release it quietly early?

cmd/context.go
```diff
@@ -0,0 +1,239 @@
...
+		PreRun: initClient,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return showContext(cl, args[0], args[1], args[2])
+		},
+		Args: cobra.ExactArgs(3),
+	}
+
+	storeCommand := &cobra.Command{
+		Short:  "Store an new secret in the named context. The value is read from stdin.",
+		Use:    "store-secret <vcs-type> <org-name> <context-name> <secret name>",
```
glenjamin: Do we call them secrets elsewhere? I thought we normally called them values

cmd/context.go
```diff
@@ -0,0 +1,239 @@
...
+		})
+	}
+	table.Render()
+
+	return nil
+}
+
+func contextByName(client *client.Client, vcsType, orgName, contextName string) (*api.CircleCIContext, error) {
+
+	contexts, err := api.ListContexts(client, orgName, vcsType)
```
glenjamin: Can the graphQL api filter for us, rather than having to download the whole list including env vars every time?

api/context.go
```diff
@@ -0,0 +1,290 @@
...
+				return fmt.Errorf("Invalid vcs-type '%s' provided, expected one of %s", strings.ToLower(details.Value), allowedValues)
+			}
+		}
+	}
+	return err
+}
+
+func CreateContext(cl *client.Client, contextName, orgName, vcsType string) error {
+
+	contexts, err := ListContexts(cl, orgName, vcsType)
```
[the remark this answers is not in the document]
marcomorain: That's a good point, this is a little sloppy to be pulling in the contexts when they are not strictly necessary. I can call `api.getOrganization` to request _just_ the Org's ID.

api/context_test.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+		}
+
+		result.CreateContext.Error.Type = ""
+
+		response := client.Response{
+			Data: result,
+		}
+
+		server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
```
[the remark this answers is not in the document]
marcomorain: that's a good idea 👌

cmd/context.go
marcomorain: 🤔 

image

We call them "Environment Variables" in the UI, and "Resources" in the GraphQL schema.

cmd/context.go
```diff
@@ -0,0 +1,239 @@
...
+	} else {
+		fmt.Print("Enter secret value and press enter: ")
+		reader := bufio.NewReader(os.Stdin)
+		str, err := reader.ReadString('\n')
+		return strings.TrimRight(str, "\n"), err
+	}
+}
+
+func createContext(client *client.Client, vcsType, orgName, contextName string) error {
+	return api.CreateContext(client, vcsType, orgName, contextName)
```
eric-hu: I ran into an error when checking this locally:

```
 $ go run main.go context create github eric-hu foo
Error: Unable to find organization eric-hu of vcs-type foo: Provided argument value `foo' is not member of enum type.
exit status 255
```


It looks like `api.CreateContext` uses a different argument ordering:

https://github.com/CircleCI-Public/circleci-cli/pull/362/files#diff-9f8031b3385a5e9aa0cdd6f639ce2aa0R51

cmd/context.go
eric-hu: +1 for hiding this subcommand. Then this PR can be merged separately from context group CRUD

cmd/context.go
eric-hu: FWIW `web-ui-org-settings` renames them "Environment Variables" as well in its shim:

https://github.com/circleci/web-ui-org-settings/blob/5bcdce25669ba7fe8c5bfc91d4035f0f5199a49c/src/graphql/schema.graphql#L23

PR #366, opened 2020-02-22
Add command to open current project

marcomorain:
Running `circleci open` will open the current git project on
circleci.com

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/CircleCI-Public/circleci-cli/git"
+	"github.com/pkg/browser"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func projectUrl(remote *git.Remote) string {
+	return fmt.Sprintf("https://app.circleci.com/%s/%s/%s/pipelines",
+		url.PathEscape(strings.ToLower(string(remote.VcsType))),
+		url.PathEscape(remote.Organization),
+		url.PathEscape(remote.Project))
+}
+
+var errorMessage = `
+Unable detect which URL should be opened. This command is intended to be run from
+a git repository with a remote named 'origin' that is hosted on Github or Bitbucket
+Error`
+
+func openProjectInBrowser() error {
+
+	remote, err := git.InferProjectFromGitRemotes()
+
+	if err != nil {
+		return errors.Wrap(err, errorMessage)
+	}
+
+	return browser.OpenURL(projectUrl(remote))
+}
+
+func newOpenCommand() *cobra.Command {
+
+	openCommand := &cobra.Command{
+		Use:   "open",
+		Short: "Open the current project in the browser.",
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return openProjectInBrowser()
+		},
+	}
+	return openCommand
+}
--- a/git/git.go
+++ b/git/git.go
@@ -1,0 +1,105 @@
+package git
+
+import (
+	"errors"
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+type VcsType string
+
+const (
+	GitHub    VcsType = "GITHUB"
+	Bitbucket VcsType = "BITBUCKET"
+)
+
+type Remote struct {
+	VcsType      VcsType
+	Organization string
+	Project      string
+}
+
+// Parse the output of `git remote` to infer what VCS provider is being used by
+// in the current working directory. The assumption is that the 'origin' remote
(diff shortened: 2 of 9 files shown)


cmd/open.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"github.com/CircleCI-Public/circleci-cli/git"
+	"github.com/spf13/cobra"
+)
+
+func openBrowser(url string) error {
```
glenjamin: Perhaps this? https://godoc.org/github.com/pkg/browser

A quick skim of the code suggests it's not obviously better, but `pkg` is generally a trusted source

git/git.go
```diff
@@ -0,0 +1,94 @@
...
+	Bitbucket VcsType = "BITBUCKET"
+)
+
+type Remote struct {
+	VcsType      VcsType
+	Organization string
+	Project      string
+}
+
+func InferOrganizationFromGitRemotes() (*Remote, error) {
```
glenjamin:
```suggestion
func InferProjectFromGitRemotes() (*Remote, error) {
```

cmd/open.go
marcomorain: good idea

cmd/open.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/CircleCI-Public/circleci-cli/git"
+	"github.com/pkg/browser"
+	"github.com/spf13/cobra"
+)
+
+func projectUrl(remote *git.Remote) string {
+	return fmt.Sprintf("https://app.circleci.com/%s/%s/%s/pipelines",
```
glenjamin: does go have a url templating function? I suspect these values should be safe given that they came from the git URL - but we should probably be using some equivalent of `encodeURIComponent` on these path components when assembling a URL

cmd/open.go
```diff
@@ -0,0 +1,40 @@
...
+		remote.Organization,
+		remote.Project)
+}
+
+func openProjectInBrowser() error {
+
+	remote, err := git.InferProjectFromGitRemotes()
+
+	if err != nil {
+		return err
```
glenjamin: Are these errors human-friendly enough? should we be wrapping here with something nicer?

cmd/open.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/CircleCI-Public/circleci-cli/git"
+	"github.com/jtacoma/uritemplates"
+	"github.com/pkg/browser"
+	"github.com/spf13/cobra"
+)
+
+func projectUrl(remote *git.Remote) (string, error) {
+
+	template, err := uritemplates.Parse("https://app.circleci.com/{vcs}/{org}/{project}/pipelines")
```
glenjamin: This lib looks a bit old, I dunno how else to tell if it's a decent choice.

`url.PathEscape` might be a safer bet

marcomorain: I'd equate old with "stable" in this case – it's a small library that does a straightforward task.

cmd/open.go
marcomorain: I've upgraded the error reporting:

```
Unable detect which URL should be opened. This command is intended to be run from
a git repository with a remote named 'origin' that is hosted on Github or Bitbucket
Error: This command must be run from inside a git repository
```

cmd/open.go
danmux: Is the new dependancy, values map and extra error handling really better than..

```
func projectUrl(remote *git.Remote) string {
	return fmt.Sprintf("https://app.circleci.com/%s/%s/%s/pipelines",
		url.PathEscape(strings.ToLower(string(remote.VcsType))),
		url.PathEscape(remote.Organization),
		url.PathEscape(remote.Project))
}
```

cmd/open.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/CircleCI-Public/circleci-cli/git"
+	"github.com/jtacoma/uritemplates"
+	"github.com/pkg/browser"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func projectUrl(remote *git.Remote) (string, error) {
+
+	template, err := uritemplates.Parse("https://app.circleci.com/{vcs}/{org}/{project}/pipelines")
```
danmux: Is the new dependancy, values map and extra error handling really better than..

```
func projectUrl(remote *git.Remote) string {
	return fmt.Sprintf("https://app.circleci.com/%s/%s/%s/pipelines",
		url.PathEscape(strings.ToLower(string(remote.VcsType))),
		url.PathEscape(remote.Organization),
		url.PathEscape(remote.Project))
}
```

marcomorain: nope, I added it at @glenjamin 's suggestion. I've removed it.

PR #369, opened 2020-02-27
CIRCLE-25004 Support Config Version 2.1 in Local Builds

marcomorain:
Allow using any version of config in local builds. Remove our
config-version guards and instead use the API to process config.

- find which config file is specified
- process the config using the API
- write the processed config to a temp file
- mount the temp file into the docker container so that the build can
  access it.


local/local.go
```diff
@@ -43,50 +37,65 @@ func UpdateBuildAgent() error {
...
 
 	if err != nil {
 		return errors.Wrap(err, "Could not find picard image")
 	}
 
-	// TODO: marc:
-	// We are passing the current environment to picard,
-	// so DOCKER_API_VERSION is only passed when it is set
-	// explicitly. The old bash script sets this to `1.23`
-	// when not explicitly set. Is this OK?
```
marcomorain: We've not had any issues with this as far as I am aware. It was indeed OK.

local/local.go
```diff
@@ -43,58 +38,53 @@ func UpdateBuildAgent() error {
...
+	if err != nil {
+		return err
+	}
+
+	if !configResponse.Valid {
+		return fmt.Errorf("config errors %v", configResponse.Errors)
 	}
 
-	if err := validateConfigVersion(opts.Args); err != nil {
+	processedConfigPath, err := writeStringToTempFile(configResponse.OutputYaml)
```
aengelberg: Could you add a `defer os.Remove(processedConfigPath)` after this, or does the file need to last longer than this function?

marcomorain: yeah, the file needs to persist until after the docker command has executed, but `syscall.Exec` replaces the current process, so a `defer` would never be called. I'll add a comment to note that a file will be left in `/tmp` after each build.

PR #371, opened 2020-02-28
Fix bin directory used by snap job.

### LINKED ISSUE #370: Snap Isn't Publishing
The snap isn't publishing because the job is failing.

I'm looking into it.

felicianotech:
Fixes #370.

Somewhere along the line the directory where the compiled binaries are stored changed. The snap job then couldn't find the binary. This PR fixes that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,7 +208,7 @@ jobs:
       - attach_workspace:
           at: .
       - run: |
-          export TAG=$(./dist/linux_amd64/circleci version)
+          TAG=$(./dist/circleci-cli_linux_amd64/circleci version) && export TAG
           sed -i -- "s/%CLI_VERSION_PLACEHOLDER%/$TAG/g" snap/snapcraft.yaml
       - run: snapcraft
       - run:


marcomorain:
Thanks @felicianotech – I must have broken this in #367 when upgrading `goreleaser` (to fix other problems in the release process). I'll check if homebrew has the same problem now.

.circleci/config.yml
```diff
@@ -208,7 +208,7 @@ jobs:
       - attach_workspace:
           at: .
       - run: |
-          export TAG=$(./dist/linux_amd64/circleci version)
+          export TAG=$(./dist/circleci-cli_linux_amd64/circleci version)
           sed -i -- "s/%CLI_VERSION_PLACEHOLDER%/$TAG/g" snap/snapcraft.yaml
```
marcomorain: I'm wondering about how to make this fail – we could try this?

```yml
      - run:
           shell: /bin/bash -eo pipefail -o nounset
           command: |
             export TAG=$(./dist/circleci-cli_linux_amd64/circleci version)
             sed -i -- "s/%CLI_VERSION_PLACEHOLDER%/$TAG/g" snap/snapcraft.yaml
```

felicianotech: Maybe? I'm not sure. How do you mean fail? I was already failing (red build).

Another way might be to define a regex for the CLI version. Then, after we pull the version here we can make sure it matches?

felicianotech:
Thanks.

> I must have broken this in #367 when upgrading goreleaser

No worries. I'm excited to see the CLI gain momentum. :+1:

PR #374, opened 2020-03-04
Fix homebrew deployments

marcomorain:
The homebrew job has been failing on master, this PR addresses two
issues:

- curl is failing to download from circle.ci due to some strange HTTP2
  issue that I don't fully understand. Work around this by forcing
  HTTP 1.1 (the same way that we do for codecov.io uploads)
- Use the latest macOS image, since Homebrew now requires a very recent
  Ruby to run, which is installed on the latest macOS images by default.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,15 @@ executors:
       HOMEBREW_NO_AUTO_UPDATE: 1
 
 commands:
+  force-http-1:
+    steps:
+      - run:
+          # Uploading to codecov has been failing due to HTTP 2.0 issues.
+          # https://app.circleci.com/jobs/github/CircleCI-Public/circleci-cli/6480
+          #     curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
+          # The issue seems to be on the server-side, so force HTTP 1.1
+          name: 'cURL: Force HTTP 1.1'
+          command: echo '--http1.1' >> ~/.curlrc
   build-docker-image:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


marcomorain:
This build is a failing example on master: https://app.circleci.com/jobs/github/CircleCI-Public/circleci-cli/6725

.circleci/config.yml
```diff
@@ -19,6 +19,15 @@ executors:
       HOMEBREW_NO_AUTO_UPDATE: 1
 
 commands:
+  force-http-1:
```
hannahhenderson: 😍 DRY config with meaningful names!

marcomorain:
Docs added 👌

glenjamin:
>     * curl is failing to download from circle.ci due to some strange HTTP2
>       issue that I don't fully understand. Work around this by forcing
>       HTTP 1.1 (the same way that we do for codecov.io uploads)

This is caused by a combination of bit.ly using http2, and a certain version of cURL that doesn't switch protocols if an HTTP2 response redirects to an HTTP1 server.

vcavallo:
> > ```
> > * curl is failing to download from circle.ci due to some strange HTTP2
> >   issue that I don't fully understand. Work around this by forcing
> >   HTTP 1.1 (the same way that we do for codecov.io uploads)
> > ```
> 
> This is caused by a combination of bit.ly using http2, and a certain version of cURL that doesn't switch protocols if an HTTP2 response redirects to an HTTP1 server.

I seem to have that "certain version", as I am running into the issue referenced in this PR. My curl version: 

```
curl 7.58.0 (x86_64-pc-linux-gnu) libcurl/7.58.0 OpenSSL/1.1.1 zlib/1.2.11 libidn2/2.0.4 libpsl/0.19.1 (+libidn2/2.0.4) nghttp2/1.30.0 librtmp/2.3
Release-Date: 2018-01-24
```

The error:

```
$ curl -fLSs https://circle.ci/cli | bash
curl: (16) Error in the HTTP2 framing layer
```

PR #375, opened 2020-03-04
Unhide the `context` command

marcomorain:
Update the description of the sub-commands to be in line with the rest
of the UI in the product, and un-hide the command so that it shows up in
`help`.


cmd/context.go
```diff
@@ -25,13 +25,12 @@ func newContextCommand(config *settings.Config) *cobra.Command {
 	}
 
 	command := &cobra.Command{
-		Hidden: true,
```
marcomorain: I removed `Hidden`

PR #397, opened 2020-04-23
Pass string array arguments to docker correctly.

### LINKED ISSUE #391: Environment variables are not properly set

**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
When i set environment variables using `-e` on the command line they are not available on the build. When i execute:
```
circleci build --job test -e FOO1=BAR1 -e FOO2=BAR2
```
the env vars FOO1 and FOO2 are not accessible in the build, executing `env` as part of the build i can see:
```
[FOO1=BAR1,FOO2=BAR2]
DOCKER_VERSION=18.06.3
HOSTNAME=c48870140fea
SSH_AUTH_SOCK=/tmp/circleci-692488189/ssh_auth_sock
CIRCLE_WORKING_DIRECTORY=~/project
....
```
**What is the expected behavior?**
The variables set on the command line are available on the build.
**Which version of the CLI and OS are you using? Did this work in previous versions?**
```
$ circleci version
0.1.6949+d951526
$ circleci diagnostic
circleci diagnostic

---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: /home/fgimenez/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API... 
Ok.
Hello, Federico Gimenez.
```

marcomorain:
String array arguments were being converted to strings using golang
array formatting. So `--foo a --foo b --foo c` was being passed to
docker as `--foo [a,b,c]`. Handle this case explicitly in the code.

Fixes #391 #395

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -135,7 +135,7 @@ func buildAgentArguments(flags *pflag.FlagSet) ([]string, string) {
 	// build a list of all supplied flags, that we will pass on to build-agent
 	flags.Visit(func(flag *pflag.Flag) {
 		if flag.Name != "config" && flag.Name != "debug" {
-			result = append(result, "--"+flag.Name, flag.Value.String())
+			result = append(result, unparseFlag(flags, flag)...)
 		}
 	})
 	result = append(result, flags.Args()...)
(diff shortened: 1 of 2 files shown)


glenjamin:
I wonder if it would be simpler to read from the original `argv` when we're doing passthrough?

PR #403, opened 2020-05-06
Add Docker CLI to Alpine image

fsproru:

Hi there, noticed that running `circleci local execute` in Docker fails with the following error:

```
Error: could not find `docker` on the PATH; please ensure that docker is installed
```

Was running it using the following command:

```
docker run --rm -v $(pwd):/project-root  -v /var/run/docker.sock:/var/run/docker.sock --workdir /project-root circleci/circleci-cli:alpine local execute
```

Looking through the code, looks like the CLI runs `docker version` and errors out with the error I'm seeing if the call is unsuccessful. I connected to a container based on the latest alpine image. It looks like the image itself doesn't have docker command line installed. After installing it with `apk add docker` and passing in the socket with `-v /var/run/docker.sock:/var/run/docker.sock` I was able to get `docker version` to return a version inside the Docker container successfully.

Adding `docker` to the list of packages should work, you may need to add `apk update` before installing it to update the package index.

Wasn't able to find easily how to build the image with the new changes to see if it fixes the error. If you can either help to get this PR over the finish line or show me how to build a Docker image with the changes that would be great.

Thank you.

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -2,6 +2,6 @@ FROM alpine:3.8
 
 COPY ./dist/circleci-cli_linux_amd64/circleci /usr/local/bin
 
-RUN apk add --no-cache --upgrade git openssh ca-certificates
+RUN apk add --no-cache --upgrade git openssh ca-certificates docker
 
 ENTRYPOINT ["circleci"]


fsproru:
Hey @marcomorain, @johnswanson, just following up on this one. Would you mind reviewing it so we can release our CircleCI integration?

Thank you 🙌

marcomorain:
Hi @fsproru thanks for the PR.

I've been thinking about this, trying to weigh up the cost (the Docker image size increases from 30mb to ~250mb) with the benefits.

Can you explain a little about what you are planning to build on top of this please? I'm not clear on what the use-case is, and there might be a better way to achieve the same effect.

The command to [run local builds shells out directly to docker to run](https://github.com/CircleCI-Public/circleci-cli/blob/93b9d40689829c229f6aef55ba2d6c5c3d2ea498/local/local.go#L289-L299), so running docker, and passing in the docker socket, in order to shell out to docker seems like a lot of overhead. It might be more straightforward for you to implemented [this code](https://github.com/CircleCI-Public/circleci-cli/blob/93b9d40689829c229f6aef55ba2d6c5c3d2ea498/local/local.go#L289-L299) yourself from a shell script?

fsproru:
@marcomorain, thanks for the hint. We would like to be able to run a
CircleCI job from our app. I traced down the execution path and it looks
like circleci-cli does more things than just `docker run` (like processing
a config, etc). This means that we should use cirlceci CLI instead of
calling `docker run` directly.

You May still want to update README.md for circleci-cli though since
`docker run` example there results into this issue.

PR #405, opened 2020-05-07
Send fabricated pipeline values when validating config

johnswanson:
Construct fake pipeline values and send them along to the `buildConfig` graphQL query when validating config. This fixes config validation for configs that use pipeline values.

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -1,0 +1,45 @@
+package pipeline
+
+import "sort"
+
+// CircleCI provides various `<< pipeline.x >>` values to be used in your config, but sometimes we need to fabricate those values when validating config.
+type Values map[string]string
+
+
+func FabricatedValues() Values {
+	return map[string]string{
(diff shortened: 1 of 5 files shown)


pipeline/pipeline.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// PrepareForGraphQL takes a golang homogenous map, and transforms it into a list of keyval pairs, since GraphQL does not support homogenous maps.
+func PrepareForGraphQL(kvMap Values) []KeyVal {
+	// we need to create the slice of KeyVals in a deterministic order for testing purposes
+	keys := make([]string, 0, len(kvMap))
+	for k := range kvMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
marcomorain, line 38: Is the sorting here to make the order deterministic (for testing, etc)?

aengelberg: Indeed, there's a comment about this above.

PR #407, opened 2020-05-11
Infer the git pipeline parameters

marcomorain:
If git is available, read the remote, sha, branch and tag from it.
This creates more realistic pipeline parameters.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -103,3 +103,33 @@ func getRemoteUrl(remoteName string) (string, error) {
 	}
 	return string(out), nil
 }
+
+func commandOutputOrDefault(cmd *exec.Cmd, defaultValue string) string {
+	output, err := cmd.CombinedOutput()
+
+	if err != nil {
+		return defaultValue
+	}
+
+	return strings.TrimSpace(string(output))
+}
+
+func Branch() string {
+	// Git 2.22 (Q2 2019)added `git branch --show-current`, but using
+	// `git rev-parse` works in all versions.
+	return commandOutputOrDefault(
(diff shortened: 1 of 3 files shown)


git/git.go
```diff
@@ -103,3 +103,33 @@ func getRemoteUrl(remoteName string) (string, error) {
...
+	}
+
+	return strings.TrimSpace(string(output))
+}
+
+func Branch() string {
+	// Git 2.22 (Q2 2019)added `git branch --show-current`, but using
+	// `git rev-parse` works in all versions.
+	return commandOutputOrDefault(
+		exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD"),
```
johnswanson, line 121: If the user has a detached HEAD, this will return `HEAD`, which doesn't seem right, but it seems like some inaccuracy in these values is probably fine.

pipeline/pipeline.go
```diff
@@ -1,22 +1,41 @@
...
+	"github.com/CircleCI-Public/circleci-cli/git"
+)
 
 // CircleCI provides various `<< pipeline.x >>` values to be used in your config, but sometimes we need to fabricate those values when validating config.
 type Values map[string]string
 
-
 func FabricatedValues() Values {
+
+	revision := git.Revision()
```
johnswanson, line 15: It initially threw me off that we use this value for both `git.revision` and `git.base_revision` but I think it may make sense. The reality is the working directory may have state that isn't part of any commit, and in that case the current `HEAD` really is the "previous" revision. Plus I think we're only aiming for "realism" here, not 100% accuracy.

PR #418, opened 2020-05-19
Skip auto-update check while running in CI.

gmemstr:

The CLI will always check for an update when run, unless explicitly told
not to. This can lead to rate limiting on shared IPs, especially in
CircleCI jobs. This helps alleviate this issue by checking for the $CI
environment variable before attempting to check for an update on GitHub.

It might be better to throw a message that the CLI is outdated but
continue with the command, however, this would have to account for being
piped into configuration files.

### DIFF
--- a/cmd/check_test.go
+++ b/cmd/check_test.go
@@ -57,7 +57,7 @@ var _ = Describe("Check", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 
 			command = commandWithHome(checkCLI, tempSettings.Home,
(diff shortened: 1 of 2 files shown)


gmemstr:
Realized slightly too late this would break tests - looking into it.

marcomorain:
@gmemstr I threw a new commit on top of yours with an alternate approach to this - taking CI into account earlier in the process lifecycle.

gmemstr:
@marcomorain Good call! Sooner we capture it the better.

PR #425, opened 2020-05-29
fix clone url

KyleTryon:
URL was previously not suitable for cloning. This change will allow the user to copy and paste the command.

- [ x] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/master/CONTRIBUTING.md).
- [ x] I have checked for similar issues and haven't found anything relevant.
- [ x] This is not a security issue (which should be reported here: https://circleci.com/security/)


corinnesollows (approved):
PR is switching to use the SSH for cloning

PR #264, opened 2019-01-15
WIP DO NO MERGE: Disable update checks in the CLI Docker images

eric-hu:
This approach does not use the approach as discussed in this PR comment,
as I tried CIRCLECI_CLI_SKIP_UPDATE_CHECK and couldn't get it working
successfully within Docker on my local machine.

https://github.com/CircleCI-Public/circleci-cli/issues/226

- [x] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/master/CONTRIBUTING.md).
- [x] I have checked for similar issues and haven't found anything relevant.
- [x] This is not a security issue (which should be reported here: https://circleci.com/security/)

The tests aren't currently passing on my local machine. The tests around
`circleci update` are failing. I haven't changed any Go code, so I want
to see how they behave on CI.

Issue #226, opened 2018-12-11
Docker image shouldn't check for new version


CaptainPatate:
I think it is useless to check for new `circleci-cli` version each time we run the Docker image. I am not even sure that it makes sense to verify `circleci-cli` version at all inside Docker.

CaptainPatate:
We found this https://github.com/CircleCI-Public/circleci-cli/commit/002bec6978ab1da3985edbb808522131e7502e8b. Is there any reason to mark `--skip-update-check` option as hidden? This option would solve this issue.

tgermain:
> Btw, when you say "verify `circleci-cli` version inside Docker", what do you mean exactly? Can you elaborate on your use-case? I want to be sure I understand, if this means you're using the official docker image of the CLI -- I think that's a good idea we can maybe disable the update check and command in this version.

(disclaimer: @CaptainPatate and I are working together)

We are talking about the official docker image for the CLI here indeed. 
For tools running in docker I do not think it is the best way to have software trying to auto-update themselves in the container when you run it.

The docker image tag already specify which version we are willing to run.

As a user if I want to use the latest version of the CLI available I can use the `latest` tag (or `alpine` tag for more lightweight image) IMHO.

To give you a little bit more context about how we are using the CLI : we have a pre-commit hook which simply validate the CircleCI configuration. 
Hence we want to do that check fast. Also we want to reduce the amount of tools needed to make the pre-commit hook work.
So to avoid installing CircleCI CLI locally we think of running the validation using the CircleCI CLI docker image. 

But here are our first comparison between local CLI and dockerized CLI : 
```
# docker image circleci/circleci-cli:alpine was already pulled before the test
$ time docker  run -v $(pwd)/.circleci/config.yml:/config.yml -ti circleci/circleci-cli:alpine config validate /config.yml 
Config file at /config.yml is valid.

real    0m2,569s
user    0m0,023s
sys     0m0,064s
$ time circleci config validate .circleci/config.yml 
Config file at .circleci/config.yml is valid.

real    0m0,535s
user    0m0,067s
sys     0m0,016s
```

dockerized CLI is **5 times** slower than local CLI. This make the pre-commit checks feel sluggish 😞.

I do not have the exact number at hands but with  `--skip-update-check` it is much better (~1s IIRC).

marcomorain:
Setting `CIRCLECI_CLI_SKIP_UPDATE_CHECK` to true inside the official docker image is one way to work around this?

[...]

PR #428, opened 2020-06-08
Add ENV flag to skip update check in Docker container.

gmemstr:

See #226 - This builds off a previous PR (#418) to ensure the Docker image doesn't try to auto-update by implementing the `CIRCLECI_CLI_SKIP_UPDATE_CHECK` flag. #264 previous attempted to do this, but the variable wasn't checked within the CLI itself.


gmemstr:
Was initially confused why it still checked for an update in the `deploy-test` job, but it appears the `circleci update check` command is being explicitly run.

PR #429, opened 2020-06-09
Add job to start Chocolatey package deployment [skip ci]

gmemstr:
Adds a job to the configuration to trigger a new job in the https://github.com/CircleCI-Public/chocolatey-circleci-cli repository, which builds and deploys a new package to the Chocolatey repository.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -301,6 +301,7 @@ jobs:
           name: Trigger Chocolatey repository build
           command: |
             curl -X POST https://circleci.com/api/v2/project/gh/CircleCI-Public/chocolatey-circleci-cli/pipeline \
+              --fail --silent \
               -H 'Content-Type: application/json' \
               -H 'Accept: application/json' \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -293,6 +293,19 @@ jobs:
...
 
+  trigger-chocolatey-deploy:
+    docker:
+      - image: cimg/base:edge
+    steps:
+      - run:
+          name: Trigger Chocolatey repository build
+          command: |
+            curl -X POST https://circleci.com/api/v2/project/gh/CircleCI-Public/chocolatey-circleci-cli/pipeline \
+              -H 'Content-Type: application/json' \
```
marcomorain, line 304:
```suggestion
              --fail --silent \
              -H 'Content-Type: application/json' \
```

marcomorain: This will cause the step to fail if the server returns a non-2xx response. And `--silent` makes the output neater in non-interactive settings.

PR #430, opened 2020-06-09
Fix SSH CVE

marcomorain:


pete-woods:
@marcomorain I took the liberty of tweaking this a bit, adding some tests, using the (hopefully going to be) common REST client

PR #431, opened 2020-06-12
Fixes bash command error when DESTDIR is a non-path directory

frshwtr:

**If you have any questions, feel free to ping us at @CircleCI-Public/x-team.**
## Problem
We found an issue with the install script on MacOS Catalina with bash 3.2 and Ubuntu 18 with bash 4.4, zsh 5.4.2.

When `DESTDIR` is set to a directory that is not in `PATH`, CircleCI CLI will successfully install but the script will error out on [line 59](https://github.com/CircleCI-Public/circleci-cli/blob/master/install.sh#L59).

Example output:
```
➜ DESTDIR=$(pwd)/clitemp ./install.sh
Starting installation.
Installing CircleCI CLI v0.1.7971
Installing to /home/ben/git/frshwtr/circleci-cli/clitemp
An error occured installing the tool.
The contents of the directory /tmp/tmp.HAAnGm47op have been left in place to help to debug the issue.
```

We suspect this is because `command` is being executed with no directory context. Meaning any install to a non-path directory will fail, causing the install script to fail.

## Testing

We have ran this locally on macOS Catalina with the following cases:
* setting`DESTDIR` and examining the exit code and status of the installed binary.
* doing the same with the default (/usr/local/bin) set by the script.

All appears to work well in both cases.


install.sh
```diff
@@ -56,7 +56,7 @@ curl -sL --retry 3 "${RELEASE_URL}" | tar zx --strip 1
 echo "Installing to $DESTDIR"
 install circleci "$DESTDIR"
 
-command -v circleci
+command -v "$DESTDIR"circleci
```
jvincent42, line 59: We might want to remove any potential trailing `/` in DESTDIR and force a `/` before the binary name.
```suggestion
command -v "${DESTDIR%/}"/circleci
```

jvincent42 (changes requested):
Thanks for your contribution! Sorry for the delay in reviewing your work.
If you still want this PR to be merged, can you please rebase it on main and address the comment I've left ?
Thank you for your time.

pete-woods:
Closing: this targets the legacy v0 install.sh. The v1 CLI's install.sh was rewritten and already handles DESTDIR not being on PATH (it prints a note instead of failing), so the fix here no longer applies.

PR #432, opened 2020-06-15
Orb pack

gmemstr:
* New `orb pack` command, which behaves similar to the `config pack`, but with a stricter project structure.
 * Added `<<include([filepath])>>` syntax to Orb packing to include scripts and files.
 * Add ability to enforce directories walked with `NewTree()` (backwards-compatible)
 * Move everything to `yaml.v3` to consolidate (previously both `gopkg.in/yaml.v2` and `github.com/go-yaml/yaml` were being used)
 * Orb and orb usage example schema

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -293,6 +293,20 @@ jobs:
       - run: ./.circleci/brew-deploy.sh
       - slack-notify-on-failure
 
+  trigger-chocolatey-deploy:
+    docker:
+      - image: cimg/base:edge
+    steps:
+      - run:
+          name: Trigger Chocolatey repository build
+          command: |
+            curl -X POST https://circleci.com/api/v2/project/gh/CircleCI-Public/chocolatey-circleci-cli/pipeline \
+              --fail --silent \
+              -H 'Content-Type: application/json' \
+              -H 'Accept: application/json' \
+              -H "Circle-Token: $CIRCLE_TOKEN" \
+              --data '{ "parameters": { "deploy-production": true } }'
+
 workflows:
   ci:
     jobs:
@@ -329,6 +343,12 @@ workflows:
             - lint
             - deploy-test
             - shellcheck/check
+          filters:
+            branches:
+              only: master
+      - trigger-chocolatey-deploy:
+          requires:
+            - deploy
           filters:
             branches:
               only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
           path: coverage.txt
       - slack-notify-on-failure
 
-  docs:
+  docs: 
     executor: go
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
           path: coverage.txt
       - slack-notify-on-failure
 
-  docs: 
+  docs:
     executor: go
     steps:
       - checkout
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -254,7 +254,8 @@ Please note that at this time all orbs created in the registry are world-readabl
 		RunE: func(_ *cobra.Command, _ []string) error {
 			return packOrb(opts)
 		},
-		Args: cobra.ExactArgs(1),
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
 	}
 
 	orbCreate.Flags().BoolVar(&opts.integrationTesting, "integration-testing", false, "Enable test mode to bypass interactive UI.")
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -770,9 +770,17 @@ type OrbSchema struct {
 	Examples    map[string]ExampleSchema `yaml:"examples,omitempty"`
 }
 
+type ExampleUsageSchema struct {
+	Version   string      `yaml:"version,omitempty"`
+	Orbs      interface{} `yaml:"orbs,omitempty"`
+	Jobs      interface{} `yaml:"jobs,omitempty"`
+	Workflows interface{} `yaml:"workflows"`
+}
+
 type ExampleSchema struct {
-	Description string    `yaml:"description,omitempty"`
-	Usage       OrbSchema `yaml:"usage,omitempty"`
+	Description string             `yaml:"description,omitempty"`
+	Usage       ExampleUsageSchema `yaml:"usage,omitempty"`
+	Result      ExampleUsageSchema `yaml:"result,omitempty"`
 }
 
 func packOrb(opts orbOptions) error {
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -1,11 +1,10 @@
 package cmd
 
 import (
-	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"os"
+	"io/ioutil"
 	"regexp"
 	"sort"
 	"strconv"
@@ -17,8 +16,8 @@ import (
 	"github.com/CircleCI-Public/circleci-cli/prompt"
 	"github.com/CircleCI-Public/circleci-cli/references"
 	"github.com/CircleCI-Public/circleci-cli/settings"
-	"github.com/go-yaml/yaml"
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
 
 	"github.com/spf13/cobra"
 )
@@ -759,6 +758,8 @@ https://circleci.com/orbs/registry/orb/%s
 }
 
 func packOrb(opts orbOptions) error {
+	// Travel our Orb and build a tree from the YAML files.
+	// Non-YAML files will be ignored here.
 	tree, err := filetree.NewTree(opts.args[0])
 	if err != nil {
 		return errors.Wrap(err, "An error occurred trying to build the tree")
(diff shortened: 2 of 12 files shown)


.circleci/config.yml
```diff
@@ -166,7 +166,7 @@ jobs:
           path: coverage.txt
       - slack-notify-on-failure
 
-  docs:
+  docs:
```
KyleTryon, line 169: lint space

cmd/orb.go
```diff
@@ -740,3 +758,144 @@ https://circleci.com/orbs/registry/orb/%s
...
+	// Travel our Orb and build a tree from the YAML files.
+	// Non-YAML files will be ignored here.
+	_, err := os.Stat(filepath.Join(opts.args[0], "@orb.yml"))
+	if err != nil {
+		return errors.Wrap(err, "@orb.yml file not found, are you sure this is the Orb root?")
+	}
+
+	tree, err := filetree.NewTree(opts.args[0], "executors", "jobs", "commands", "examples")
+	if err != nil {
+		return errors.Wrap(err, "An error occurred trying to build the tree")
```
johnswanson, line 796: Many of these error messages seem confusing to the user:
```
Error: @orb.yml file not found, are you sure this is the Orb root?: stat @orb.yml: no such file or directory
Error: An error occurred trying to build the tree: ...
Error: Failed trying to travel Orb YAML: ...
```

The colon separation is one issue (maybe we should look at replacing that with newlines at some point but probably out of scope here), making errors that end with punctuation look bad.

I'd also say these seem more directed towards the programmer (saying where in the code things failed) than at the user, who might not understand something like "what does it mean to travel Orb YAML" or "what tree is orb pack building?" I think an error message like "an unexpected error occurred" might be better (still wrapping the underlying error) because it removes information that isn't relevant to the user.

cmd/orb.go
```diff
@@ -740,3 +758,144 @@ https://circleci.com/orbs/registry/orb/%s
...
+			filepath := filepath.Join(orbRoot, includeMatches[1])
+			file, err := ioutil.ReadFile(filepath)
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("Could not open %s for inclusion in Orb", filepath))
+			}
+
+			node.Value = string(file)
+		}
+
+		// Escape unclosed `<<` for heredocs. First we partially match the syntax,
```
johnswanson, line 867: I don't think this should be a feature of orb pack, as it's difficult to do right in all situations (what happens if a heredoc is already escaped here? what happens if I am doing something like `cat <<EOF >> appended-to.txt`?) and it means that copying config from a config into your orb will have sharp edges.

I think we should be really wary of making "orb config" (the input to orb packing) different from regular config (the output of orb packing), except in really obvious ways (like inserting scripts).

cmd/orb.go
```diff
@@ -740,3 +758,144 @@ https://circleci.com/orbs/registry/orb/%s
...
+		return errors.Wrap(err, "Failed trying to marshal Orb YAML")
+	}
+
+	fmt.Println(string(finalOrdered))
+
+	return nil
+}
+
+// Travel down a YAML node, replacing values as we go.
+func travelOrbTree(node *yaml.Node, orbRoot string) error {
```
johnswanson, line 840: nit: I'd prefer a name about what it's doing when it travels instead - `inlineIncludes` or something like that.

cmd/orb_test.go
```diff
@@ -2581,4 +2581,69 @@ https://circleci.com/orbs/registry/orb/my/orb
 			})
 		})
 	})
+
```
johnswanson, line 2584: Since all of our Go tests have this really painful format but are fundamentally just doing "run the thing with some conditions, check the output," we've been using (and really enjoying) cucumber for this, for example: https://github.com/CircleCI-Public/circleci-cli/blob/master/integration_tests/features/contexts.feature

I'd highly recommend writing these as cucumber tests, they should only be a few lines each.

johnswanson: We should have tests for the failure cases too - e.g. what happens when `@orb.yml` doesn't exist, what happens when an included file doesn't exist, etc.

cmd/orb.go
```diff
@@ -740,3 +758,144 @@ https://circleci.com/orbs/registry/orb/%s
...
+	err = yaml.Unmarshal(final, &orb)
+	if err != nil {
+		return errors.Wrap(err, "Failed unmarshal YAML tree")
+	}
+
+	finalOrdered, err := yaml.Marshal(&orb)
+	if err != nil {
+		return errors.Wrap(err, "Failed trying to marshal Orb YAML")
+	}
+
```
johnswanson, line 833: can/should we validate the orb before we print it out?

cmd/orb.go
```diff
@@ -740,3 +758,144 @@ https://circleci.com/orbs/registry/orb/%s
...
+	if err != nil {
+		return errors.Wrap(err, "@orb.yml file not found, are you sure this is the Orb root?")
+	}
+
+	tree, err := filetree.NewTree(opts.args[0], "executors", "jobs", "commands", "examples")
+	if err != nil {
+		return errors.Wrap(err, "An error occurred trying to build the tree")
+	}
+
+	y, err := yaml.Marshal(&tree)
```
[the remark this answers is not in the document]
gmemstr: Yeah, I completely agree that it's a bit of a weird string of transformations. I'll see about implementing your suggestion, sounds like it could work better.

cmd/orb.go
gmemstr: I agree. This is something we added somewhat after the fact, but I'll look into pulling it out - would it be worth making a new PR to add this behaviour to both the config and orb pack (if we can refine the behaviour)? I think it's generally a positive addition, but hard to get right.

cmd/orb.go
gmemstr: Excellent catch - yes, we should. Could this extend to the `config pack` command as well, in the same way the heredoc escaping might?

cmd/orb.go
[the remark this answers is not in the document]
gmemstr: I've done some digging/trial and error, and I don't think we can get around this without removing the patch for ordering the example stanzas in a predictable way, or repeating calls to `inlineIncludes`/`travelOrbTree`  since we can't iterate over a struct (not without some reflection anyways, which I would rather avoid). Not sure the best way to approach this outside of those two solutions, but very open to suggestions @johnswanson

cmd/orb.go
johnswanson: I agree that it'd be positive if we could do it right, but it feels too magical to me, I think - it feels simpler to just always require escaping.

If we could do it, I think it'd be better to do it in config compilation itself (server side) so it'd work everywhere, including config being executed.

cmd/orb.go
johnswanson: Yeah - though since it's not in `config pack` maybe it's scope creep to do it here anyway.

I'm not totally sure whether people would ever want to pack something even if it was invalid. Maybe we could just echo something on stdout to warn them, but still output the orb (or config) on stdout. But no need to do it here for now.

cmd/orb.go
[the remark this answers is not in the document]
johnswanson: What would you think of just doing it with multiple calls to `inlineIncludes`? One each for `executors`, `commands`, and `jobs`? I think that would be cleaner - annoyingly repetitive, but closer to the intent.

integration_tests/features/orb_pack.feature
```diff
@@ -0,0 +1,48 @@
...
+                    command: echo Hello, world!
+    """
+    And the exit status should be 0
+
+  @mocked_home_directory
+  Scenario: Missing @orb.yml for orb packing
+    When I run `circleci orb pack src`
+    Then the output should contain:
+    """
+    Error: @orb.yml file not found, are you sure this is the Orb root?
```
johnswanson, line 29: :tada:

integration_tests/features/orb_pack.feature
```diff
@@ -0,0 +1,48 @@
...
+    commands:
+        test:
+          steps:
+            - run:
+                command: <<include(script.sh)>>
+    """
+    When I run `circleci orb pack src`
+    Then the output should contain:
+    """
+    Error: An unexpected error occurred: Could not open src/script.sh for inclusion in Orb
```
johnswanson, line 46: :tada:

integration_tests/features/orb_pack.feature
```diff
@@ -0,0 +1,48 @@
...
+    When I run `circleci orb pack src`
+    Then the output should contain:
+    """
+    commands:
+        test:
+            steps:
+                - run:
+                    command: echo Hello, world!
+    """
+    And the exit status should be 0
```
johnswanson, line 22: This is great, thanks for adding these tests!

cmd/orb.go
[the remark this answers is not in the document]
gmemstr: I suppose that would work, but then we remove the ability to `<<include()>>` in examples - which isn't a huge deal imo.

gmemstr:
Hiding the `orb pack` for beta test as per John's suggestion :)

roopakv:
juts fwiw @gmemstr this broke our CI pipelines because the way the config is now packed is different from the previous versions.

And the mac version view brew is not updated to have the latest version. what would be the suggested fix

gmemstr:
@roopakv Can you confirm whether you switched to the `orb pack` command, or are you still using `config pack`? Further, do you have any examples of this?

roopakv:
@gmemstr the spacing on config pack output changed with this release which was not backward compatible that ended up breaking a few build pipelines.

We use config pack for our main config yml

PR #435, opened 2020-06-22
Validate that a circle token has been set before running context commands

aengelberg:
This shows a more helpful error message in the case that a circle token has never been set, but not when the token is set and authn/authz fails.

This uses a function `validateToken` that other commands already use, so we probably just forgot to use it here.

### DIFF
--- a/integration_tests/features/contexts.feature
+++ b/integration_tests/features/contexts.feature
@@ -1,0 +1,16 @@
+
+# Aruba reference
+# https://gist.github.com/bdunn313/4199906
+
+Feature: Context integration tests
+
+  @mocked_home_directory
+  Scenario: when listing contexts without a token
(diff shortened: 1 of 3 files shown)


integration_tests/features/contexts.feature
```diff
@@ -0,0 +1,16 @@
+
+# Aruba reference
+# https://gist.github.com/bdunn313/4199906
+
+Feature: Context integration tests
+
+  @mocked_home_directory
+  Scenario: when listing contexts without a token
+    When I run `circleci context list github foo --skip-update-check --token ""`
```
marcomorain, line 9: @aengelberg wdyt of using a cucumber test for this?

aengelberg: Didn't know about cucumber till now, but I don't see why not :+1:

PR #389, opened 2020-03-28
Handle build array type arguments the right way.

simi:
Closes #385, but with different approach than #386. It fixes `--volume` as well. pflag dependency needs to be updated to newer version to support `pflag.SliceValue` interface.

:information_source: I'm not go programmer, but I did my best to follow code standards. IMHO this would deserve test coverage as well. If anyone can confirm this is good way to solve this problem, I can take a look and add some test coverage as well.

PR #440, opened 2020-07-09
--env contains commas get unexpectedly split

Kuniwak:
`StringSliceP` splits an argument by commas, so parsing results get corrupt if the argument contains comma.
So we should use `StringArrayP` instead of `StringSliceP`.

* StringSliceP: `--ss a,b --ss c`   →  `["a", "b", "c"]` 😕 
* StringArrayP: `--ss a,b --ss c`  →  `["a,b", "c"]` 😄 

Arguments contain commas are frequently present when you use private images on Google Container Registry (GCR).
GCR can use a password encoded by JSON (called as JSON key file [^1]) for native basic auth [^2] of Docker.
And only we can do without embedding the password into .circleci/config.yml is specifying the password via --env.

[^1]: https://cloud.google.com/container-registry/docs/advanced-authentication#json-key
[^2]: https://docs.docker.com/registry/deploying/#native-basic-auth

Related #389 

---


Kuniwak:
**FYI**: We can avoid this problem using `--` before `--env`. For example:

```console
$ circleci local execute --job build -- --env FOO=BAR,BAZ
```

Kuniwak:
@aengelberg

Hi, how do I do to request to review this pull request?  (I mentioned you because of you seems recent active committer)

Kuniwak:
@aengelberg @gmemstr @polymeris @sulami

Hi, we are facing the problem in our CircleCI Enterprise server, so we are strongly hoping to fix it as soon as possible...

aengelberg (approved):
Thanks for the PR, and sorry for the delay on reviewing. This seems like a good change. It potentially affects existing users if they were utilizing the comma splitting behavior intentionally, but in that case it's easy to switch to passing in multiple `--env` flags.

PR #444, opened 2020-07-21
Escape `<<` in `<<include(file)>>` directives

gmemstr:
This allows the `<<include(file)>>` directives to escape any `<<` strings in the script itself, encouraging orb authors to write test-able scripts rather than injecting parameters directly into their scripts. e.g

```bash
# src/scripts/xyz.sh
echo << parameters.xyz >>
```

becomes

```bash
# rest of the orb
commands: |
  echo \<< parameters.xyz >>
```

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -1,0 +1,33 @@
+package process
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// IncludeFile replaces intsances of <<include(file)>> with the contents
+// of "file", escaping instances of "<<" within the file before returning.
+func IncludeFile(template string, root string) (string, error) {
+	// View: https://regexr.com/582gb
+	includeRegex, err := regexp.Compile(`(?U)^<<\s*include\((.*\/*[^\/]+)\)\s*?>>$`)
+	if err != nil {
+		return template, err
+	}
+
+	includeMatches := includeRegex.FindStringSubmatch(template)
+	if len(includeMatches) > 0 {
+		filepath := filepath.Join(root, includeMatches[1])
+		file, err := ioutil.ReadFile(filepath)
+		if err != nil {
+			return "", fmt.Errorf("could not open %s for inclusion", filepath)
+		}
+		escaped := strings.ReplaceAll("<<", string(file), "\\<<")
+
+		return escaped, nil
+	}
+
+	return template, nil
+}
--- a/process/process.go
+++ b/process/process.go
@@ -24,7 +24,7 @@ func IncludeFile(template string, root string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("could not open %s for inclusion", filepath)
 		}
-		escaped := strings.ReplaceAll("<<", string(file), "\\<<")
+		escaped := strings.ReplaceAll(string(file), "<<", "\\<<")
 
 		return escaped, nil
 	}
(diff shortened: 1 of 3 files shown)


marcomorain:
How can someone test a bash script that contains a line like `<< parameters.xyz >>`?

I would expect a script to take parameters as arguments, rather than using moustache templates.

gmemstr:
@marcomorain This PR is to enforce that, and to encourage authors to write tests for their scripts :)

mkly:
I'm not 100% sure of the requirements of `include` but some things that didn't work that I expected to work(possibly mistakenly)

I felt like these should work
```
<< include (scripts/x.sh) >>
<< include(scr.sh) >>
```

And then ones that work that I'm not sure should
```
<< include(scripts/x.sh)) >>
```

```
<< include(test/x.sh))))) >>))>>
```

```
<<

include(scripts/x.sh)>>
text
test

))>>
```

Not sure if that is relevant specifically for this PR and was just based on the regex site linked. Not sure if these all translate directly. Apologies if not relevant. Here's a fork

https://regexr.com/58t6p

gmemstr:
@mkly I think that's a little out of scope of this particular ticket, but in the same component. I believe this behaviour is intended, but maybe @KyleTryon can comment on this.

process/process.go
```diff
@@ -0,0 +1,33 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// IncludeFile replaces intsances of <<include(file)>> with the contents
+// of "file", escaping instances of "<<" within the file before returning.
+func IncludeFile(template string, root string) (string, error) {
+	// View: https://regexr.com/582gb
+	includeRegex, err := regexp.Compile(`(?U)^<<\s*include\((.*\/*[^\/]+)\)\s*?>>$`)
```
johnswanson, line 15: Is the intent here that the entire YAML node must be exactly one `include` stanza? That seems reasonable for scripts, but maybe there are cases where someone will want to do something different. If we do go with that, we should handle user error gracefully.

Cases I'd think about:

- `sh -c '<< include(script) >>'`
- gracefully handling multiple includes in one string?

Currently the first one will be silently ignored and the second one will get an error message like

```
Error: An unexpected error occurred: could not open /home/jds/src/slack-orb/src/scripts/notify.sh) >> << include(scripts/notify.sh for inclusion
```

johnswanson: (Sorry I didn't notice this before, since it isn't actually a new part of this PR)

process/process.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// IncludeFile replaces intsances of <<include(file)>> with the contents
+// of "file", escaping instances of "<<" within the file before returning.
+func IncludeFile(template string, root string) (string, error) {
```
johnswanson, line 13: Naming nits:

`IncludeFile` -> `MaybeIncludeFile` (since 99% of the time it's going to be called with a string that doesn't have an `include` statement

`template` -> ... not sure. `s` is a lot vaguer but I think less confusing than `template` because it is definitely a string but not necessarily a template. `maybeTemplate`? Not sure.

`root` -> `orbDirectory` or `orbRootDirectory` or something like that - I think this would be more clear

process/process.go
gmemstr: This is as intended, there has been a fair amount of dicsussion around this and the conclusion is this is how it is intended to function. Although the error is something we probably want to eliminate.

johnswanson: Feel free to merge as is but I definitely agree that we should fix the errors.

PR #446, opened 2020-07-24
[CIRCLE-24330] Use the REST context API for listing

johnswanson:
This is an attempt to shoot a narrow thread through the CLI, as a POC of
using the REST API instead of the GraphQL API.

- I added a `rest_client` package, 99% of which is autogenerated using
https://github.com/deepmap/oapi-codegen. I added a few helper functions
and defined types (like `Context`) in the `client.go` file in that
package. I'm not entirely sure this is the best way to do things, but it
seems to work pretty well.

- I updated the modules that pull together the config so that they use a
new `rest_client` setting in the cli.yaml. The default value is
`https://circleci.com/api/v2`.

- I changed the context module to initialize a rest API client and use
that client when listing contexts.


johnswanson:
closing in favor of #447

PR #453, opened 2020-08-12
Complain when an `<< include ... >> ` has multiple

johnswanson:
Before we were failing silently or giving a confusing error message -
this gives the user a more useful error message.


gmemstr:
-- Don't merge yet, planning a larger PR with a few additional changes to `orb init` --

PR #460, opened 2020-09-07
Orb init

gmemstr:
## Demos
First time setup: https://asciinema.org/a/65eIr2T0x7g8Mz9tgYHEIHnFy
Already gone through flow:  https://asciinema.org/a/g2NbHGaEHE9utIQDmdil75YBh


api/api.go
```diff
@@ -1377,3 +1383,29 @@ func ListOrbCategories(cl *graphql.Client) (*OrbCategoriesForListing, error) {
 	}
 	return &orbCategories, nil
 }
+
+// FollowProject initiates an API request to follow a specific project on
+// CircleCI. Project slugs are case-sensitive.
+func FollowProject(vcs string, owner string, projectName string, cciToken string) (FollowedProject, error) {
+	requestPath := fmt.Sprintf("https://circleci.com/api/v1.1/project/%s/%s/%s/follow", vcs, owner, projectName)
```
johnswanson, line 1390: We shouldn't hardcode the endpoint here - right now it's not a huge deal because Server customers can't use orbs, but this is going to change soon. We can use `Config.RestEndpoint` as the host.

It might also be good to move this to its own file, maybe `api_rest.go`, even if it's in the same package.

johnswanson: One option, if you wanted, would be to refactor a bit: `context_rest.go` defines a `ContextRestClient` type and functions to create it, but there's actually nothing contexts specific about it. We could change that type's name to `RestClient` and then define additional methods on it to allow it to follow projects too.

That might be too much refactoring for right now though, up to you.

johnswanson (changes requested):
A small fix, otherwise LGTM

api/api.go
gmemstr: Too much refactoring for this right now, but definitely something I've noted down to tackle soon.

johnswanson (approved):

KyleTryon:
Have tested. Approved.

PR #461, opened 2020-09-09
Bump JSON version for brew outdated

sgerrand:
Prevents errors from being raised when attempting to use a deprecated API version with Homebrew.

    $ brew outdated --json=v1 ; echo $?
    Error: Calling brew outdated --json=v1 is deprecated! Use brew outdated --json=v2 instead.
    1


aengelberg (approved):
Thanks for the change.

PR #474, opened 2020-09-30
[CIRCLE-29501] Add commands to manage runner types

pete-woods:
Commands added:
- `circleci runner resource-class create <resource-class> <description>`
- `circleci runner resource-class delete <resource-class>`
- `circleci runner resource-class list <namespace>`
- `circleci runner token create [--config] <resource-class> <nickname>`
- `circleci runner token delete <token-id>`
- `circleci runner token list <resource-class>`

Aliases of `delete` to `rm` and `list` to `ls` have been added.

### DIFF
--- a/api/runner/runner.go
+++ b/api/runner/runner.go
@@ -1,0 +1,116 @@
+package runner
+
+import (
+	"net/url"
+	"time"
+
+	"github.com/CircleCI-Public/circleci-cli/api/rest"
+)
+
+type Runner struct {
+	rc *rest.Client
+}
+
+func New(rc *rest.Client) *Runner {
+	return &Runner{rc: rc}
+}
+
+type ResourceClass struct {
+	ID            string `json:"id"`
+	ResourceClass string `json:"resource_class"`
+	Description   string `json:"description"`
+}
+
+func (r *Runner) CreateResourceClass(resourceClass, desc string) (rc *ResourceClass, err error) {
+	req, err := r.rc.NewRequest("POST", &url.URL{Path: "runner/resource"}, struct {
+		ResourceClass string `json:"resource_class"`
+		Description   string `json:"description"`
+	}{
+		ResourceClass: resourceClass,
+		Description:   desc,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	rc = &ResourceClass{}
+	_, err = r.rc.DoRequest(req, rc)
+	return rc, err
+}
+
+func (r *Runner) GetResourceClassesByNamespace(namespace string) ([]ResourceClass, error) {
+	query := url.Values{}
+	query.Add("namespace", namespace)
+	req, err := r.rc.NewRequest("GET", &url.URL{Path: "runner/resource", RawQuery: query.Encode()}, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resp := struct {
+		Items []ResourceClass `json:"items"`
+	}{}
+	_, err = r.rc.DoRequest(req, &resp)
+	return resp.Items, err
+}
+
+func (r *Runner) DeleteResourceClass(id string) error {
+	req, err := r.rc.NewRequest("DELETE", &url.URL{Path: "runner/resource/" + url.QueryEscape(id)}, nil)
+	if err != nil {
+		return err
+	}
+
+	_, err = r.rc.DoRequest(req, nil)
+	return err
+}
+
+type Token struct {
+	ID            string    `json:"id"`
+	Token         string    `json:"token"`
+	ResourceClass string    `json:"resource_class"`
+	Nickname      string    `json:"nickname"`
+	CreatedAt     time.Time `json:"created_at"`
+}
+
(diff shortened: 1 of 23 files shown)


api/runner/runner.go
```diff
@@ -0,0 +1,150 @@
...
+
+	resp := struct {
+		Items []ResourceClass `json:"items"`
+	}{}
+	_, err = r.rc.DoRequest(req, &resp)
+	return resp.Items, err
+}
+
+func (r *Runner) DeleteResourceClass(id string) error {
+	req, err := r.rc.NewRequest("DELETE", &url.URL{Path: fmt.Sprintf("runner/resource/%s", id)}, nil)
```
glenjamin, line 61: `id` should be URL escaped

api/runner/runner.go
```diff
@@ -0,0 +1,150 @@
...
+
+type Token struct {
+	ID            string    `json:"id"`
+	Token         string    `json:"token"`
+	ResourceClass string    `json:"resource_class"`
+	Nickname      string    `json:"nickname"`
+	CreatedAt     time.Time `json:"created_at"`
+}
+
+func (t *Token) ToAgentConfig() *AgentConfig {
```
glenjamin, line 78: I think this should be outside of the `API` package

api/runner/runner.go
```diff
@@ -0,0 +1,150 @@
...
+
+	resp := struct {
+		Items []Token `json:"items"`
+	}{}
+	_, err = r.rc.DoRequest(req, &resp)
+	return resp.Items, err
+}
+
+func (r *Runner) DeleteToken(id string) error {
+	req, err := r.rc.NewRequest("DELETE", &url.URL{Path: fmt.Sprintf("runner/token/%s", id)}, nil)
```
glenjamin, line 125: `id` should be url-escaped

cmd/root_test.go
```diff
@@ -15,7 +15,7 @@ var _ = Describe("Root", func() {
 	Describe("subcommands", func() {
 		It("can create commands", func() {
 			commands := cmd.MakeCommands()
-			Expect(len(commands.Commands())).To(Equal(17))
+			Expect(len(commands.Commands())).To(Equal(18))
```
glenjamin, line 18: maybe replace this with a golden test on `--help`?

cmd/runner/resource_class.go
```diff
@@ -0,0 +1,71 @@
...
+		Aliases: []string{"ls"},
+		Args:    cobra.ExactArgs(1),
+		PreRunE: preRunE,
+		RunE: func(_ *cobra.Command, args []string) error {
+			rcs, err := r.GetResourceClassesByNamespace(args[0])
+			if err != nil {
+				return err
+			}
+
+			table := tablewriter.NewWriter(os.Stdout)
```
glenjamin, line 59: ~ maybe extract and share the table with create?

cmd/runner/token.go
```diff
@@ -0,0 +1,81 @@
...
+		Short:   "Create a token for a resource-class",
+		Args:    cobra.ExactArgs(2),
+		PreRunE: preRunE,
+		RunE: func(_ *cobra.Command, args []string) error {
+			token, err := r.CreateToken(args[0], args[1])
+			if err != nil {
+				return err
+			}
+
+			if createOpts.config {
```
glenjamin, line 33: This is a great idea!

Maybe it should be on by default?

cmd/runner/token.go
```diff
@@ -0,0 +1,81 @@
...
+		Short:   "List tokens for a resource-class",
+		Args:    cobra.ExactArgs(1),
+		PreRunE: preRunE,
+		RunE: func(_ *cobra.Command, args []string) error {
+			tokens, err := r.GetRunnerTokensByResourceClass(args[0])
+			if err != nil {
+				return err
+			}
+
+			table := tablewriter.NewWriter(os.Stdout)
```
glenjamin, line 70: ~ same comment as resource class about sharing the table helper

cmd/root_test.go
pete-woods: Not sure I wanna start re-doing all of the CLI's tests just now!

cmd/runner/token.go
pete-woods: Can't do this one, as they are different (one contains the token value, the other does not)

cmd/runner/token.go
pete-woods: 🤔

cmd/root_test.go
pete-woods: I've put in a bunch of usage test assertions for the runner commands.

Once this merges, I'll have a look at how hard it'd be to do the same for all the commands in the CLI.

danmux (approved):

cmd/root_test.go
glenjamin: I was thinking just one that lists all of the commands - but I had forgotten that there isn't a help command which lists all commands, and these are hidden anyway!

pete-woods: Good point!

marcomorain (approved):

PR #479, opened 2020-10-05
Send proper user agent, and delete resource-classes by name

pete-woods:
- Path escape runner IDs and send proper user agent
- Delete resource-classes by name, not ID

### DIFF
--- a/api/rest/client.go
+++ b/api/rest/client.go
@@ -10,6 +10,8 @@ import (
 	"net/url"
 	"strings"
 	"time"
+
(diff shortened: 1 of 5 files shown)


glenjamin (approved):

glenjamin:
Is this what was causing the 500?

pete-woods:
> Is this what was causing the 500?

Nope - this was just something I spotted us doing elsewhere.

It's now been updated to add the changed `rm` behaviour, too.

PR #494, opened 2020-10-16
CIRCLE-29574: add a remove-alias command

johnswanson:
Add a CLI command to delete a namespace alias

### DIFF
--- a/cmd/namespace.go
+++ b/cmd/namespace.go
@@ -54,6 +54,35 @@ func newNamespaceCommand(config *settings.Config) *cobra.Command {
 		Short: "Operate on namespaces",
 	}
 
+	deleteAliasCmd := &cobra.Command{
+		Use:   "delete-alias <name>",
+		Short: "(Server only) Delete a namespace alias",
+		Long: `Delete a namespace alias.
+
+A namespace can have multiple aliases (names). This command deletes an alias left behind by a rename. The most recent alias cannot be deleted.
+
+Example:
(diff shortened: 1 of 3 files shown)


cmd/namespace.go
```diff
@@ -54,6 +54,35 @@ func newNamespaceCommand(config *settings.Config) *cobra.Command {
...
+		Use:   "delete-alias <name>",
+		Short: "(Server only) Delete a namespace alias",
+		Long: `Delete a namespace alias.
+
+A namespace can have multiple aliases. This command deletes an alias. The current alias cannot be deleted.
+
+Example:
+- namespace A is renamed to B
+- alias A->B is automatically created
+- we can delete the A alias.
```
aengelberg, line 67: I think this description might be confusing, because it implies that aliases point at other aliases, whereas they actually all point to a common invisible id.

Here's an alternative wording, but not sure if it's clearer:
```suggestion
A namespace can have multiple aliases (names). This command deletes an alias left behind by a rename. The most recent alias cannot be deleted.

Example:
- namespace A is renamed to B
- alias B is created, coexisting with alias A
- after migrating config accordingly, we can delete the A alias.
```

aengelberg (approved):
I had some thoughts on the docstring for this command, but otherwise LGTM.

cmd/namespace.go
skimke: Thanks, much more detailed!

PR #493, opened 2020-10-14
Fix `orb init` error `invalid cross device link`

mkly:
Fix `orb init` error `invalid cross device link` when run with snap

Rename will cause an error with snap. This instead removed the `rename`
call and instead does not write the top level directory

* Remove `os.Rename` in favor of unzipping directly to orb directory.
* Change `unzip` function to `unzipToOrbPath` and modify function to
  specifically handly orb use-case as opposed to being general purpose

Fixes #491

PR #496, opened 2020-10-20
`orb init` improvements

gmemstr:
Brings #493, #453, and snap detection when creating a git commit, under one PR.

Snap detection is necessary because of the way snaps are sandboxed. Normal installations allow the `git` library used to read the configuration and pull details, whereas a snap install disallows this and requires us to supply custom details.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -15,6 +15,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/CircleCI-Public/circleci-cli/api"
 	"github.com/CircleCI-Public/circleci-cli/api/graphql"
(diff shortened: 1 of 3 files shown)


cmd/orb.go
```diff
@@ -1276,9 +1272,24 @@ func initOrb(opts orbOptions) error {
...
-	if err != nil {
-		return err
+
+	if version.PackageManager() != "snap" {
+		_, err = w.Commit("[semver:skip] Initial commit.", &git.CommitOptions{})
+		if err != nil {
+			return err
+		}
+	} else {
+		fmt.Println("We detected you installed the CLI via snap\nThe commit generated will not match your actual git username or email due to sandboxing.")
```
KyleTryon, line 1282: Seems fair/good to me

KyleTryon (approved):
Looks like great improvements 👍

KyleTryon:
CC @gmemstr @johnswanson  I think we need a revert here. Just got this error on CCI that i had not been getting previously: https://app.circleci.com/pipelines/github/CircleCI-Public/aws-elastic-beanstalk-orb/72/workflows/b6dc5cd1-2352-409a-b6b6-940f68c51b85/jobs/292

```
Error: Error calling command: 'setup'
Unclosed '<<' tag in string: '<<include(scripts/install-setup.sh)>>'
('<<' must be escaped in config v2.1+)
Error calling job: 'deploy'
Error calling command: 'setup'
Unclosed '<<' tag in string: '<<include(scripts/install-setup.sh)>>'
('<<' must be escaped in config v2.1+)

```

Even shown in that error message it shows the tag is closed.

PR #505, opened 2020-10-23
[CIRCLE-30416] Move Server-specific administrative commands into new "admin" category

aengelberg:
This renames the following commands:
* `circleci orb import` => `circleci admin import-orb`
* `circleci namespace rename` => `circleci admin rename-namespace`
* `circleci namespace delete-alias` => `circleci admin delete-namespace-alias`


kelvinkfli (approved):
LGTM! I originally separated the import tests into its own file, do you think it's worth moving those into `admin_test.go` as well?

aengelberg:
@kelvinkfli I think I'd rather `orb_import_test.go` still be separate, to mirror `orb_import.go` which is still separate.

Issue #501, opened 2020-10-22
Arrow keys don't work on Windows 10


ThatCopy:
## Meta:

**CircleCI CLI Version:**

0.1.11146+275feba (release)
**Operating System:**

Windows 10 2004
**CircleCI CLI Diagnostic:**

```
---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: C:\Users\David\.circleci\cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API...
Ok.
Hello, David.
```

---

## Current behavior:

When I run `circleci orb init ./test/` I cant select a different option. The arrow keys don't work, the enter key works. On WSL2 it works.

## When did this begin / Was this previously working?:
From release v0.1.10993 onwards 

## Additional Information:
![screenshot](https://i.imgur.com/d13WZUm.png)
When I press a arrow key nothing happens

gmemstr:
This appears to be a bug with the upstream library we're using: https://github.com/manifoldco/promptui/issues/33.

seongwon-kang:
Key j, k works.

gmemstr:
It's an okay workaround, but after some digging into that aforementioned issue, I think a better solution is to switch to https://github.com/AlecAivazis/survey - it appears that it will work on windows and offer more functionality that we can build upon, e.g for a more specific pipeline configuration.

ThatCopy:
Another issue that I found is that entering any text on windows is really slow. So I also think that switching to another lib is a better solution

PR #506, opened 2020-10-26
Move from "promptui" to "survey" package.

gmemstr:
"promptui" has some incompatibilities with Windows, which makes navigating menus much harder (see #501) - [survey](https://github.com/AlecAivazis/survey) appears to resolve these, so this PR completely moves over to survey instead. Survey also offers a number of other prompt types natively.


aengelberg (approved):
LGTM

PR #516, opened 2020-11-02
Properly deserialize graphQL response when creating an imported orb

kelvinkfli:
This PR fixes a bug where we weren't deserializing graphQL responses properly when importing orbs. The response struct name was changed to match that of the mutation being performed: `ImportOrb`

### DIFF
--- a/cmd/orb_import_test.go
+++ b/cmd/orb_import_test.go
@@ -799,7 +799,7 @@ The following orb versions already exist:
 					}
 				}`
 
-			createOrbReq := `{
+			importOrbReq := `{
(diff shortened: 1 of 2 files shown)


johnswanson (approved):

cmd/orb_import_test.go
```diff
@@ -808,7 +808,7 @@ The following orb versions already exist:
 				  }`
 
 			createOrbResp := `{
-					"createOrb": {
```
johnswanson, line 811: ~~ the var name `createOrbResp` could get changed to `importOrbResp`

kelvinkfli: Thanks for the catch!

PR #517, opened 2020-11-05
Fix bug listing env vars

johnswanson:
This caused only the first 20 environment variables to be printed,
rather than paging through if there were more.


kelvinkfli (approved):
LGTM!

aengelberg (approved):

PR #520, opened 2020-11-18
[CIRCLE-31080] Add no-prompt option to admin commands

cayennes:
Add --no-prompt option to admin commands to import orbs and delete namespace
aliases. We need these for use with automated integation tests of a server
install.

### DIFF
--- a/cmd/admin.go
+++ b/cmd/admin.go
@@ -65,6 +65,11 @@ Example:
 			return validateToken(nsOpts.cfg)
 		},
 		RunE: func(_ *cobra.Command, _ []string) error {
+			if nsOpts.integrationTesting {
+				nsOpts.tty = createNamespaceTestUI{
+					confirm: true,
+				}
+			}
 			return deleteNamespaceAlias(nsOpts)
 		},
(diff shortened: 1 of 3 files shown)


cmd/admin.go
```diff
@@ -68,6 +72,7 @@ Example:
 	}
 
 	deleteAliasCommand.Annotations["<name>"] = "The name of the alias to delete"
+	deleteAliasCommand.Flags().BoolVar(&nsOpts.noPrompt, "no-prompt", false, "Disable prompt to bypass interactive UI.")
```
kelvinkfli, line 75: I just realized that the [deleteNamespaceAlias](https://github.com/CircleCI-Public/circleci-cli/blob/master/cmd/namespace.go#L98) function doesn't actually do a `if opts.NoPrompt { ... }` check. Do you think theres value in adding it for non-integration test purposes? 🤔

cayennes: Yes, thanks for noticing that. Done.

kelvinkfli (approved):

PR #543, opened 2021-01-21
[CIRCLE-32386] OrbPublish and OrbPromote now function through an orb's name + namespace instead of its ID

kelvinkfli:
This PR includes the following changes:
- Adds `OrbPublishByName` and `OrbPromoteByName` functions
- Removes `OrbPublishByID` and `OrbPromoteByID` functions
- Removes unnecessary `OrbId` calls

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -625,8 +625,9 @@ func OrbImportVersion(cl *graphql.Client, orbSrc string, orbID string, orbVersio
 	return &response.ImportOrbVersion.Orb, nil
 }
 
-// OrbPublishByID publishes a new version of an orb by id
-func OrbPublishByID(cl *graphql.Client, configPath string, orbID string, orbVersion string) (*Orb, error) {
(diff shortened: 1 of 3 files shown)


api/api.go
```diff
@@ -635,9 +636,10 @@ func OrbPublishByID(cl *graphql.Client, configPath string, orbID string, orbVers
 	}
 
 	query := `
-		mutation($config: String!, $orbId: UUID!, $version: String!) {
+		mutation($config: String!, $orbName: String!, $namespaceName: String!, $version: String!) {
```
skimke, line 639: From the api-service GraphQL schema, it seemed these two were nullable (`version` definitely is non nullable). Should these be `String` and not `String!`?
Same question for https://github.com/CircleCI-Public/circleci-cli/pull/543/files#diff-bf98d5fab5bcfabded9769069e7318683f549ec91cff84c5f87a9ae3a3ca3d8dR1190

skimke (approved):
had one question for you about the mutations

PR #545, opened 2021-01-28
Listing Orbs now optionally accepts the '--private' flag

kelvinkfli:
This PR allows CLI users to return private orbs within a given namespace by providing the `--private` flag. When this flag is set, the user must also provide a namespace, otherwise an error will be returned.

Co-authored-by: @ryanbarkley <ryanbarkley@users.noreply.github.com>

### DIFF
--- a/cmd/orb_test.go
+++ b/cmd/orb_test.go
@@ -1990,11 +1990,11 @@ Search, filter, and view sources for all Orbs online at https://circleci.com/dev
 				By("setting up a mock server")
 
 				query := `
-query namespaceOrbs ($namespace: String, $after: String!) {
+query namespaceOrbs ($namespace: String, $after: String!, $view: OrbListViewType) {
 	registryNamespace(name: $namespace) {
 		name
                 id
-		orbs(first: 20, after: $after) {
+		orbs(first: 20, after: $after, view: $view) {
 			edges {
 				cursor
 				node {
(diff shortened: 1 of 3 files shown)


skimke (approved):

cmd/orb_test.go
```diff
@@ -2158,7 +2161,104 @@ query namespaceOrbs ($namespace: String, $after: String!) {
...
+				Expect(string(stdout)).To(Equal(`Orbs found: 5. Showing only certified orbs.
+Add --uncertified for a list of all orbs.
+
+circleci/delete-me (Not published)
+circleci/delete-me-too (Not published)
+circleci/gradle (0.0.1)
+circleci/heroku (Not published)
+circleci/rollbar (0.0.1)
+
+In order to see more details about each orb, type: ` + "`circleci orb info orb-namespace/orb-name`" + `
```
skimke, line 2256: I'm a little worried that they may run this on private orbs :/ but I'm also not sure that we want to introduce temporary messaging.

cmd/orb_test.go
```diff
@@ -2158,7 +2161,104 @@ query namespaceOrbs ($namespace: String, $after: String!) {
...
+
+Search, filter, and view sources for all Orbs online at https://circleci.com/developer/orbs/
```
skimke, line 2258: This part would benefit from saying "except private orbs"

skimke (approved):
LGTM - a few questions for you on the messages returned when successfully listing orbs

cmd/orb_test.go
kelvinkfli: I agree. I re-wrote the existing function responsible for constructing this response, let me know what you think!

cmd/orb_test.go
kelvinkfli: In the new formatting function, I opted to just not prompt users (until we've reached parity) to take any action and eliminate these lines when `--private` is provided. Any thoughts?

skimke: Make sense, thanks for updating!

PR #548, opened 2021-02-02
[CIRCLE-32051] CLI Commands now optionally accept --org-slug flags

kelvinkfli:
This PR completes the functionality provided by: https://github.com/circleci/api-service/pull/793

For configs that use private orbs, we now require that users provide an org-slug for the following commands:
1. `circleci config validate`
2. `circleci config process`
3. `circleci local execute`

Co-authored-by: @cayennes

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -513,28 +513,44 @@ func WhoamiQuery(cl *graphql.Client) (*WhoamiResponse, error) {
 }
 
 // ConfigQuery calls the GQL API to validate and process config
-func ConfigQuery(cl *graphql.Client, configPath string, pipelineValues pipeline.Values) (*ConfigResponse, error) {
+func ConfigQuery(cl *graphql.Client, configPath string, orgSlug string, pipelineValues pipeline.Values) (*ConfigResponse, error) {
 	var response BuildConfigResponse
+	var query string
 
 	config, err := loadYaml(configPath)
 	if err != nil {
 		return nil, err
 	}
 
-	query := `
-		query ValidateConfig ($config: String!, $pipelineValues: [StringKeyVal!]) {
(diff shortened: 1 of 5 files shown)


api/api.go
```diff
@@ -513,29 +513,43 @@ func WhoamiQuery(cl *graphql.Client) (*WhoamiResponse, error) {
...
-		query ValidateConfig ($config: String!, $pipelineValues: [StringKeyVal!]) {
-			buildConfig(configYaml: $config, pipelineValues: $pipelineValues) {
-				valid,
-				errors { message },
-				sourceYaml,
-				outputYaml
-			}
-		}`
+	if orgSlug != "" {
+		query = `query ValidateConfig ($config: String!, $pipelineValues: [StringKeyVal!], $orgSlug: String!) {
```
skimke, line 526: Is `orgSlug` an optional `String`, according to the schema?

cmd/config_test.go
```diff
@@ -139,5 +145,185 @@ var _ = Describe("Config", func() {
...
+						valid,
+						errors { message },
+						sourceYaml,
+						outputYaml
+					}
+				}`
+
+				r := graphql.NewRequest(query)
+				r.Variables["config"] = config
+				r.Variables["pipelineValues"] = pipeline.PrepareForGraphQL(pipeline.FabricatedValues())
```
skimke, line 178: Bit of a nitpick - wdyt about this being a var set with `BeforeEach` as well, since we're doing that for `config` too? Would `FabricatedValues` return something different each time?

skimke (approved):

api/api.go
kelvinkfli: Good catch, it is!

PR #550, opened 2021-02-05
Move chocolatey scripts to main repository

gmemstr:
Moving files from https://github.com/CircleCI-Public/chocolatey-circleci-cli to here to remove overhead of calling other project pipeline. We do loose out on tagging individual versions, but I'm not sure that's a huge deal given it's a relatively static file.


.circleci/config.yml
```diff
@@ -294,19 +294,21 @@ jobs:
...
-              -H 'Content-Type: application/json' \
-              -H 'Accept: application/json' \
-              -H "Circle-Token: $CIRCLE_TOKEN" \
-              --data '{ "parameters": { "deploy-production": true } }'
+          name: Run update script
+          command: .\update-version.ps1
+      - run: choco pack circleci-cli\circleci-cli.nuspec
+      - run: copy-item circleci-cli.*.nupkg circleci-cli.nupkg
+      - store_artifacts:
+          path: circleci-cli.nupkg
```
kelvinkfli, line 308: Out of curiosity, do we use this build artifact anywhere outside of the job?

kelvinkfli (approved):
LGTM

.circleci/config.yml
gmemstr: No, this is just helpful for examination - I think we can safely remove it though :)

PR #552, opened 2021-02-05
Make the orb init command respect the --private flag

kelvinkfli:
In our [docs](https://circleci.com/docs/2.0/orb-author/), we recommend that new users use the automated orb authoring process to author orbs.

This PR adds support so that `orb init`can be used (in addition to `orb create`) to author private orbs.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -317,6 +317,7 @@ Please note that at this time all orbs created in the registry are world-readabl
 		},
 		Args: cobra.ExactArgs(1),
 	}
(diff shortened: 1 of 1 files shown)


skimke (commented):
These changes LGTM!
Could we add a test case for accepting the addition `opts`?
Also, the confirmation message might need changing for private orbs in the [`finalizeOrbInit`](https://github.com/CircleCI-Public/circleci-cli/pull/552/files#diff-6f0d687500dbb1f4262b6a551c3be48e7164da158b71b0e62ec4ee813399899dR1376) function

skimke (approved):

PR #580, opened 2021-02-24
resolve CVE-2020-15114 in etcd v3.3.10+incompatible

bhamail:
While creating a PR to add a [nancy](https://github.com/sonatype-nexus-community/nancy) scan to CI, I ran into the vulnerability below (CVE-2020-15114):

https://ossindex.sonatype.org/vuln/bba60acb-c7b5-4621-af69-f4085a8301d0

This PR bumps the version of etcd to a newer, non-vulnerable version.

The version previously in use was:
```shell
$ go list -m all | grep 'github.com/coreos/etcd'
github.com/coreos/etcd v3.3.10+incompatible
```

relates to PR #581 add nancy scan to CI

PR #581, opened 2021-02-24
add nancy scan to CI

bhamail:
This PR adds a job to run a scan using [nancy](https://github.com/sonatype-nexus-community/nancy)

possibly depends on PR #580 resolve CVE-2020-15114 in etcd v3.3.10+incompatible

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,7 @@ orbs:
   shellcheck: circleci/shellcheck@1.2.0
   slack: circleci/slack@0.1.14
   windows: circleci/windows@2.2.0
+  nancy: sonatype-nexus-community/circleci-nancy-orb@0
 
 executors:
   go:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,6 +5,7 @@ orbs:
   shellcheck: circleci/shellcheck@1.2.0
   slack: circleci/slack@0.1.14
   windows: circleci/windows@2.2.0
+  nancy: sonatype-nexus-community/circleci-nancy-orb@0
```
gmemstr, line 8: The tagging of the orb version makes me a little nervous. Since there is no 1.0 version (is this in the works?), we should absolutely lock this to the patch version.

taxonomic-blackfish:
Hi @bhamail. We're going to close this as we have internal services that scan our repos. There may be something out of sync between the different scanning services. It's not in the CI config yet, but that's on our roadmap. Thanks for the PR, but we're going to close this.

PR #583, opened 2021-02-25
[CIRCLE-33115] Check if orb published/promoted is private

skimke:
Makes an additional call to fetch the orb when publishing or promoting an orb version to restore the messages on open source orbs removed with https://github.com/CircleCI-Public/circleci-cli/pull/573.

Merge after https://github.com/circleci/api-service/pull/803

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -490,6 +490,20 @@ func orbToSimpleString(orb api.OrbWithData) string {
 	return buffer.String()
 }
 
+func orbIsOpenSource(cl *graphql.Client, namespace string, orbName string) (bool) {
+	// TODO: make a separate api func to query for the orb's isPrivate field
+	// when available in the schema
+	orbExists, err := api.OrbExists(cl, namespace, orbName)
+
+	// we can ignore any errors with the call
+	if err != nil {
+		return false
+	}
+
+	// orbExists will also be false if it is a private orb
+	return orbExists
+}
+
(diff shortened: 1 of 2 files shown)


cmd/orb.go
```diff
@@ -673,6 +673,17 @@ func publishOrb(opts orbOptions) error {
 		fmt.Printf("Note that your dev label `%s` can be overwritten by anyone in your organization.\n", version)
 		fmt.Printf("Your dev orb will expire in 90 days unless a new version is published on the label `%s`.\n", version)
 	}
+
+	orbResponse, err := api.OrbIsPrivate(opts.cl, namespace, orb)
+
+	if err != nil {
+		return nil
```
kelvinkfli, line 680: Non-blocking thought: since we have an expectation that calling `api.OrbIsPrivate` with a private orb will actually return an error, what are your thoughts on wrapping this call in a function like: 
```go
func orbIsPrivate(cl *graphql.Client, namespace, orb string) bool {
    ...
    if err != nil {
        // maybe a comment explaining why an error here means
        // the orb either doesn't exist or is private
        return true
    }
    ...
}
```
This would be one way to avoid the unidiomatic `if err != nil { return nil }`.

kelvinkfli (approved):

cmd/orb.go
skimke: I wonder if we might want to make the existing api.OrbIsPrivate func return boolean instead, instead of adding more abstraction 🤔

kelvinkfli (approved):

PR #584, opened 2021-03-08
feat: allow keeping local exec containers

zepatrik:
When running a job locally, it would be nice to keep the containers in some case to do after-run analysis, just as you would with remote SSH. It seems that this is not working as I would like it to, because the picard image also needs to implement this feature. As I could not find it on GitHub, I assume it is a private repo. Can you add the feature there as well? This PR is more of a POC and I will improve it once it is technically possible to implement this.
Until then, I will unfortunately still have to use SSH on the remote.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -16,6 +16,7 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 	}
 
 	local.AddFlagsForDocumentation(buildCommand.Flags())
+	buildCommand.Flags().Bool("rm", true, "remove containers automatically")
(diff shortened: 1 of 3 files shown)


abdelDriowya:
@zepatrik thanks for your contribution.
Is there a news on this ? or should I close

JulesFaucherre:
Hey @zepatrik , we're actually considering another future for `circleci local execute` which has never really been stable. Know that we actually really take your proposition into consideration!
For this actual PR, know that we can not change the picard image local execute so I am afraid I'm gonna have to close this PR.
Thanks a lot!

PR #585, opened 2021-03-10
Runner subcommand respecting flags

kkowalski:
When instantiating runner and client structs the flags are not parsed yet, so client ends up with the default values. 
This roughly fallows how other commands initialize and share instances.

Also I'm making runner command not hidden.


danmux (approved):
lgtm

glenjamin (approved):

PR #586, opened 2021-03-11
Add CPEng as code owners for orb-related files.

gmemstr:
This will allow CPE to have some insight into changes to orb-related commands and functionality, especially regarding CPE owned commands such as `orb init`.


gmemstr:
I'm not 100% sure what the best way to validate this change is, but from what I can gather from here https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners this pattern _should_ work.

aengelberg (approved):

PR #590, opened 2021-04-16
Remove slack webhook configuration in CI

kelvinkfli:
We need to roll the `SLACK_WEBHOOK` environment variable. Because this webhook was configured with an old method, Bear and I couldn't detect exactly which of the 54 configurations the variable pointed to.

As as result, we are instead going to remove slack notification commands from all jobs and we'll add them back at
a later time if necessary.

https://circleci.slack.com/archives/C01ULMLG820/p1618590142465600

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,6 @@ version: 2.1
 orbs:
   codecov: codecov/codecov@1.1.3
   shellcheck: circleci/shellcheck@1.2.0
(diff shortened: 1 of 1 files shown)


sanajaved7 (approved):

aengelberg (approved):

aengelberg:
I think having working slack notifications for the deploy jobs would be useful someday, but LGTM.

PR #596, opened 2021-05-25
Allow users to specify self-signed certificates or ignore TLS verification

kelvinkfli:
Adds a method on `settings.Config` which establishes a HTTPClient with the correct TLS specifications and reasonable timeouts. 

The following values in config should now be supported:
- `tls_cert`, which designates the filepath of a self-signed certificate.
- `tls_insecure`, which allows users to skip TLS verification entirely

There is a lot of refactoring involved in this PR because we establish clients in various places throughout the codebase (and gofmt), but the important changes are in `settings.go`.


johnswanson (approved):
Very nice, I like the refactor!

johnswanson (approved):

Issue #601, opened 2021-06-18
`circle orb init` takes the `--private` parameter, but will sometimes create a public orb instead


mikebryant:
## Meta:

**CircleCI CLI Version:**

0.1.15338+850f9ac (homebrew)

**Operating System:**

Mac OSX

**CircleCI CLI Diagnostic:**

```

---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: /Users/mikebryant/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API...
Ok.
Hello, Mike Bryant.
```

---

## Current behavior:

We ran `circleci orb init --private .`

The resulting orb was not private, and has led to the exposure of confidential information

## Expected behavior:

It should create a private orb, as documented

## Additional Information:

The `createOrb` command passes the parameter appropriately: https://github.com/CircleCI-Public/circleci-cli/blob/850f9acdbf24e4222bd4b36035e20c182e7db385/cmd/orb.go#L826
If you choose not to set up your git project, the call in initOrb hardcodes private to false: https://github.com/CircleCI-Public/circleci-cli/blob/850f9acdbf24e4222bd4b36035e20c182e7db385/cmd/orb.go#L1259

gmemstr:
@mikebryant Seems some fixes were merged :) Would you mind updating your CLI and giving it another go?

Issue #599, opened 2021-06-07
`orb init` fails when not setting up a git project


aengelberg:
## Current behavior:
```
$ circleci orb init ./hello-world-orb
...
? Would you like to set up your git project? No
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/CircleCI-Public/circleci-cli/cmd.initOrb(0xc00072e300, 0xc00075fec0, 0xc000089160, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, 0x1a579c0, ...)
	/home/circleci/project/cmd/orb.go:1264 +0x3f81
github.com/CircleCI-Public/circleci-cli/cmd.newOrbCommand.func21(0xc000182a00, 0xc000089160, 0x1, 0x1, 0x0, 0x0)
	/home/circleci/project/cmd/orb.go:316 +0x67
github.com/spf13/cobra.(*Command).execute(0xc000182a00, 0xc0000890d0, 0x1, 0x1, 0xc000182a00, 0xc0000890d0)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0xc000767900, 0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
github.com/CircleCI-Public/circleci-cli/cmd.Execute()
	/home/circleci/project/cmd/root.go:37 +0x27
main.main()
	/home/circleci/project/main.go:9 +0x20
```

## Expected behavior:
Should successfully create the orb.

## When did this begin / Was this previously working?:
unknown

## Additional Information:
I am pretty sure these lines need to be fixed:
* https://github.com/CircleCI-Public/circleci-cli/blob/master/cmd/orb.go#L1264 - change this to a for-loop over all categories rather than use `opts.args[1]`
* https://github.com/CircleCI-Public/circleci-cli/blob/master/cmd/orb.go#L1260 - change `false` to `opts.private`
* when the orb is successfully created in this `if` branch, log some information rather than silently terminating

pete-woods:
The CLI has been rewritten from scratch for v1 (the `main` branch — `v0` is the legacy CLI that ships today and now takes critical fixes only).

[...]

PR #604, opened 2021-06-21
Fix for successfully creating a private orb with orb dev kit

elimt:
When a user uses the orb development kit to create an orb, it does not respect the private orb argument passed. This only happens when a user says `no` setting up a git project.
This addresses both https://github.com/CircleCI-Public/circleci-cli/issues/601 and https://github.com/CircleCI-Public/circleci-cli/issues/599

- Updating the params for `CreateOrb` to read the `--private` orb argument passed.
- There is also an issue where choosing categories breaks the flow. Adding a fix for that rather than reading `opts.args[1]` irrespective of the categories passed

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -1256,13 +1256,15 @@ func initOrb(opts orbOptions) error {
 	}()
 
 	if !gitAction {
-		_, err = api.CreateOrb(opts.cl, namespace, orbName, false)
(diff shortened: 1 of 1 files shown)


mikebryant (approved):

johnswanson (approved):
LGTM assuming tests pass!

PR #611, opened 2021-07-21
Don't panic on bad semver versions

johnswanson:
Homebrew versions may look like `0.1.1234_5` if there is a revision made
to the homebrew formula. This won't parse with semver.MustParse, so the
update check panics in this case.

This makes two changes. First, if we can't parse a version as semver,
return an error with normal go error handling, rather than using
`MustParse` and panicking. Second, change `rootCmdPreRun` to print
errors, then throw them away and continue along. This way, if the
check-for-upgrade piece of the CLI is broken, the CLI is still usable.

### DIFF
--- a/update/update.go
+++ b/update/update.go
@@ -69,6 +69,24 @@ func checkFromSource(check *Options) error {
 	return err
 }
 
+// Homebrew revisions get added to the version with an underscore.
+// So `1.2.3 revision 4` becomes `1.2.3_4`. This fails to parse as valid semver
+// version. We can work around this by replacing underscores with `-` to convert
+// the revision to a semver tag.
+// https://github.com/CircleCI-Public/circleci-cli/issues/610
+func ParseHomebrewVersion(homebrewVesion string) (semver.Version, error) {
+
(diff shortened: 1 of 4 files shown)


update/update.go
```diff
@@ -86,11 +91,23 @@ func checkFromHomebrew(check *Options) error {
 	for _, o := range outdated.Formulae {
 		if o.Name == "circleci" {
 			if len(o.InstalledVersions) > 0 {
-				check.Current = semver.MustParse(o.InstalledVersions[0])
+				// homebrew versions may have a revision number appended, like
+				// `1.2.3_<revision_number>`. This is not valid semver, but we
+				// can make it valid by replacing the underscore with a hyphen.
+				current, err := semver.Parse(strings.Replace(o.InstalledVersions[0], "_", "-"))
+				if err != nil {
+					return errors.Wrap(err, "failed to parse current version from `brew outdated --json=v2`")
```
marcomorain, line 99: Let's add the value of `o.InstalledVersions[0]` to this message.

marcomorain (commented):
I threw a commit on top so that we can test the parsing code. I didn't write a test though since I'm late for dinner!

marcomorain (approved):

PR #612, opened 2021-07-28
Update Snapcraft auth notice

felicianotech:
I've updated the creds. The snap is now publishing. I'll work with @johnswanson to get this to be handled by the proper team moving forward.


rlegan (approved):

JulesFaucherre:
The snapcraft has since been updated, we'll do a new PR to change that comment, thank you @felicianotech

PR #613, opened 2021-07-29
Check if the orb is private

makotom:
This PR implements a test to check whether the orb passed to `orbIsOpenSource` is private or not.
This will provide an effect to suppress a message saying `Please note that this is an open orb and is world-readable.` for private orbs.

Credits to @chisato-ygc for originally reporting the issue.


ganezasan (approved):
LGTM! 

It seems the test data isn't organizing well, but it's a another issue from this 🤣

PR #617, opened 2021-08-09
Add HTTP header indicating the command name being run

aengelberg:
When making calls to our API, we'll now include a header `Circleci-Cli-Command` set to the current command being called, with no args or flags (e.g. `circleci orb list`).

This gives us some more server-side usage data for commands like `circleci local execute` that make the same API calls as other commands. It does not add any usage data for non-API-based commands like `circleci open` and `circleci setup`.

### DIFF
--- a/api/graphql/client.go
+++ b/api/graphql/client.go
@@ -54,7 +54,10 @@ func NewRequest(query string) *Request {
 	}
 
 	request.Header.Set("User-Agent", version.UserAgent())
-	request.Header.Set("Circleci-Cli-Command", header.GetCommandStr())
+	commandStr := header.GetCommandStr()
+	if commandStr != "" {
+		request.Header.Set("Circleci-Cli-Command", commandStr)
+	}
 	return request
 }
 
--- a/api/graphql/client.go
+++ b/api/graphql/client.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -33,12 +35,29 @@ var rootTokenFromFlag string
...
+// for API headers.
+func CommandStr() string {
+	command := MakeCommands()
+	subCmd, _, err := command.Find(os.Args[1:])
+	if err != nil {
+		return ""
+	}
+	parentNames := []string{subCmd.Name()}
+	subCmd.VisitParents(func(parent *cobra.Command) {
+		parentNames = append([]string{parent.Name()}, parentNames...)
```
aengelberg, line 56: there's gotta be a better way to build up a slice of strings backwards, right?

-- a Go noob

api/graphql/client.go
```diff
@@ -53,6 +54,10 @@ func NewRequest(query string) *Request {
 	}
 
 	request.Header.Set("User-Agent", version.UserAgent())
+	commandStr := header.GetCommandStr()
+	if commandStr != "" {
+		request.Header.Set("Circleci-Cli-Command", commandStr)
```
marcomorain, line 59: Is setting a header to the empty string a problem? Can we remove the `if`?

marcomorain (approved):
LGTM, a couple of bikeshed questions.

PR #621, opened 2021-08-12
Add --pipeline-parameters flag to config process

aaronstillwell:
This PR attempts to add pipeline parameters to the `config process` command. Without these, it is only possible to retrieve a processed config with the parameter defaults used.

The command takes either raw JSON or a JSON filename if prepended with `@` e.g `{"param1": true}` or `@params.json`.


aengelberg (commented):
Looks good so far! A couple of thoughts on my first pass.

aaronstillwell:
@aengelberg I concur with both suggestions and have now implemented them. Please let me know any further comments or suggestions.

aengelberg (approved):
Thanks for making this change!

cmd/config.go
```diff
@@ -147,8 +149,31 @@ func validateConfig(opts configOptions, flags *pflag.FlagSet) error {
 
 func processConfig(opts configOptions, flags *pflag.FlagSet) error {
 	orgSlug, _ := flags.GetString("org-slug")
+	paramsYaml, _ := flags.GetString("pipeline-parameters")
+
+	var params map[string]string
+
+	if len(paramsYaml) > 0 {
+		if paramsYaml[0] == '@' {
```
kelvinkfli, line 157: What are your thoughts on preemptively reading the file and using the presence/absence of an error to determine if its valid file or not?

Here's a proposal meant to be called with `paramsYaml` as its argument.
```go
func parsePipelineValues(src string) (map[string]string, error) {
	if src == "" {
		return nil, nil
	}

	// The 'src' value can be a filepath, or a yaml string. If the file cannot be read sucessfully,
	// proceed with the assumption that the value is already valid yaml.
	raw, err := ioutil.ReadFile(src)
	if err != nil {
		raw = []byte(src)
	}

	values := map[string]string{}
	err = yaml.Unmarshal(raw, &values)
	if err != nil {
		return nil, fmt.Errorf("invalid 'pipeline-parameters' provided: %s", err.Error())
	}

	return values, nil
}
```

This might make it easier to parse, since users won't need to prefix their filepaths with `@` but we'll still be able to distinguish between the two supported strings.

kelvinkfli (commented):
Thank you for your work on adding this feature!

aaronstillwell:
@kelvinkfli @aengelberg your changes have been incorporated

kelvinkfli (approved):
Thank you for the changes @aaronstillwell!

aaronstillwell:
Closing in favor of @kelvinkfli's replica

PR #622, opened 2021-08-18
Update context-specific http client to have consistent headers

aengelberg:
Our context commands use a special REST HTTP client, not the general REST HTTP client, which we should probably fix at some point, but in the meantime I'm updating this to use consistent headers for our metrics.


kelvinkfli (approved):
LGTM! +1 on standardizing how HTTP clients are created

PR #623, opened 2021-08-20
Add new 'admin delete-namespace' command

kelvinkfli:
**Changes made:**
- Add 'DeleteNamespace' function in the api package which calls out to our [new mutation](https://github.com/circleci/api-service/pull/837)
- Add new cmd resolver that parses the given namespace, prints out a deletion plan (much like importing orbs), and executes it when the prompt is confirmed

Co-authored-by: @halie-kastl <halie-kastl@users.noreply.github.com>

**Demo: (pointing to https://extensibility.eks-dev.sphereci.com)**

delete-namespace-demo

### DIFF
--- a/cmd/admin.go
+++ b/cmd/admin.go
@@ -83,6 +83,31 @@ Example:
 		panic(err)
 	}
 
+	deleteNamespaceCommand := &cobra.Command{
+		Use:   "delete-namespace <name>",
+		Short: "Delete a namespace",
+		Long: `Delete a namespace and its related orbs.
+
+	This command deletes a namespace, as well as any orbs that have been created in the namespace.`,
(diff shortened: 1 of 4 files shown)


cmd/admin.go
```diff
@@ -83,6 +83,31 @@ Example:
 		panic(err)
 	}
 
+	deleteNamespaceCommand := &cobra.Command{
+		Use:   "delete-namespace <name>",
+		Short: "Delete a namespace",
+		Long: `Delete a namespace and it's related orbs.
```
johnswanson, line 89: So nitty, but should be `its` with no `'` 😅

johnswanson (approved):

PR #630, opened 2021-10-04
Generate a default token on runner resource class creation [RT-6]

christian-stephen:


pete-woods:
If people have scripted this command, it's going to be a breaking change.

It might make more sense to make it non-default, and have our install instructions include a `--generate-token` flag.

christian-stephen:
> If people have scripted this command, it's going to be a breaking change.
> 
> It might make more sense to make it non-default, and have our install instructions include a `--generate-token` flag.

Very valid point that also crossed my mind. I'll make the change and discuss with the team.

Issue #631, opened 2021-10-13
Can't pass non-string parameters using --pipeline-parameters


cdhunt:
I'm trying to run config process with a boolean pipeline parameter and I can't figure out how to provide it to the CLI.

    circleci config process .\.circleci\config.yml --pipeline-parameters "force-create-packages: true"
    Error: Type error for argument force-create-packages: expected type: boolean, actual value: "true" (type string)

    circleci config process .\.circleci\config.yml --pipeline-parameters '{"force-create-packages": true}'
    Error: Type error for argument force-create-packages: expected type: boolean, actual value: "true" (type string)

https://github.com/CircleCI-Public/circleci-cli/blob/master/cmd/config.go#L149-L153

PR #632, opened 2021-10-15
Update 'process config' to properly pass pipeline parameters + values

kelvinkfli:
Addresses the issue of pipeline parameters being properly coerced when a non-string type is used.

https://github.com/CircleCI-Public/circleci-cli/issues/631

Changes made:
- Preserve the original type of all parameters and values by passing a JSON-encoded string downstream

### DIFF
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -1,6 +1,7 @@
 package cmd_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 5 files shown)


cmd/config_test.go
```diff
@@ -176,7 +177,10 @@ var _ = Describe("Config", func() {
 
 				r := graphql.NewRequest(query)
 				r.Variables["config"] = config
-				r.Variables["pipelineValues"] = pipeline.PrepareForGraphQL(pipeline.LocalPipelineVars(map[string]string{}))
+
+				pipelineValues, err := json.Marshal(pipeline.LocalPipelineVars())
+				Expect(err).ToNot(HaveOccurred())
+				r.Variables["pipelineValuesJson"] = string(pipelineValues)
```
aengelberg, line 183: I can't tell how much test coverage we already have for pipeline parameters, but it would be awesome if we had tests for
* the `--pipeline-parameters` flag being recognized and propagated through
* specifically, passing a mix of string and boolean values and confirming those types are preserved in the graphql query

aengelberg (approved):
A few suggestions, but LGTM! Thanks!

PR #637, opened 2021-11-03
Ensure that a user token is passed when fetching orb info

kelvinkfli:
This fixes a bug where `api-service` could not propagate a `user-id` downstream for authorization.


halie-kastl (commented):
LGTM

halie-kastl (approved):

PR #649, opened 2021-11-24
Rewrite shellcheck job to fix SSH error

aengelberg:
See if this fixes the SSH key mismatch error.


aengelberg:
OK, should be unnecessary since we've figured out a fix to the build agent canary.

PR #666, opened 2022-02-03
added error message to FollowProject

am-bui:


halie-kastl (approved):
This is really good work. Way to go! don't forget to squash commits before merging

PR #667, opened 2022-02-13
config validate should error when using deprecated images

felicianotech:
- does the flag name make sense?
- once the blog post goes up, I need to update the error message with the link
- I need to go through to see where I can clean up and optimize
- more comments
- `machine: true`, which also deprecated, can be very tricky to flag in this context. CircleCI Server and Runner configs also use it in addition to our default/deprecated image.
- we need to grab people's attention, ideally before the brownouts, and definitely before EOL, thus this feature is on my default with a flag to turn off
- this feature processes the config before running. This pulls in images used by executor keys as well as orbs. This does mean that it won't pull in images used by a job that isn't attached to a workflow.


cmd/deprecated-images.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Processes the config down to v2.0, then checks image used against the block list
+func deprecatedImageCheck(opts configOptions, flags *pflag.FlagSet, path string) error {
+
+	var params pipeline.Parameters
+
+	orgSlug, _ := flags.GetString("org-slug")
+
+	// first process the config so we can get all images
+	response, err := api.ConfigQuery(opts.cl, path, orgSlug, params, pipeline.LocalPipelineValues())
```
aengelberg, line 51: Ah, good idea to process the config first, since Reusable Executors make this a lot more complicated to try to detect otherwise. :+1:

cmd/deprecated-images.go
```diff
@@ -0,0 +1,86 @@
...
+		image := aConfig.Jobs[key].Machine.(map[string]interface{})["image"]
+
+		// using the `docker`/`xcode` executors
+		if image == nil {
+			continue
+		}
+
+		for _, v := range deprecatedImages {
+			if image.(string) == v {
+				return errors.New("The config is using a deprecated Linux VM image. Please see <insert URL here>. This error can be ignored by using the '--ignore' flag.")
```
aengelberg, line 80: Did you mean to replace `<insert URL here>`, or are you planning to merge with the placeholder?

aengelberg: Also, I would suggest including the specific image name in this error, for clarity.

cmd/config.go
```diff
@@ -26,6 +26,7 @@ type configOptions struct {
 // Path to the config.yml file to operate on.
 // Used to for compatibility with `circleci config validate --path`
 var configPath string
+var ignore bool // should we ignore deprected images warning
```
aengelberg, line 29: minor typo
```suggestion
var ignore bool // should we ignore deprecated images warning
```

cmd/deprecated-images.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"github.com/CircleCI-Public/circleci-cli/api"
+	"github.com/CircleCI-Public/circleci-cli/pipeline"
+	"github.com/pkg/errors"
+	"github.com/spf13/pflag"
+	"gopkg.in/yaml.v3"
+)
+
+// CircleCI Linux VM images that will be permanately removed on May 31st.
```
aengelberg, line 11: small typo
```suggestion
// CircleCI Linux VM images that will be permanently removed on May 31st.
```

aengelberg (approved):
Thanks for the changes. I left a few comments, the only major request I have is about renaming the flag (`--ignore`) to something more specific. Otherwise, LGTM.

cmd/config.go
```diff
@@ -71,6 +72,7 @@ func newConfigCommand(config *settings.Config) *cobra.Command {
 	}
 	validateCommand.Annotations["<path>"] = configAnnotations["<path>"]
 	validateCommand.PersistentFlags().StringVarP(&configPath, "config", "c", ".circleci/config.yml", "path to config file")
+	validateCommand.PersistentFlags().BoolVar(&ignore, "ignore", false, "ignore deprecated images error")
```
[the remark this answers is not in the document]
KyleTryon: +1 to this, I actually had the same comment and forgot to make it.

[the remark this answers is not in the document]
felicianotech: Sounds good.

cmd/deprecated-images.go
felicianotech: Yes, that was on purpose. I was waiting for the Content Team to publish the blog post, which they did late last night. Now that it's public, I can put the proper URL.

cmd/deprecated-images.go
felicianotech: 👍🏼

cmd/deprecated-images.go
felicianotech: Added the image to the error message.

felicianotech:
Hey @aengelberg, thank you for the review. I've addressed all your comments and spelling corrections.

I still don't know how to deal with the failing tests though. Do you? My guess is that it's because I'm calling the API to process the config, and the test function can't seem to handle that. I don't know what to do about it though.

aengelberg:
@felicianotech I just took a quick look and I think the unit tests for config validation expect a very specific sequence of graphql requests to be made, and since you're now making one more request as part of the validate command, those assertions are failing.

I think that `tempSettings.AppendPostHandler` is the function we call to set those expectations, so if you update the tests to call that function one more time, with a similar request/response, that should get around this.

Another idea though: what if you refactored the `circleci config validate` command to still make only one call to `api.ConfigQuery`, then use that result for both the regular validation logic _and_ the image deprecation warnings? i.e. it would take the processed config from the response, and pass that onto the deprecated image checker after confirming there are no syntactical issues? That would kill two birds with one stone, make the command more performant, and as a bonus, the unit tests probably wouldn't need to be changed.

felicianotech:
Makes perfect sense. Great idea. I'll try that out by Monday hopefully.

felicianotech:
That worked.

jkzilla:
is there a way to just add machine: true to this tool and then have a note that says if you are Runner or Server customer it doesn't apply to you

PR #679, opened 2022-03-21
Remove picard build-agent cache

marcomorain:
Remove the cache entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ We publish the tool to [Homebrew](https://brew.sh/). The tool is [part of `homeb
 
 The particular considerations that we make are:
 
+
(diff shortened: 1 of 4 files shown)


marcomorain:
https://circleci.atlassian.net/browse/EXT-180

am-bui (approved):
Previously, `circleci local execute` caches the latest build-agent version indefinitely unless it is manually updated via `circleci update build-agent`.

Reason why it caches:
It was a behavior from before the CLI tool through a bash script. The script does the caching, so it's faster to start. The reason being, we don't have to get the specific sha every time. 

Eliminating the caching altogether:
We do not want to store the cache anymore to avoid staleness.

PR #680, opened 2022-03-25
[JIRA EXT-294] Orb Tools 11 Support

### LINKED ISSUE #646: `circleci orb init` creates repository with wrong branch
## Meta:

**CircleCI CLI Version:**
0.1.16225+8d5e389 (homebrew)

**Operating System:**
Mac OS Big Sur - 11.6 (20G165)

**CircleCI CLI Diagnostic:**
```
---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: /Users/kieranosgood/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API...
Ok.
Hello, Kieran Osgood.
```

---

## Current behavior:
When running the `circleci orb init` command at the point of selecting branch, I selected the default of main, and the output of the next steps where it says to push the branch correctly re-iterate main, but upon inspecting the branch it actually created the repository with it had used master, see:
image

## Expected behavior:
Creates repository with the correct branch

## When did this begin / Was this previously working?:
Not sure, this is first time I've used the command and I only downloaded the cli today

KyleTryon:
**Note: This PR is ready for review but please do not merge until this message has been removed**

These changes to the CLI enable users to use the `circleci orb init` command to create their orb based on the new Orb Tools 11 pipeline.
Orb Tools 11 Docs: https://github.com/circleci/circleci-docs/pull/6560
Release Notes: https://github.com/CircleCI-Public/orb-tools-orb/releases/tag/v11.0.0

# CLI Changes

- The orb template repository URL has been changed to move to a new repo.
    - This was required to ensure that we did not affect current users of the CLI as the CLI will currently download the latest tag from the original repo. It also gave us a chance to better "clean up" the existing tag structure
- The latest _semver_ tag will be downloaded from the new Orb Template
    - Tags that do not follow `vX.Y.Z` exactly are ignored in the CLI
- More automated templating
    - Orb tools 11 uses two config files with dynamic config, both are now supported by the templatizing
    - The `@orb.yml` file is now also templatized to include your GitHub URL automatically.
- Message to set branch
    - We currently require that users run a git command separately in a new window to push their code.
    - If their git client is still set to `master` this will result in error.
    - Added a second command to have the user manually set the `main` branch before pushing.
    - Fixes #646 
 
# Orb Development Changes

These changes are implemented in the new orb project template and orb-tools v11 orb which is configured by the above CLI changes.

Existing orbs can migrate to Orb Tools 11 via: https://github.com/CircleCI-Public/orb-tools-orb/blob/master/MIGRATION.md

## Notable changes:

- **Removed the 30 day limit**
  - Previously, the configuration relied on calling a `dev:alpha` tagged version of the orb for testing. Due to "dev" tags on CircleCI being ephemeral with a 30-day life span, if it had been over 30 days since the orb was last published this would result in an error in the CircleCI pipeline. Users would have to manually publish their orb locally to re-start the ci pipeline
  - The new dynamic configuration system allows us to publish the dev version of the orb _before_ calling it for testing. This means that the 30 day limit is no longer an issue.
- **Adopting Tag/Release based publishing**
  - Publishing previously required a special text flag to be added to the commit message and a new version was published on every merge to the main branch.
  - The new tag/release based publishing system will simply publish your orb when you opt to push a versioned tag or use GitHub's [releases](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) feature, which will create a tag and give you an opportunity to create a change log via [release notes](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes).
- **Review system**
  - The new "review" job can automatically detect opportunities to improve best-practices and provide native JUNIT output which will be displayed in the CircleCI UI.
  - Modeled after _shellcheck_, it is easy to skip any "review check" by supplying its "RC" code in the "exclude" parameter of the job.
- **Simplified/Improved PR Commenting**
  - Automatically comment on the PR associated with a commit when each new orb version is published (dev or production.)
  - The comment will include a link to the Orb Registry to preview dev versions of the orb, and a live link to the production version of the orb.

# Local Testing

After going through the `orb init` command with these changes:

1. The orb was created: 
image

2. The repo is created and the projects is followed
image

3. A new side effect that is known/expected/documented is that users will need to enable Dynamic Config
image

But the configs have been properly generated and modified. Once dynamic config is enabled, you are good to go. A Jira ticket has been filed as well to improve API access around this so we can enable it on-follow.

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -76,6 +76,7 @@ type createOrbTestUI struct {
 
 type orbProtectTemplateRelease struct {
 	ZipUrl string `json:"zipball_url"`
+	Name string `json:"name"`
 }
 
 func (ui createOrbTestUI) askUserToConfirm(message string) bool {
@@ -1060,7 +1061,7 @@ func initOrb(opts orbOptions) error {
 
 	fmt.Printf("Downloading Orb Project Template into %s\n", orbPath)
 	httpClient := http.Client{}
-	req, err := httpClient.Get("https://api.github.com/repos/CircleCI-Public/Orb-Project-Template/tags")
+	req, err := httpClient.Get("https://api.github.com/repos/CircleCI-Public/Orb-Template/tags")
 	if err != nil {
 		return errors.Wrap(err, "Unexpected error")
 	}
@@ -1075,15 +1076,26 @@ func initOrb(opts orbOptions) error {
 		return errors.Wrap(err, "Unexpected error")
 	}
 
-	latestTag := tags[0].ZipUrl
-	resp, err := http.Get(latestTag)
+	// filter out any non-release tags
+	releaseTags:= []orbProtectTemplateRelease{}
+	validTagRegex := regexp.MustCompile(`^v\d+\.\d+\.\d+$`)
+	for _, tag := range tags {
+		matched := validTagRegex.MatchString(tag.Name)
+		if matched {
+			releaseTags = append(releaseTags, tag)
+		}
+	}
+
+
+	latestRelease := releaseTags[0]
+	resp, err := http.Get(latestRelease.ZipUrl)
 	if err != nil {
 		return err
 	}
 	defer resp.Body.Close()
 
 	// Create the file
-	out, err := os.Create(filepath.Join(os.TempDir(), "orb-project-template.zip"))
+	out, err := os.Create(filepath.Join(os.TempDir(), "orb-template.zip"))
 	if err != nil {
 		return err
 	}
@@ -1095,7 +1107,7 @@ func initOrb(opts orbOptions) error {
 		return err
 	}
 
-	err = unzipToOrbPath(filepath.Join(os.TempDir(), "orb-project-template.zip"), orbPath)
+	err = unzipToOrbPath(filepath.Join(os.TempDir(), "orb-template.zip"), orbPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cmd/orb.go
```diff
@@ -1297,13 +1309,24 @@ func initOrb(opts orbOptions) error {
...
 
-	circleConfig, err := ioutil.ReadFile(path.Join(orbPath, ".circleci", "config.yml"))
+	circleConfig1, err := ioutil.ReadFile(path.Join(orbPath, ".circleci", "config.yml"))
 	if err != nil {
 		return err
 	}
 
-	circle := string(circleConfig)
-	err = ioutil.WriteFile(path.Join(orbPath, ".circleci", "config.yml"), []byte(orbTemplate(circle, projectName, ownerName, orbName, namespace)), 0644)
+	config1String := string(circleConfig1)
```
michael-webster, line 1317: Is `config1string` and `config2string` still needed?

michael-webster (commented):
Do we need to provide any documentation or updates to consumers of the CLI or is this a transparent change?

cmd/orb.go
KyleTryon: If I understand correctly, these are both my new changes to this orb. We now use two config files rather than one as we did previously, as orb tools 11 uses dynamic config.

michael-webster (approved):

cmd/orb.go
```diff
@@ -1075,15 +1076,26 @@ func initOrb(opts orbOptions) error {
 		return errors.Wrap(err, "Unexpected error")
 	}
 
-	latestTag := tags[0].ZipUrl
-	resp, err := http.Get(latestTag)
+	// filter out any non-release tags
+	releaseTags:= []orbProtectTemplateRelease{}
```
nramabad, line 1080:
```suggestion
	releaseTags := []orbProtectTemplateRelease{}
```

nramabad (approved):
LGTM! left one small nit of a comment. Can you add [JIRA-###] in place of feat: in the PR title?

Approving but it looks like codecov is slightly down. not sure if that's a ongoing issue or one you can address.

KyleTryon:
Suggestion applied and PR title updated. Thank you @nramabad.

For the code coverage, no new functions were added but lines were added to an existing function (which i believe itself was actually already not tested). For the size/scope of the change I would _believe_ this is ok as increasing coverage now would require breaking up a function for the purpose of testing. Looking for input on this though!

halie-kastl (approved):

PR #682, opened 2022-03-31
[EXT-242] Upgrade Golang version in Dockerfile

KyleTryon:
The original image was based on a deprecated version of Ubuntu. Testing on 1.17

The goal is to update the base image in the public docker image, so that the depreciation notice in CircleCI is resolved.

Docker image: https://hub.docker.com/r/circleci/circleci-cli
Related Issue: https://github.com/CircleCI-Public/orb-tools-orb/issues/130#issuecomment-1082505567
Depreciation Warning: https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ commands:
         default: https://github.com/goreleaser/goreleaser/releases/download/v0.127.0/goreleaser_amd64.deb
     steps:
       - restore_cache:
-          keys: [v4-goreleaser-]
+          keys: [v5-goreleaser-]
       - run:
           name: Install GoReleaser
           command: |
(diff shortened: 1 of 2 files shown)


felicianotech (approved):
This should also happen in the CircleCI config at some point.

KyleTryon:
I am noticing I don't believe I see where the docker image is built/tested/deployed in this workflow?

wyardley (approved):

li3n3:
Updated image version further, since we noticed it's been updated more (1.18 versus 1.17) since this PR was originally submitted to us.

However, changes to the `master` branch here seemed to introduce a problem: the "Restoring cache" step skyrocketed from ~3 minutes to ~41 (including in PR #741 yesterday, where it didn't technically fail or time out, but did take FAR longer than it used to). This began to cause this PR's commits to fail CI.

It seemed like our cache was woefully stale, so I've updated the cache keys here, and the latest run is consistent with its historical timing, and also (crucially) passes again.

wyardley:
@li3n3 any chance you can look at #713 too?

PR #685, opened 2022-04-08
Remove errantly committed file.

michael-webster:


adamdmharvey:
Thinking this could be closed?  The commits just remove some whitespace, so whatever errantly committed file was there, it may have been cleaned up separately?

rlegan:
as @adamdmharvey  pointed out, this pr is not needed since it now only removes some whitespaces. closing it!

PR #688, opened 2022-04-16
Applied gofmt to unformatted files

mkusaka:
Applied gofmt to unformatted files.
Also, added an automatic check on ci.

### DIFF
--- a/.circleci/install-lint.sh
+++ b/.circleci/install-lint.sh
@@ -10,6 +10,6 @@ function error() {
 
 trap error SIGINT
 
-curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.35.2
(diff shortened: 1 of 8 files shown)


.circleci/install-lint.sh
```diff
@@ -10,6 +10,6 @@ function error() {
 
 trap error SIGINT
 
-curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.35.2
+curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.45.2
```
wyardley, line 13: another option (not necessarily in this PR) would be to use the `golangci-lint` docker image directly, which I imagine would work here...

wyardley:
Nice.

LMK what you think of #713

mkusaka:
@wyardley 
Yeah. I think that's the best solution. :)

I'm closing this PR because these changes are included to #712.

PR #689, opened 2022-04-21
orb source set token missing

corinnesollows:
**Issue**
Command "orb source" for private orbs was not functioning. 

**Changes made**
SetToken before running the query

**Reasoning**
Customers could not use source command for private orbs, but they could for public ones
This fix increases feature parity between public and private orbs. 

**Testing**
Running the cli command orb source and seeing that it now works for private orbs as well as public


ruben1 (approved):

am-bui (approved):
Glad you followed your hunch and compared the diff between `orb info` and `orb source`! I see that we just needed to add this crucial line of setting the token. Nice job!

LGTM!

PR #690, opened 2022-04-22
[EXT-278] Support org id in namespace create command

am-bui:
Co-authored-by: @ruben1 <ruben1@users.noreply.github.com>
Co-authored-by: @corinnesollows <corinnesollows@users.noreply.github.com>
Co-authored-by: @rykimcircle <rykimcircle@users.noreply.github.com>

- Supported `org-id` in namespace create command as a flag
- Included `vcs-type` and `org-name` as optional arguments
- Created examples for users

namespace create help
namespace create orgid


halie-kastl (approved):
Lgtm! Nice work

api/api.go
```diff
@@ -802,7 +802,7 @@ func CreateImportedNamespace(cl *graphql.Client, name string) (*ImportNamespaceR
 	return &response, nil
 }
 
-func createNamespaceWithOwnerID(cl *graphql.Client, name string, ownerID string) (*CreateNamespaceResponse, error) {
+func CreateNamespaceWithOwnerID(cl *graphql.Client, name string, ownerID string) (*CreateNamespaceResponse, error) {
```
rykimcircle, line 805: Just curious on why this was changed from private to public? Not a big deal either way just wondering.

rykimcircle (approved):

cmd/namespace.go
```diff
@@ -118,17 +147,32 @@ To change the namespace, you will have to contact CircleCI customer support.
...
+		_, err = uuid.Parse(*opts.orgID)
+		if err == nil {
+			return createNamespaceWithOrgId(opts, namespaceName, *opts.orgID)
+		}
+
+	//skip if no vcs type and org name provided
+	} else if len(opts.args) > 1{
+		return createNamespaceWithVcsTypeAndOrgName(opts, namespaceName, opts.args[1], opts.args[2])
+	}
+	return fmt.Errorf("please provide org-id or vcs type and org name")
```
ruben1, line 175: I'd say it's more useful here to return `cmd.Help()` to print the whole help command

cmd/namespace.go
```diff
@@ -118,17 +147,32 @@ To change the namespace, you will have to contact CircleCI customer support.
...
+	namespaceName := opts.args[0]
+	//skip if no orgid provided
+	if opts.orgID != nil && strings.TrimSpace(*opts.orgID) != ""{
+		_, err = uuid.Parse(*opts.orgID)
+		if err == nil {
+			return createNamespaceWithOrgId(opts, namespaceName, *opts.orgID)
+		}
+
+	//skip if no vcs type and org name provided
+	} else if len(opts.args) > 1{
```
ruben1, line 172: should check for len == 3, otherwise `opts.args[2]` will throw if there's only 2 args

cmd/namespace.go
```diff
@@ -118,17 +147,32 @@ To change the namespace, you will have to contact CircleCI customer support.
...
 		fmt.Printf("Namespace `%s` created.\n", namespaceName)
 		fmt.Println("Please note that any orbs you publish in this namespace are open orbs and are world-readable.")
 	}
+	return err
+}
 
-	return nil
+func createNamespace(opts namespaceOptions) error {
+	var err error
+	namespaceName := opts.args[0]
```
ruben1, line 163: same, this will throw if no args passed. in line 79, we can use a range to force at least 1 arg `Args: cobra.RangeArgs(1, 3),` instead of `Args: cobra.MaximumNArgs(3),`

cmd/namespace.go
```diff
@@ -103,9 +108,33 @@ func deleteNamespaceAlias(opts namespaceOptions) error {
...
+		}
+
+		fmt.Printf("Namespace `%s` created.\n", namespaceName)
+		fmt.Println("Please note that any orbs you publish in this namespace are open orbs and are world-readable.")
+	}
+	return nil
+}
 
+func createNamespaceWithVcsTypeAndOrgName(opts namespaceOptions, namespaceName, vcsType, orgName string) error {
+	var err error
```
ruben1, line 137: is this variable needed?

corinnesollows: Yes, if it is not declared here it will be declared shortly after within the same function. :)

am-bui: I changed it to an implicit assignment!

cmd/namespace.go
am-bui: Good idea!

PR #693, opened 2022-05-04
Remove uses of packr

conormcd:
We don't need to embed a YAML file and then parse it just for the sake of three static URLs that could live in code. This change removes the need for architecture-specific stuff and should solve #683 and #692.


data/data.go
```diff
@@ -1,39 +1,21 @@
...
-
-	err = yaml.Unmarshal(bts, &d)
-	if err != nil {
-		return nil, err
-	}
-
-	return d, nil
+var Data = DataBag{
+	Links: struct {
+		CLIDocs     string `yaml:"cli_docs"`
```
ruben1, line 13: not a blocker but these `yaml:` references look like deserialization mappings so might not be needed any more

ruben1:
Resolving a merge conflict ^

ruben1 (approved):
looks good, thanks for the fix!

PR #697, opened 2022-05-05
Ext 281 context update

rykimcircle:


cmd/context_test.go
```diff
@@ -119,17 +119,16 @@ https://circleci.com/account/api`))
 			It("prints all in-band errors returned by the API", func() {
 				By("setting up a mock server")
 				tempSettings.AppendRESTPostHandler(clitest.MockRequestResponse{
-					Status:        http.StatusOK,
-					Request:       fmt.Sprintf(`{"name": "%s","owner":{"id":"\"%s\""}}`, contextName, orgID),
+					Status:        http.StatusInternalServerError,
```
rykimcircle, line 122: Forces our client to actually error

cmd/context_test.go
```diff
@@ -119,17 +119,16 @@ https://circleci.com/account/api`))
...
 					Response:      fmt.Sprintf(`{"id": "497f6eca-6276-4993-bfeb-53cbbbba6f08", "name": "%s", "created_at": "2015-09-21T17:29:21.042Z" }`, contextName),
 					ErrorResponse: `{ "message": "ignored error" }`,
 				})
 				By("running the command")
 				session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
 
 				Expect(err).ShouldNot(HaveOccurred())
-				Eventually(session.Err).Should(gbytes.Say(`Error: error1
-error2`))
+				Eventually(session.Err).Should(gbytes.Say(`Error: ignored error`))
```
rykimcircle, line 131: Which is then returned and printed to terminal

corinnesollows:
We're missing a few function declaration comments. 😵‍💫

api/context.go
```diff
@@ -25,8 +25,8 @@ type ContextInterface interface {
 	Contexts(vcs, org string) (*[]Context, error)
 	ContextByName(vcs, org, name string) (*Context, error)
 	DeleteContext(contextID string) error
-	CreateContext(vcs, org, name string) (error)
-
+	CreateContext(vcs, org, name string) error
+	CreateContextWithOrgID(orgID *string, name string) error
```
joeyorlando, line 29: should `orgID` be `string` or `*string`?

I _think_ the main distinction between the two is that `string` is stricter in that it cannot be `nil` whereas `*string` can. With `*string` you'd have to make sure the pointer is `nil` anywhere you dereference it (I found this [quick reference](https://dhdersch.github.io/golang/2016/01/23/golang-when-to-use-string-pointers.html) insightful).

corinnesollows:
> We're missing a few function declaration comments. 😵‍💫

Added a bunch, We are good now with comments

rykimcircle:
> > We're missing a few function declaration comments. 😵‍💫
> 
> Added a bunch, We are good now with comments

awesome ty

api/context_rest_test.go
```diff
@@ -0,0 +1,77 @@
...
+			Status:   http.StatusOK,
+			Request:  fmt.Sprintf(`{"name": "%s","owner":{"id":"%s"}}`, name, orgID),
+			Response: fmt.Sprintf(`{"id": "%s", "name": "%s", "created_at": "2015-09-21T17:29:21.042Z" }`, orgID, name),
+		})
+
+		err = client.CreateContextWithOrgID(&orgID, name)
+		Expect(err).To(BeNil())
+	})
+
+})
```
joeyorlando, line 77: nitpicking here but maybe an error-checking test similar to the gql flavour [here](https://github.com/CircleCI-Public/circleci-cli/pull/697/files#diff-6e1c5b6b63a92c6db7c81885caecc79815d1a1eabb913c9b8f415751e2d259ebR99)

cmd/context.go
```diff
@@ -174,9 +196,22 @@ func readSecretValue() (string, error) {
...
 
-func createContext(client api.ContextInterface, vcsType, orgName, contextName string) error {
-	err := client.CreateContext(vcsType, orgName, contextName)
-	return err
+//createContext determines if the context is being created via orgid or vcs and org name
+//and navigates to corresponding function accordingly
+func createContext(cmd *cobra.Command, client api.ContextInterface, args []string) error {
+	//skip if no orgid provided
+	if orgID != nil && strings.TrimSpace(*orgID) != "" && len(args) == 1 {
+		_, err := uuid.Parse(*orgID)
```
joeyorlando, line 204: what happens if the user inputs an invalid UUID?

cmd/context.go
```diff
@@ -103,6 +119,12 @@ func newContextCommand(config *settings.Config) *cobra.Command {
 
 	deleteContextCommand.Flags().BoolVarP(&force, "force", "f", false, "Delete the context without asking for confirmation.")
 
+	orgID = createContextCommand.Flags().String("org-id", "", "The id of your organization.")
+	createContextCommand.Flags().BoolVar(&integrationTesting, "integration-testing", false, "Enable test mode to setup rest API")
```
joeyorlando, line 123: is this flag visible to the end-user?

cmd/context.go
```diff
@@ -103,6 +119,12 @@ func newContextCommand(config *settings.Config) *cobra.Command {
 
 	deleteContextCommand.Flags().BoolVarP(&force, "force", "f", false, "Delete the context without asking for confirmation.")
 
+	orgID = createContextCommand.Flags().String("org-id", "", "The id of your organization.")
```
joeyorlando, line 122: checked out the branch and tried running:
```bash
$ make build
$ ./build/darwin/amd64/circleci context create --help
Create a new context


Usage:
  circleci context create <vcs-type> <org-name> <context-name> [flags]

Flags:
  -h, --help   help for create

Global Flags:
      --host string         URL to your CircleCI host, also CIRCLECI_CLI_HOST (default "https://circleci.com/")
      --skip-update-check   Skip the check for updates check run before every command.
      --token string        your token for using CircleCI, also CIRCLECI_CLI_TOKEN
```

I think it'd be useful if the `--org-id` showed up under "Flags" 🙂

joeyorlando:
aside from the two minor comments LGTM 🚢

cmd/context.go
rykimcircle: It'll just fall down to the return cmd.Help() command which shows the help menu. We tried to use the same logic as the comment [here](https://github.com/CircleCI-Public/circleci-cli/pull/690#discussion_r863687349) but I think it makes more sense just to print the error instead. 

Maybe something like `Org ID error: invalid uuid` or something?

joeyorlando (approved):

PR #702, opened 2022-05-11
Add API bindings for the pipeline scheduling REST API

sulami:
This is actually meant to be used in the CircleCI terraform provider,
which uses the CLI as a library. As such I'm not adding any
CLI-user-visible way of interacting with schedules as part of this.

### DIFF
--- a/api/schedule.go
+++ b/api/schedule.go
@@ -1,0 +1,42 @@
+package api
+
+import (
+	"time"
+)
+
+type Timetable struct {
+	PerHour    uint     `json:"per-hour"`
+	HoursOfDay []uint   `json:"hours-of-day"`
+	DaysOfWeek []string `json:"days-of-week"`
+}
+
+type Actor struct {
+	ID    string `json:"id"`
+	Login string `json:"login"`
+	Name  string `json:"name"`
+}
+
+type Schedule struct {
+	ID          string            `json:"id"`
+	ProjectSlug string            `json:"project-slug"`
+	Name        string            `json:"name"`
+	Description string            `json:"description,omitempty"`
+	Timetable   Timetable         `json:"timetable"`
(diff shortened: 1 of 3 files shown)


api/schedule.go
```diff
@@ -0,0 +1,42 @@
...
+	Description string            `json:"description,omitempty"`
+	Timetable   Timetable         `json:"timetable"`
+	Actor       Actor             `json:"actor"`
+	Parameters  map[string]string `json:"parameters"`
+	CreatedAt   time.Time         `json:"created_at"`
+	UpdatedAt   time.Time         `json:"updated_at"`
+}
+
+type ScheduleInterface interface {
+	Schedules(vcs, org, project string) (*[]Schedule, error)
```
elliotforbes, line 32: ❓  should we add `ctx context.Context` ad the first argument to all of these functions? 

Will whatever is calling this ever care about us honouring timeouts or cancellations of requests that could be in-flight?

sulami: In this case it seems that neither the CLI nor the terraform provider seem to, at least in existing code.

elliotforbes (approved):

api/schedule.go
elliotforbes: Very fair, it's not worth adding in this case then

corinnesollows (changes requested):
I would love to see more comments added before each func declaration. This is one of the top rated '[best practises](https://www.bacancytechnology.com/blog/go-best-practices)' in Golang. (I linked a reference for you :) )
"Use comments
Using comments might seem very basic practice but it’s very helpful when there are so many developers working on the same project and for transferring knowledge as well. The comments should have a brief description stating the purpose of the business logic. In this way, any other developer going through your logic will have an idea of what a particular block does."

corinnesollows (approved):

PR #706, opened 2022-05-12
Convert context table to use Charm table instead

rykimcircle:
convert context table to use charm instead.


corinnesollows (changes requested):
I would love to see more comments added before each func declaration. This is one of the top rated '[best practises](https://www.bacancytechnology.com/blog/go-best-practices)' in Golang. (I linked a reference for you :) )
"Use comments
Using comments might seem very basic practice but it’s very helpful when there are so many developers working on the same project and for transferring knowledge as well. The comments should have a brief description stating the purpose of the business logic. In this way, any other developer going through your logic will have an idea of what a particular block does."

corinnesollows:
Code has been extracted and made into smaller PRs

PR #715, opened 2022-05-30
[SECENG-573] Added list-policies command.

sagar-connect:
# Checklist

=========

## Changes

=======

- Added ListPolicies command in the CLI
- Added tests
- added .idea to to gitignore
- fixed root_test.go

## Rationale

=========

Jira: https://circleci.atlassian.net/browse/SECENG-573

We want to add a new group of commands for allowing build config policy management via the CLI.
This PR adds the first command called-> list policies which gets and prints all policies corresponding to the given ownerID.

## Considerations

==============

NA


api/policy/policy.go
```diff
@@ -0,0 +1,124 @@
...
+
+// Client communicates with the CircleCI policy-service to ask questions
+// about policies. It satisfies policy.ClientInterface.
+type Client struct {
+	token     string
+	serverUrl string
+	client    *http.Client
+}
+
+type errorResponse struct {
```
davidmdm, line 28: wrong error struct... Policy-Service returns errors in this format:

```go
type HTTPError struct {
        Error   string                 `json:"error"`
        Context map[string]interface{} `json:"context,omitempty"`
}
```

api/policy/policy.go
```diff
@@ -0,0 +1,124 @@
...
+	if activeFilter != "" {
+		urlParams.Add("active", activeFilter)
+	}
+
+	queryURL.RawQuery = urlParams.Encode()
+
+	return c.newHTTPRequest("GET", queryURL.String(), nil)
+}
+
+func (c *Client) newHTTPRequest(method, url string, body io.Reader) (*http.Request, error) {
```
davidmdm, line 85: Should we use context?

cmd/policy/policy.go
```diff
@@ -0,0 +1,69 @@
...
+		Short: "Policies ensures security of build configs via security policy management framework. " +
+			"This group of commands allows the management of polices to be verified against build configs.",
+	}
+
+	listPoliciesCommand := &cobra.Command{
+		Short:   "List all policies",
+		Use:     "list <ownerID> [<active>]",
+		PreRunE: initClient,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listPolicies(policyClient, args)
```
davidmdm, line 36: simply create policyClient here instead of having singleton style initialization.

cmd/policy/policy.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+		Args:        cobra.RangeArgs(1, 2),
```
davidmdm, line 38: using flags for owner and active filter would be much nicer than positional args.

api/policy/policy.go
sagar-connect: Right!
How do you feel about not parsing the error response at all, and just return the raw response body in case of error?

davidmdm: seems ok. You could parse out the error string and simply print the string rather than the entire json. Might be nicer visually, but that being said, I am 100% ok with raw json as well.

PR #717, opened 2022-05-31
Ext 283 remove legacy org slug

corinnesollows:
# Checklist

=========

## Changes

=======

- Added the ability to use orgId when running local execute
- If orgId is not provided an orgSlug can still be used

## Considerations

==============

In case of users currently using the legacy slug org, it remains for now.


cmd/build.go
```diff
@@ -17,6 +17,7 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 
 	local.AddFlagsForDocumentation(buildCommand.Flags())
 	buildCommand.Flags().StringP("org-slug", "o", "", "organization slug (for example: github/example-org), used when a config depends on private orbs belonging to that org")
+	buildCommand.Flags().String("org-id", "", "organization id, used when a config depends on private orbs belonging to that org")
```
joeyorlando, line 20: nitpick here but I'd argue keeping the wording the same (or extract to a variable even) for the three `<command>.Flags().String("org-id"...`

corinnesollows:
Hey @joeyorlando , this isnt ready for review. lol
Great initiative though!

cmd/build.go
corinnesollows: Could you give a full example? The wording is the same as far as I can see...

local/local.go
```diff
@@ -22,13 +23,24 @@ var picardRepo = "circleci/picard"
...
+	var err error
+	var configResponse *api.ConfigResponse
 	cl := graphql.NewClient(cfg.HTTPClient, cfg.Host, cfg.Endpoint, cfg.Token, cfg.Debug)
-	configResponse, err := api.ConfigQuery(cl, configPath, orgSlug, nil, pipeline.LocalPipelineValues())
 
-	if err != nil {
-		return err
+	processedArgs, configPath := buildAgentArguments(flags)
+
+	orgID, _ := flags.GetString("org-id")
```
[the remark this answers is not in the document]
corinnesollows: Thats a great question. I was reusing functionality that already existed, so I looked this up and for some reason it seems to be [standard not to get this error.](https://ordina-jworks.github.io/development/2018/10/20/make-your-own-cli-with-golang-and-cobra.html#adding-flags). I linked one of several sites I found this in.

api/api.go
```diff
@@ -556,17 +555,76 @@ func ConfigQuery(cl *graphql.Client, configPath string, orgSlug string, params p
...
+	var query string
+	config, err := loadYaml(configPath)
+	if err != nil {
+		return nil, err
+	}
+	// GraphQL isn't forwards-compatible, so we are unusually selective here about
+	// passing only non-empty fields on to the API, to minimize user impact if the
+	// backend is out of date.
+	var fieldAddendums string
+	if orgId != "" {
```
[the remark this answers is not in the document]
corinnesollows: Exactly! If its empty it should be using the orgSlug instead and if the orgSlug is not provided it also works with no flags

api/api.go
```diff
@@ -556,17 +555,76 @@ func ConfigQuery(cl *graphql.Client, configPath string, orgSlug string, params p
...
 	}
+	if len(response.BuildConfig.ConfigResponse.Errors) > 0 {
+		return nil, &response.BuildConfig.ConfigResponse.Errors
+	}
+
+	return &response.BuildConfig.ConfigResponse, nil
+}
 
+// ConfigQuery calls the GQL API to validate and process config with the org id
+func ConfigQuery(cl *graphql.Client, configPath string, orgId string, params pipeline.Parameters, values pipeline.Values) (*ConfigResponse, error) {
```
rykimcircle, line 577: Nitpick so optional, but I think we're capitalizing ID in the rest of the project, so orgID.

api/api.go
```diff
@@ -513,16 +513,14 @@ func WhoamiQuery(cl *graphql.Client) (*WhoamiResponse, error) {
 	return &response, nil
 }
 
-// ConfigQuery calls the GQL API to validate and process config
-func ConfigQuery(cl *graphql.Client, configPath string, orgSlug string, params pipeline.Parameters, values pipeline.Values) (*ConfigResponse, error) {
+// ConfigQueryLegacy calls the GQL API to validate and process config with the legacy orgSlug
+func ConfigQueryLegacy(cl *graphql.Client, configPath string, orgSlug string, params pipeline.Parameters, values pipeline.Values) (*ConfigResponse, error) {
```
rykimcircle, line 517: I think we can simplify these two functions  (ConfigQuery, and ConfigQueryLegacy) into one. Since this is graph, couldn't we just send everything that we've collected, like both the org slug and orgID? The API should already be handling this logic so there's no need to do it here too.

corinnesollows: We tried that at first, remember? Graph did not like it when one of those fields wasn't sent. Also.... this is the API....

joeyorlando (approved):

api/api.go
rykimcircle: Really? I just tried refactoring it locally and it worked just fine? Not a big deal though but I think less code is better than more code. Let me know if you want to spend the time to refactor it and I can help ya out :D

rykimcircle (approved):

PR #724, opened 2022-06-02
added update policy command

ssyms:
# Checklist

=========

- [ x] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [ x] I have checked for similar issues and haven't found anything relevant.
- [ x] This is not a security issue (which should be reported here: https://circleci.com/security/)
- [x ] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/master/CONTRIBUTING.md).

## Changes

=======

- Added `policy update` command in `cmd/policy/policy.go` (and added appropriate tests)
- Added code to update policy in policy-service in `api/policy/policy.go` (and added appropriate tests)
- Added usage text for testing in `cmd/policy/testdata`

## Rationale

=========

Add the ability to update a policy from the Circle CLI tool

## Considerations

==============

* The `cobra` library used to parse commands does not allow boolean values to be set to nil. To reduce the risk of a user accidentally disabling or enabling a policy, this flag was set to required with a `true` default value. The user will always need to include the `--active` flag (or could set it to `--active=false` if needed).

## Screenshots

============

<h4>Before</h4>
Image or [gif](https://giphy.com/apps/giphycapture)

<h4>After</h4>
Image or gif where change can be clearly seen

### DIFF
--- a/api/policy/policy.go
+++ b/api/policy/policy.go
@@ -110,7 +110,6 @@ func (c Client) CreatePolicy(ownerID string, policy CreationRequest) (interface{
 	return &response, nil
 }
 
-
 type UpdateRequest struct {
 	Name    *string `json:"name,omitempty"`
 	Context *string `json:"context,omitempty"`
@@ -118,6 +117,7 @@ type UpdateRequest struct {
 	Active  *bool   `json:"active,omitempty"`
 }
 
+// UpdatePolicy calls the UPDATE policy API in the policy-service. It updates a policy in the policy-service matching the given owner-id and policy-id.
 func (c Client) UpdatePolicy(ownerID string, policyID string, policy UpdateRequest) (interface{}, error) {
 	data, err := json.Marshal(policy)
 	if err != nil {
--- a/api/policy/policy.go
+++ b/api/policy/policy.go
@@ -103,6 +103,52 @@ func (c Client) CreatePolicy(ownerID string, policy CreationRequest) (interface{
 	return &response, nil
 }
 
+type UpdateRequest struct {
+	Name    *string `json:"name,omitempty"`
+	Context *string `json:"context,omitempty"`
+	Content *string `json:"content,omitempty"`
+	Active  *bool   `json:"active,omitempty"`
+}
+
+func (c Client) UpdatePolicy(ownerID string, policyID string, policy UpdateRequest) (interface{}, error) {
+	data, err := json.Marshal(policy)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode policy payload: %w", err)
+	}
+
+	req, err := http.NewRequest(
+		"PATCH",
+		fmt.Sprintf("%s/api/v1/owner/%s/policy/%s", c.serverUrl, ownerID, policyID),
+		bytes.NewReader(data),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct request: %v", err)
+	}
+
+	req.Header.Set("Content-Length", strconv.Itoa(len(data)))
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get response from policy-service: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		var response httpError
+		if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+			return nil, fmt.Errorf("unexpected status-code: %d", resp.StatusCode)
+		}
+		return nil, fmt.Errorf("unexpected status-code: %d - %s", resp.StatusCode, response.Error)
+	}
+
+	var response interface{}
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return nil, fmt.Errorf("failed to parse response: %w", err)
+	}
+
+	return &response, nil
+}
+
 func (c Client) GetPolicy(ownerID string, policyID string) (interface{}, error) {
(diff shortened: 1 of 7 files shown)


sagar-connect:
We chose to accept policy-id as positional argument, instead of a flag.
https://github.com/CircleCI-Public/circleci-cli/pull/723/files

api/policy/policy.go
```diff
@@ -103,6 +103,52 @@ func (c Client) CreatePolicy(ownerID string, policy CreationRequest) (interface{
 	return &response, nil
 }
 
+type UpdateRequest struct {
+	Name    string `json:"name"`
```
sagar-connect, line 107: omitempty for all fields?

cmd/policy/policy.go
```diff
@@ -134,9 +134,62 @@ func NewCommand(config *settings.Config, preRunE validator) *cobra.Command {
...
+
+		cmd.Flags().StringVar(&policyID, "policy-id", "", "id of the policy to update")
+		cmd.Flags().StringVar(&updateRequest.Name, "name", "", "name of policy to update")
+		cmd.Flags().StringVar(&updateRequest.Context, "context", "config", "policy context")
+		// Default value of bool flag must be true or false (cannot be nil).
+		// To avoid accidently enabling a policy, will make this flag required.
+		cmd.Flags().BoolVarP(&updateRequest.Active, "active", "a", true, "is policy active (to ensure a policy is inactive, use --active=false)")
+		cmd.Flags().StringVar(&policyPath, "policy", "", "path to rego policy file")
+
+		cmd.MarkFlagRequired("policy-id")
```
sagar-connect, line 181: I'm not sure if it's best to make all flags mandatory. Say a user once created a policy long back (they don't have content for policy with them anymore locally), and they want to set the policy to active, they would just want to update the active status.

cmd/policy/policy.go
```diff
@@ -134,9 +134,62 @@ func NewCommand(config *settings.Config, preRunE validator) *cobra.Command {
...
+				enc.SetIndent("", "  ")
+
+				if err := enc.Encode(result); err != nil {
+					return fmt.Errorf("failed to encode result to stdout: %w", err)
+				}
+
+				return nil
+			},
+			Args:    cobra.ExactArgs(0),
+			Example: `policy update --owner-id 462d67f8-b232-4da4-a7de-0c86dd667d3f --policy-id e9e300d1-5bab-4704-b610-addbd6e03b0b --name policy_name --active --policy ./policy.rego`,
```
davidmdm, line 170: policy-id as the resource ID would be a good candidate for positional:

Example:
```
circle policy update some-resource-uuid-string --name new-name
```

cmd/policy/policy.go
davidmdm: Technically none are mandatory! You could be smart and in the API code check if the UpdateRequest is empty and do a noop

davidmdm (changes requested):
Just a couple comments. Especially around the `active` flag and required flags.
Otherwise it is Supercalifragilisticexpialidocious, as they say.

cmd/policy/policy.go
```diff
@@ -134,9 +134,62 @@ func NewCommand(config *settings.Config, preRunE validator) *cobra.Command {
...
+				policyData, err := os.ReadFile(policyPath)
+				if err != nil {
+					return fmt.Errorf("failed to read policy file: %w", err)
+				}
+
+				updateRequest.Content = string(policyData)
+
+				client := policy.NewClient(*policyBaseURL, config)
+
+				result, err := client.UpdatePolicy(*ownerID, policyID, updateRequest)
```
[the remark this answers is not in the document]
davidmdm: imagine the scenario where a user wants to update the name or the policy of an inactive policy but the default is active true and they accidentally activate a policy for their org... :(

cmd/policy/policy.go
ssyms: since we also require `owner-id` as a non-positional flag I think it makes sense to make `policy-id` non-positional too. Also reduces confusion since we have two uuids - easy to mix up

ryan-wren (approved):

cmd/policy/policy.go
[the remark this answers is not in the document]
davidmdm: 100% ! Just wanted to make sure we could communicate an absent active value to the API and this does that!

cmd/policy/policy.go
davidmdm: It would only be confusing if the both were unflagged. I think we should rely on prior art here in  CLI design. Examples include kubectl, when working on the resource the id is positional:

```
kubectl describe pod {PodID} --flags...
docker container rm {containerID} --flags...

circle-cli policy update {policyID} --flags ...
```

ssyms: those are both examples of commands with only a single UUID, so there's never any confusion. If anything, the org id should have been positional since it's required for every policy command

cmd/policy/policy.go
```diff
@@ -134,9 +134,88 @@ func NewCommand(config *settings.Config, preRunE validator) *cobra.Command {
...
+				}
+
+				client := policy.NewClient(*policyBaseURL, config)
+
+				if cmd.Flag("active").Changed {
+					updateRequest.Active = &active
+				}
+
+				if cmd.Flag("context").Changed {
+					if context != "config" {
```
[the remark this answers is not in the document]
ssyms: I don't think it will impact us that much to have to maintain an allow-list but I don't have any issues with removing it, so done

davidmdm (approved):

sagar-connect (approved):

PR #731, opened 2022-06-03
Added `policy` command

sagar-connect:
# Checklist

=========

## Changes

=======

- added new policy command at `cmd/policy/policy.go`
- added policy subcommands  `get, list, create, delete`
- added policy api at `api/policy/policy.go`
- added tests
- added policy command to root circle command
- added godocs to functions
- added .idea to .gitignore

## Rationale

The Seceng team is releasing a new Policy-Service and the first way for customers to interact wit the policy apis is via the circleci-cli. 

The CLI must be able to create, update, delete, view and so on, policies as needed.

Policy-service endpoints documentation : https://github.com/circleci/policy-service/blob/main/cmd/server/openapi.yaml

Relevant motivations:
Epic: https://circleci.atlassian.net/browse/SECENG-270
Confluence: https://circleci.atlassian.net/wiki/spaces/SD/pages/6437569182/Policy+Enforcement+on+Configs+Feature+Brief

## Considerations

- Attempted to isolate seceng relevant code in `policy` subfolders: `cmd/policy` & `api/policy`
- Used a policy flag `policy-base-url` in the policy root command to specify policy-api instead of using `settings.Config.Host`


cmd/policy/policy.go
```diff
@@ -0,0 +1,242 @@
...
+				enc.SetIndent("", "  ")
+
+				if err := enc.Encode(result); err != nil {
+					return fmt.Errorf("failed to encode result to stdout: %w", err)
+				}
+
+				return nil
+			},
+			Args:    cobra.ExactArgs(1),
+			Example: `policy update e9e300d1-5bab-4704-b610-addbd6e03b0b --owner-id 462d67f8-b232-4da4-a7de-0c86dd667d3f --name policy_name --active --policy ./policy.rego`,
```
rykimcircle, line 224: Not a huge deal but these policy commands go against most of our input patterns found in other commands where it's usually `<command> <action> <name>` over name being set as a flag (examples: context, namespace). 

Might be a bit weird for our users for these commands to be different. As a user, personally I would find it a bit annoying to remember all the different syntaxes for all the different commands.

rykimcircle (approved):
Looks good to me on this side but sort of working half blind without any documentation regarding these endpoints. Could you link where these endpoints were implemented? Thanks!

cmd/policy/policy.go
sagar-connect: In our case, name is a required input for `create` subcommand, we can change it to <command> <action> <name> format.
However for `update` command, id is the primary input, and name is optional param with which user can update the name as well, so we won't be able use <command> <action> <name> format.

So I think it's better we keep using the flag for name for consistent experience at least for all `policy` commands.

PR #737, opened 2022-06-11
Add support for multiple <<includes(file)>> and embedded <<includes(file)>>

Mongey:
# Checklist

=========

## Changes

=======

- Add support for multiple `<<includes(file)>>` and embedded `<<includes(file) >>`

## Rationale

=========

I see there are cucumber tests to assert this behaviour is not supported, but I would like it to work this way 😄 
Usecase: I have a template file in my orb, and rather than embedded the contents directly in my command 
```yaml
 - run: echo 'my-extremely-large-template\nis here' > foo.tmpl
```
I want to have
```yaml
- run: echo '<< include(template.tmpl) >>' > foo.tmpl
```
where `template.tmpl` is
```
my-extremely-large-template
is here
```

## Considerations

==============
n/a


pete-woods:
Superseded by #1684 (merged), which implements this proposal — multiple and embedded `<< include(...) >>` directives — on the v1 CLI's `internal/pack`. The base include directive it builds on was restored in #1683 (merged). Thanks for the original work here; the behaviour you proposed now ships in v1.

PR #745, opened 2022-07-04
EXT-282-follow-project

corinnesollows:
# Checklist

## Changes
Added a warning in the error for Follow command that only Github and Bitbucket are supported at this time
Further commenting and formatting on the follow.go file

## Rationale
We are no longer supporting the 'follow' command going forward, but we do not wish to remove the legacy functionality

## Screenshots
image

### DIFF
--- a/cmd/follow.go
+++ b/cmd/follow.go
@@ -25,7 +25,7 @@ func followProject(opts options) error {
 	//check that project url contains github or bitbucket; our legacy vcs
 	if remote.VcsType == git.GitHub || remote.VcsType == git.Bitbucket {
 		vcsShort := "gh"
-		if remote.VcsType == "BITBUCKET" {
+		if remote.VcsType == git.Bitbucket {
 			vcsShort = "bb"
 		}
 		res, err := api.FollowProject(*opts.cfg, vcsShort, remote.Organization, remote.Project)
@@ -38,8 +38,10 @@ func followProject(opts options) error {
 			fmt.Println("Unable to determine project slug for CircleCI (slug is case sensitive).")
 		}
 
+	} else {
+		//if not warn user their vcs is not supported
+		return errors.New(errorMessage)
 	}
-	//if not warn user their vcs is not supported
 	return errors.New(errorMessage)
 }
 
--- a/cmd/follow.go
+++ b/cmd/follow.go
@@ -42,7 +42,7 @@ func followProject(opts options) error {
 		//if not warn user their vcs is not supported
 		return errors.New(errorMessage)
 	}
-	return errors.New(errorMessage)
+	return nil
 }
 
 //followProjectCommand follow cobra command creation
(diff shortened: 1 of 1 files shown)


cmd/follow.go
```diff
@@ -14,31 +14,36 @@ type options struct {
...
-	if err != nil {
-		return err
-	}
-	if res.Followed {
-		fmt.Println("Project successfully followed!")
-	} else if res.Message == "Project not found" {
-		fmt.Println("Unable to determine project slug for CircleCI (slug is case sensitive).")
-	}
+	//check that project url contains github or bitbucket; our legacy vcs
+	if remote.VcsType == git.GitHub || remote.VcsType == git.Bitbucket {
```
joeyorlando, line 26: I noticed there's similar logic in `cmd/open.go`. Would it make sense to add an `isLegacyVcs` method to the `Remote` struct in `git/git.go`?

cmd/follow.go
```diff
@@ -14,31 +14,36 @@ type options struct {
...
+			return err
+		}
+		if res.Followed {
+			fmt.Println("Project successfully followed!")
+		} else if res.Message == "Project not found" {
+			fmt.Println("Unable to determine project slug for CircleCI (slug is case sensitive).")
+		}
 
-	return nil
+	}
```
joeyorlando, line 41: I _think_ you might need an `else` clause here otherwise the conditions on lines 36 and 38 would print messages and then print an error

```suggestion
		return nil
	} else {
		//if not warn user their vcs is not supported
		return errors.New(errorMessage)
	}
```

cmd/follow.go
```diff
@@ -14,31 +14,36 @@ type options struct {
...
-	}
-	if res.Followed {
-		fmt.Println("Project successfully followed!")
-	} else if res.Message == "Project not found" {
-		fmt.Println("Unable to determine project slug for CircleCI (slug is case sensitive).")
-	}
+	//check that project url contains github or bitbucket; our legacy vcs
+	if remote.VcsType == git.GitHub || remote.VcsType == git.Bitbucket {
+		vcsShort := "gh"
+		if remote.VcsType == "BITBUCKET" {
```
joeyorlando, line 28:
```suggestion
		if remote.VcsType == git.Bitbucket {
```

cmd/follow.go
corinnesollows: That also would definitely work. While I was writing this I was under the assumption that it was part of an external package. That being said, I dont know if its worth going back and changing, since these are most likely going to be the only two places using this and we're looking to eventually get rid of VCSType completely.

cmd/follow.go
corinnesollows: sure I'll change this, I did not write it however :)

cmd/follow.go
corinnesollows: Yup, added

joeyorlando (approved):

PR #776, opened 2022-09-07
[SECENG-821] Get policy decision log by decision id

sagar-connect:
# Checklist

=========

## Changes

=======

- Added support for `decisionID` argument in existing `policy logs` command
- Added support for `--policy-bundle` flag in existing `policy logs` command
- Updated help text
- Updated unit-tests

## Rationale

=========

The new command will allow the users to fetch policy bundle associated with policy decision log

### DIFF
--- a/api/policy/policy.go
+++ b/api/policy/policy.go
@@ -174,15 +174,11 @@ func (c Client) GetDecisionLogs(ownerID string, context string, request Decision
 }
 
 // GetDecisionLog calls the GET decision query API of policy-service for a DecisionID.
-// It also accepts a getPolicyBundle bool param; If set to true will return only the policy bundle corresponding to that decision log.
-func (c Client) GetDecisionLog(ownerID string, context string, decisionID string, getPolicyBundle bool) (interface{}, error) {
+// It also accepts a policyBundle bool param; If set to true will return only the policy bundle corresponding to that decision log.
+func (c Client) GetDecisionLog(ownerID string, context string, decisionID string, policyBundle bool) (interface{}, error) {
 	path := fmt.Sprintf("%s/api/v1/owner/%s/context/%s/decision/%s", c.serverUrl, ownerID, context, decisionID)
-	var err error
-	if getPolicyBundle {
-		path, err = url.JoinPath(path, "policy-bundle")
-		if err != nil {
-			return nil, err
-		}
+	if policyBundle {
+		path = fmt.Sprintf("%s/%s", path, "policy-bundle")
 	}
 	req, err := http.NewRequest("GET", path, nil)
 	if err != nil {
--- a/api/policy/policy.go
+++ b/api/policy/policy.go
@@ -173,6 +173,44 @@ func (c Client) GetDecisionLogs(ownerID string, context string, request Decision
 	return body, nil
 }
 
+// GetDecisionLog calls the GET decision query API of policy-service for a DecisionID.
+// It also accepts a getPolicyBundle bool param; If set to true will return only the policy bundle corresponding to that decision log.
+func (c Client) GetDecisionLog(ownerID string, context string, decisionID string, getPolicyBundle bool) (interface{}, error) {
+	path := fmt.Sprintf("%s/api/v1/owner/%s/context/%s/decision/%s", c.serverUrl, ownerID, context, decisionID)
+	var err error
+	if getPolicyBundle {
+		path, err = url.JoinPath(path, "policy-bundle")
+		if err != nil {
+			return nil, err
+		}
+	}
+	req, err := http.NewRequest("GET", path, nil)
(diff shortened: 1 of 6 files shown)


api/policy/policy.go
```diff
@@ -173,6 +173,40 @@ func (c Client) GetDecisionLogs(ownerID string, context string, request Decision
 	return body, nil
 }
 
+// GetDecisionLog calls the GET decision query API of policy-service for a DecisionID.
+// It also accepts a policyBundle bool param; If set to true will return only the policy bundle corresponding to that decision log.
+func (c Client) GetDecisionLog(ownerID string, context string, decisionID string, policyBundle bool) (interface{}, error) {
+	path := fmt.Sprintf("%s/api/v1/owner/%s/context/%s/decision/%s", c.serverUrl, ownerID, context, decisionID)
+	if policyBundle {
+		path = fmt.Sprintf("%s/%s", path, "policy-bundle")
```
davidmdm, line 181: nit
```go
path += "/policy-bundle"
```

cmd/policy/policy.go
```diff
@@ -163,13 +163,23 @@ func NewCommand(config *settings.Config, preRunE validator.Validator) *cobra.Com
 	}()
 
 	logs := func() *cobra.Command {
-		var after, before, outputFile, ownerID, context string
+		var after, before, outputFile, ownerID, context, decisionID string
+		var policyBundle bool
 		var request policy.DecisionQueryRequest
 
 		cmd := &cobra.Command{
```
davidmdm, line 170: Args with range 0 to 1 ?

cmd/policy/policy.go
```diff
@@ -163,13 +163,23 @@ func NewCommand(config *settings.Config, preRunE validator.Validator) *cobra.Com
...
-			Use:   "logs",
-			RunE: func(cmd *cobra.Command, _ []string) (err error) {
+			Short: "Get policy decision logs / Get decision log (or policy bundle) by decision ID",
+			Use:   "logs [decision_id]",
+			RunE: func(cmd *cobra.Command, args []string) (err error) {
+				if len(args) == 1 {
+					decisionID = args[0]
+				}
+				if decisionID != "" && (after != "" || before != "" || request.Status != "" || request.Offset != 0 || request.Branch != "" || request.ProjectID != "") {
+					return fmt.Errorf("filters are not accepted when decision_id is provided")
```
davidmdm, line 178: I consider this to be almost a little too defensive. If the decision ID is present, I would just ignore the filters...

Either way is ok. Maybe get more than 1 opinion on this :)

cmd/policy/policy.go
```diff
@@ -163,13 +163,23 @@ func NewCommand(config *settings.Config, preRunE validator.Validator) *cobra.Com
...
+				}
+				if policyBundle && decisionID == "" {
```
davidmdm, line 180: same here for me... If no decision ID than I would just ignore policy-bundle...

cmd/policy/policy.go
```diff
@@ -163,13 +163,23 @@ func NewCommand(config *settings.Config, preRunE validator.Validator) *cobra.Com
...
+		var policyBundle bool
 		var request policy.DecisionQueryRequest
 
 		cmd := &cobra.Command{
-			Short: "Get policy decision logs",
-			Use:   "logs",
-			RunE: func(cmd *cobra.Command, _ []string) (err error) {
+			Short: "Get policy decision logs / Get decision log (or policy bundle) by decision ID",
+			Use:   "logs [decision_id]",
+			RunE: func(cmd *cobra.Command, args []string) (err error) {
```
davidmdm, line 173: I would probably try and break this function into two instead of having a big `if` block.

```go
if decisionID != "" {
  return getDecisionByID(...)
}
return getLogList(...)
```

davidmdm (approved):
Added a couple comments. Approving so as to not be blocking, but I would urge to think on them, or gather find another person to either way in on some of my more subjective opinions... @davidmparrott @vallieres

cmd/policy/policy.go
sagar-connect: It is configured as such:
`Args:    cobra.MaximumNArgs(1),`

cmd/policy/policy.go
sagar-connect: The real benefit of this check is the feedback it gives to the user, added defensiveness is just a bonus.
I would like to keep it for its documentation value to the user.

cmd/policy/policy.go
sagar-connect: It lets the user know that currently we support policy-bundle fetching for single decision ID only.
I don't see any issues in keeping it.

vallieres (approved):
LGTM!

LGTM

PR #791, opened 2022-10-03
[SECENG-839] added command to get/set policy settings

sagar-connect:
# Checklist

=========

## Changes

=======

- Added `settings` command, which allow reading/writing settings for config-policy-management

## Rationale

=========

- This command will allow our users to enable / disable policy decision as part of their build pipeline.


api/policy/policy.go
```diff
@@ -216,6 +216,69 @@ type DecisionRequest struct {
...
+	if err := json.NewDecoder(resp.Body).Decode(&body); err != nil {
+		return nil, fmt.Errorf("failed to decode response body: %v", err)
+	}
+
+	return body, nil
+}
+
+// DecisionSettings represents a request to Policy-Service to configure decision settings.
+type DecisionSettings struct {
+	Enabled *bool `json:"enabled,omitempty"`
```
davidmdm, line 251: We don't need a `*bool` from the client side, since we know we MUST send a boolean value and not null. On the server we had to make the distinction between false and absent value

sagar-connect: This is to have a tri-state bool. In case, the user wants to modify other settings, but keep enabled as-is, client won't send the enabled flag in the body.

cmd/policy/policy.go
```diff
@@ -380,12 +380,63 @@ This group of commands allows the management of polices to be verified against b
...
+			RunE: func(cmd *cobra.Command, args []string) error {
+				client := policy.NewClient(*policyBaseURL, config)
+
+				//Set settings
+				if cmd.Flag("enabled").Changed {
+					request.Enabled = &enabled
+					if err := client.SetSettings(ownerID, context, request); err != nil {
+						return fmt.Errorf("failed to set settings : %w", err)
+					}
+					_, _ = io.WriteString(cmd.ErrOrStderr(), "OK")
```
davidmdm, line 404: I would write a message that contains whether it was enabled or disabled

davidmdm (approved):

api/policy/policy.go
davidmdm: We should update our Rest Verbage to Patch then before releasing

sagar-connect: I was about to say the exact same thing. Sure, I'll do it.

Issue #676, opened 2022-03-18
[circleci local execute] error looking up cgroup: not implemented for cgroup v2 unified hierarchy - 2022 edition


matschaffer:
## Meta:

**CircleCI CLI Version:** 0.1.16947+263902a (homebrew)

**Operating System:** macOS Monterey - 12.3 (21E230)

 Model Name:	MacBook Pro
  Model Identifier:	MacBookPro16,1
  Processor Name:	8-Core Intel Core i9
  Processor Speed:	2.4 GHz
  Number of Processors:	1
  Total Number of Cores:	8
  L2 Cache (per Core):	256 KB
  L3 Cache:	16 MB
  Hyper-Threading Technology:	Enabled
  Memory:	64 GB
  System Firmware Version:	1731.100.130.0.0 (iBridge: 19.16.14242.0.0,0)

**CircleCI CLI Diagnostic:**
```
Debugger mode: false
Config found: /Users/matschaffer/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API...
Ok.
Hello, Mat Schaffer.
```

---

## Current behavior:

When running from https://github.com/LLK/scratch-gui

```
❯ circleci local execute
Docker image digest: sha256:ea206e01751a9d4acc6dc21fb8184bdf67b0c2a524c156e526bdb991b1bd2e23
====>> Spin up environment
Build-agent version  ()
System information:
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Kernel Version: 5.10.104-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64

Starting container circleci/node:10-browsers
Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.
  image is cached as circleci/node:10-browsers, but refreshing...
10-browsers: Pulling from circleci/node
Digest: sha256:f6e04fb16889bc6c91eda5aeaa62bfa2c995bb1a6f7124db358a148fd05418b5
Status: Image is up to date for circleci/node:10-browsers
circleci/node:10-browsers:
  using image circleci/node@sha256:f6e04fb16889bc6c91eda5aeaa62bfa2c995bb1a6f7124db358a148fd05418b5
  pull stats: Image was already available so the image was not pulled
  time to create container: 62ms
[...]

Issue #589, opened 2021-04-16
[circleci local execute] error looking up cgroup: not implemented for cgroup v2 unified hierarchy


ldorau:
## Meta:

**CircleCI CLI Version:**
0.1.15195+595cc9e (release)

**Operating System:**
Arch Linux (latest)

**CircleCI CLI Diagnostic:**
```
---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: /home/ldorau/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API... 
Ok.
Hello, Lukasz Dorau.
```

---

## Current behavior:
```
$ circleci local execute
Docker image digest: sha256:fc6b14595a231dd8417c97fad26f5f5f294f64b6f58443d93dfba17186f9f738
====>> Spin up environment
Build-agent version  ()
System information:
 Server Version: 20.10.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
 Cgroup Driver: systemd
 Cgroup Version: 2
 Kernel Version: 5.9.14-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64

Error: 
Unexpected environment preparation error: error looking up cgroup: not implemented for cgroup v2 unified hierarchy

Step failed
Task failed
Error: Unhandled prepare executor error: error looking up cgroup: not implemented for cgroup v2 unified hierarchy
```

## Expected behavior:
"circleci local execute" works

## When did this begin?
Now

## Was this previously working?:
No, never.

mshang816:
got the same issue. I have docker installed and working properly.

drownbes:
Same for me
`
Build-agent version  ()
System information:
 Server Version: 20.10.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
 Cgroup Driver: systemd
 Cgroup Version: 2
 Kernel Version: 5.12.9-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
`

dreamsmasher:
Same here.

```
> circleci version
0.1.15384+149d854 (release)

Build-agent version  ()
System information:
 Server Version: 20.10.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
 Cgroup Driver: systemd
 Cgroup Version: 2
 Kernel Version: 5.12.12-arch1-1
 Operating System: Arch Linux
[...]

PR #826, opened 2023-01-16
feat: enabled the configuration of the build-agent through argument of config file

JulesFaucherre:
# Checklist

## Changes

For the commands `circleci build` and `circleci local execute`, added the arguments `--picard-version` that enable to use a version of the picard image to use for the build. This option can be configured by creating a file in `$HOME/.circleci/build_agent_settings.json` which would contains ```{"LatestSha256":"<your-image>"}```

## Rationale

Because the latest picard image has actually some errors on the latest mac versions, a solution was use an old version of the picard executor. There was no way to do it with the actual version of the CLI, this pr allows this configuration
More information about this issue can be found the following issues:
 - https://github.com/CircleCI-Public/circleci-cli/issues/676
 - https://github.com/CircleCI-Public/circleci-cli/issues/589

## Screenshots

Capture d’écran 2023-01-16 à 11 32 50

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -16,6 +16,8 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 	}
 
 	local.AddFlagsForDocumentation(buildCommand.Flags())
+	buildAgentVersionUsage := `The version of the build agent image you want to use. This can be configured by writing in $HOME/.circleci/build_agent_settings.json: '{"LatestSha256":"<version-of-build-agent>"}'`
+	buildCommand.Flags().String("build-agent-version", "", buildAgentVersionUsage)
 	buildCommand.Flags().StringP("org-slug", "o", "", "organization slug (for example: github/example-org), used when a config depends on private orbs belonging to that org")
 	buildCommand.Flags().String("org-id", "", "organization id, used when a config depends on private orbs belonging to that org")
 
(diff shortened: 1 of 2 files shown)


local/local.go
```diff
@@ -155,6 +160,27 @@ func picardImage(output io.Writer) (string, error) {
...
+		return picardVersion, nil
+	}
+
+	var sha string
+	var err error
+
+	sha, err = loadBuildAgentShaFromConfig()
+	if sha != "" && err == nil {
+		return sha, nil
+	}
```
rlegan, line 175: if `err != nil` then we never inform the user its configuration file is not valid, shouldn't we ? 
either told them that the configuration could not have been red/parsed, and we're using the latest picard version, or crash

JulesFaucherre: True, let's add a message

cmd/build.go
```diff
@@ -16,6 +16,8 @@ func newLocalExecuteCommand(config *settings.Config) *cobra.Command {
 	}
 
 	local.AddFlagsForDocumentation(buildCommand.Flags())
+	picardVersionUsage := `The version of the build agent image you want to use. This can be configured by writing in $HOME/.circleci/build_agent_settings.json: '{"LatestSha256":"<version-of-picard>"}'`
```
rlegan, line 19: `picardVersionUsage` -> `buildAgentVersionUsage`

PR #855, opened 2023-02-28
task: Print filled pipeline values on validate & process

zbenhadi:


JulesFaucherre (approved):

wyardley:
Hi - could you all take a look at #860 / #861 (would also be good to get a little more information on what the goal of this change was). As mentioned by other commenters in #860, this has some worse effects than just being hard to read.

PR #861, opened 2023-03-08
feat: allow hiding pipeline values in validation output

wyardley:
# Checklist

=========

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have checked for similar issues and haven't found anything relevant.
- [ ] This is not a security issue (which should be reported here: https://circleci.com/security/)
- [ ] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/main/CONTRIBUTING.md).

### Internal Checklist
- [ ] I am requesting a review from my own team as well as the owning team
- [ ] I have a plan in place for the monitoring of the changes that I am making (this can include new monitors, logs to be aware of, etc...)

## Changes
=======
- Add a `-q` / `--quiet` flag to support suppressing the pipeline values added in #855
- Print the config values one line at a time, and add newlines

## Rationale

Fixes #860

## Considerations

==============

Why you made some of the technical decisions that you made, especially if the
reasoning is not immediately obvious

## Screenshots

============

<h4>Before</h4>
```
% circleci config validate              
Validating config with following values
	project.type:	github	git.tag:		git.branch:	wyardley/860	git.revision:	9ee9c17e1f5308fefa8e2a79eb017f24d00f1c04	git.base_revision:	9ee9c17e1f5308fefa8e2a79eb017f24d00f1c04	id:	00000000-0000-0000-0000-000000000001	number:	1	project.git_url:	https://github.com/wyardley/circleci-cliConfig file at .circleci/config.yml is valid.
```

<h4>After</h4>
```
% go run ./main.go config validate  -q
Config file at .circleci/config.yml is valid.
% go run ./main.go config validate    
Processing config with following values
git.branch        : wyardley/860
git.revision      : 9ee9c17e1f5308fefa8e2a79eb017f24d00f1c04
git.base_revision : 9ee9c17e1f5308fefa8e2a79eb017f24d00f1c04
id                : 00000000-0000-0000-0000-000000000001
number            : 1
project.git_url   : https://github.com/wyardley/circleci-cli
project.type      : github
git.tag           : 
Config file at .circleci/config.yml is valid.
```

Issue #860, opened 2023-03-08
messy validation output


wyardley:
## Meta:

**CircleCI CLI Version:**
```
0.1.23845+f59d90e (homebrew)
```

**Operating System:**
Mac OS X

**CircleCI CLI Diagnostic:**
```
---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: /Users/xx/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API... 
Ok.
Hello, xx.
```

## Current behavior:
```
% circleci config validate --org-slug github/joinforma    
Validating config with following values
	id:	00000000-0000-0000-0000-000000000001	number:	1	project.git_url:	https://github.com/xx/yyy	project.type:	github	git.tag:		git.branch:	foo/bar	git.revision:	b30ac0a34489c02db25b3d1a25cf4866f899fa45	git.base_revision:	b30ac0a34489c02db25b3d1a25cf4866f899fa45Config file at .circleci/config.yml is valid.
```

## Expected behavior:
This output should only be shown when a special flag is passed, _or_ there should be a config option to suppress it
It should also be formatted more nicely, and the validity statement at the end should be on a line by itself

## When did this begin / Was this previously working?:
Begin with the most recent update (post #855)

particleflux:
Not only is it potentially undesired output, it _also_ generates potentially yaml on `circleci config process`, as the line does not start with a comment character. (and it is output to stdout, so a redirect to a file will pickup both the processed config _and_ the metadata output)

wyardley:
@particleflux d'oh. yeah, I hadn't even looked at those subcommands or realized they were also printing this.

Hopefully someone from Circle will give some more context about whether this change is intentional and / or what the behavior should be.
In the meantime, I updated this to write to stdout.

PR #899, opened 2023-04-05
Additional formatting updates for config validation

wyardley:
# Checklist

=========

### Internal Checklist

## Changes
- Suppress some whitespace when verbose output is not enabled
- Write to stderr again
- Use `%v` instead of `%s` so that "number" in pipeline values gets printed correctly
- Sort values to provide stable output order for pipeline values

## Rationale
@elliotforbes sorry - I think one or both of us missed that `number` doesn't get printed out right when `%s` is used vs `%v`. This also moves most of the verbose output to stderr (esp. could be an issue for subcommands that output yaml otherwise). It also gets rid of the extra newline before validate output in the case where verbose output is not used

See #860 for original context. Resolve some issues from #861 and #896

## Screenshots

============

<h4>Before</h4>
image

<h4>After</h4>
image

### DIFF
--- a/config/commands.go
+++ b/config/commands.go
@@ -3,6 +3,7 @@ package config
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


config/commands.go
```diff
@@ -3,16 +3,27 @@ package config
...
 	"gopkg.in/yaml.v3"
 )
 
 func printValues(values Values) {
-	for key, value := range values {
-		fmt.Fprintf(os.Stderr, "%-18s %s\n", key+":", value)
+	// Provide a stable sort order for printed values
+	keys := make([]string, 0, len(values))
+	for k := range values {
+		keys = append(keys, k)
```
wyardley, line 17: open to suggestion here; I didn't find a way to do this in place for the original map.

config/commands.go
```diff
@@ -3,16 +3,27 @@ package config
...
 	}
+	sort.Strings(keys)
```
elliotforbes, line 19: This is a nice improvement 👌🏼

config/commands.go
elliotforbes: I think this is fine - it's nice and explicit what's going on and we're not necessarily resource constrained here

elliotforbes (approved):

JulesFaucherre (approved):
Great improvement, thanks a lot!

PR #915, opened 2023-04-20
First attempt at adding a fallback option which queries graphQL should the http request fail with status 404

elliotforbes:
# Checklist

=========

### Internal Checklist

## Changes
=======

- Adds the legacy graphQL ConfigQuery path back to the code. Whenever the first, preferred path, fails with a 404 status code, we then default back to the graphQL resolver such that there is no need for customers to pin to a current version as it stands.

## Rationale
=========

The decision to opt for pinned versions of the CLI has caused some customer impact and as such, we've opted to change tact such that the config process and validate commands run the new path first and fail back to the legacy path.

## Considerations

==============

We're working through testing this now with server 3 and server 4 customers.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -72,17 +72,18 @@ func TestCompiler(t *testing.T) {
 			assert.Contains(t, err.Error(), "Could not load config file at testdata/nonexistent.yml")
 		})
 
-		t.Run("handles 404 status correctly", func(t *testing.T) {
-			svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-				w.WriteHeader(http.StatusNotFound)
-			}))
-			defer svr.Close()
-			compiler := New(&settings.Config{Host: svr.URL, HTTPClient: http.DefaultClient})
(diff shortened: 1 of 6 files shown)


JulesFaucherre (approved):

config/config_test.go
```diff
@@ -72,17 +72,18 @@ func TestCompiler(t *testing.T) {
...
-			}))
-			defer svr.Close()
-			compiler := New(&settings.Config{Host: svr.URL, HTTPClient: http.DefaultClient})
-
-			_, err := compiler.ConfigQuery("testdata/config.yml", "1234", Parameters{}, Values{})
-			assert.Error(t, err)
-			assert.Contains(t, err.Error(), "this version of the CLI does not support your instance of server")
-		})
+		// commenting this out - we have a legacy_test.go unit test that covers this behaviour
+		// t.Run("handles 404 status correctly", func(t *testing.T) {
```
liamclarkedev, line 76: ~ What do you think about testing that the legacy path is only hit when we receive a 404 from the new client?

liamclarkedev (approved):

PR #922, opened 2023-04-25
Add org-slug and org-id flags to orb validate & process commands

zbenhadi:
# Checklist

=========

### Internal Checklist

## Changes

=======

- Added a new orbOrgOptions struct, to represent org-slug and org-id values
- Added a new api package: collaborators - presents an interface with a single method "GetOrgCollaborations"
- Changed "ValidateOrb" GQLQuery to include "ownerId" param, based on org-slug and org-id flags

## Rationale

=========

Integrate the capability to validate an orb when using other private orbs from the same namespace.


cmd/orb.go
```diff
@@ -1746,3 +1792,29 @@ func orbInformThatOrbCannotBeDeletedMessage() {
...
 }
+
+func (o *orbOptions) getOrgId(orgInfo orbOrgOptions) (string, error) {
+	if orgInfo.OrgID == "" && orgInfo.OrgSlug == "" {
+		return "", nil
+	}
+
+	var orgID string
+	if strings.TrimSpace(orgInfo.OrgID) != "" {
+		orgID = orgInfo.OrgID
```
JulesFaucherre, line 1803: Nitpicking: you could return directly here, avoiding too much nesting with the following `else if`

JulesFaucherre (changes requested):
Capture d’écran 2023-04-26 à 14 30 17
It seems the usage ask to use the `<vcs-name>/<org-name>` but the code only supports the `<vcs-short>/<org-name>` syntax

cmd/orb.go
zbenhadi: Addressed 👍

JulesFaucherre (approved):
Very great job, thanks for also fixing the slug issue on config validate!

PR #927, opened 2023-05-03
[API-677] Return correct error in ConfigQuery

professor-l:
# Checklist

=========

### Internal Checklist

## Changes

=======

- Log `originalErr` when `originalErr` is non-nil

## Rationale

=========

Attempting to validate a config while passing an invalid token fails with the following error:

```
$ circleci --token "invalid-token" config validate .circleci/config.yml -v
...
Error: config compilation request returned an error: %!w(<nil>)
```

The behavior would previously check if an error existed, but return a different one. After fixing that, the error prints correctly:

```
Error: config compilation request returned an error: invalid character 'I' looking for beginning of value
```

It's worth noting that the PR template explicitly instructs branching off of `main`, so if this should be rebased to `develop` instead then the template may also need updating.


nbialostosky (approved):
Approving commit e255781a96f53249474336d5486bf720bbd10c4e as it should address the error we were seeing before. We tested this locally too on a call.

JulesFaucherre (approved):
Thanks for fixing that!

PR #947, opened 2023-05-18
Add tmp_mount to cli.yaml and to setup command.

ceusebi-eb:
# Checklist

=========

### Internal Checklist

## Changes

=======

- Add a new parameter to `generateDockerCommand` to change the mount point of the tmp folder when running docker.
- Add a new entry to `cli.yaml` called `tmp_mount`
- Add a new flag to `setup` command called `--tmpMount` to set above entry

## Rationale

=========

Different docker servers will mount different tmp directories by default, and they might differ from the hardcoded `/tmp`.  
For example https://github.com/abiosoft/colima mounts `/tmp/colima` and it's unable to mount `/tmp` due to macOs restrictions.  
Storing files and using a path that we can't mount results in errors when executing `local` jobs.   
This change allows for users to pass along their mount path to the setup, and fixing the `local` execution of jobs

## Considerations

==============

- Decided to expose this in the setup as that's the command used to generate the file
- Kept the `/tmp` path as the default mount path

## Screenshots

============

<h4>Before</h4>
image

<h4>After</h4>
image


stevenreynolds-eb (approved):

JulesFaucherre:
Hey, thanks for the proposition. Actually it would make more sense to have "tmp_mount" defined as a parameter of `local execute` and not a global configuration as it is used only by this command.

I imagine your need to define it as a global setting was to not have to pass the argument all the time, for that we can consider having an environment variable that could be used as the argument when defined like `CIRCLECI_TMP_MOUNT`

loderunner:
Replaced by #967. Closing.

PR #949, opened 2023-06-12
Release

JulesFaucherre:


abdelDriowya (approved):

cmd/config.go
```diff
@@ -146,3 +160,27 @@ func packConfig(args []string) error {
 func migrateConfig(args []string) error {
 	return proxy.Exec([]string{"config", "migrate"}, args)
 }
+
+func generateConfig(args []string) error {
+	path := "."
+	if len(args) == 1 {
+		path = args[0]
+	}
```
loderunner, line 168: Using `.` for a path is a kind of POSIX-centric shortcut for current working directory. It should work in most settings, but I'd rather go for an approach based on the `os` package, that would do the right thing in every OS.

```suggestion
	var err error
	var path string
	if len(args) == 0 {
		// use working directory as default
		path, err = os.Getwd()
		if err != nil {
			return fmt.Errorf("couldn't get working directory")
		}
	} else {
		path = args[0]
	}
```

What do you think? Worth changing this before release?

PR #951, opened 2023-06-13
use os.Getwd instead of "." for current directory

loderunner:
# Checklist

=========

### Internal Checklist

## Changes

Replace usage of `"."` literal string as current working directory, with usage of [`os.Getwd`](https://pkg.go.dev/os#Getwd) in `config generate`.

## Rationale

`.` as a path to the current directory is a convention, shared by UNIX and Windows. Instead of relying on this convention, it seems more robust to rely on the standard library to obtain an absolute working path, in an os-specific fashion.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -162,8 +162,15 @@ func migrateConfig(args []string) error {
 }
 
 func generateConfig(args []string) error {
-	path := "."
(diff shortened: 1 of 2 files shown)


JulesFaucherre (approved):
Great! I believe it would be great to let @skaasten know about this change

Issue #716, opened 2022-05-30
Failed to read config file: read /tmp/local_build_config.yml: is a directory


samahatw:
## Meta:

**CircleCI CLI Version:**
0.1.18678+c3ad28b (homebrew)

**Operating System:**
macOS Monterey 12.3.1

**CircleCI CLI Diagnostic:**
```
---
CircleCI CLI Diagnostics
---
Debugger mode: false
Config found: /Users/samaha/.circleci/cli.yml
API host: https://circleci.com
API endpoint: graphql-unstable
OK, got a token.
Trying an introspection query on API...
Ok.
Hello, Tiago Samaha.
```

---

## Current behavior:
I'm trying to run the simplest example ([foo_ci](https://circleci.com/docs/2.0/local-cli-getting-started/#setup-and-validate-our-first-config)). I got an error when I try to execute locally, `circleci local execute`. 

```
➜  foo_ci circleci local execute
Fetching latest build environment...
Docker image digest: sha256:2eb349056027b89d707b11304da7a77692338e15367d2be8f03cee6e4a83a69e
Error: failed to start event processor: failed to compute task config: failed to read config file: read /tmp/local_build_config.yml: is a directory
```
## Expected behavior:
Run the pipeline locally.

## When did this begin / Was this previously working?:
It never worked.

## Additional Information:
Macbook Pro M1 (arm64). Using colima to run docker engine.

```
➜  foo_ci docker version
Client: Docker Engine - Community
 Version:           20.10.16
 API version:       1.41
 Go version:        go1.18.2
 Git commit:        aa7e414fdc
 Built:             Wed May 11 16:22:17 2022
 OS/Arch:           darwin/arm64
 Context:           colima
 Experimental:      true

Server:
 Engine:
  Version:          20.10.11
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.15
  Git commit:       847da184ad5048b27f5bdf9d53d070f731b43180
  Built:            Wed Apr 13 23:41:08 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.5.8
  GitCommit:        1e5ef943eb76627a6d3b6de8cd1ef6537f393a71
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:
```

r4v5:
[...]

PR #967, opened 2023-07-18
feat: configure temporary directory for local execute

loderunner:
# Checklist

=========

### Internal Checklist

## Changes

During `circleci local execute`, make it possible to configure temporary directory to store config file.

## Rationale

This was needed to support alternative (aka "not Docker") container distributions on macOS, particularly [lima](https://github.com/lima-vm/lima)-based distributions, such as colima or Rancher Desktop, that do not support mounting `/tmp` (or other ["system paths"](https://github.com/lima-vm/lima/blob/d9f0c51db0e84576692da4dde242ea150e3c16ee/pkg/limayaml/validate.go#L130C1-L131)) to their Linux VM.

## Considerations

* the temporary directory can be configured as a flag to `local execute`, or inside the `cli.yml` settings file with the `temp_dir` property, otherwise it defaults to the [system's temporary dir](https://pkg.go.dev/os#TempDir)
  * (in order of precedence)
* inspired by #716 (thanks @samahatw!)
* removes hard dependency on `/tmp`, instead reverting to system temp dir, as Docker Desktop now adds `/var/folders` to its mount paths


brunkb:
Would like to see this change go in, or if someone could point me to the workaround. I'm on Intel Mac and Rancher Desktop.

mjftw:
> ###

+1

this would solve issues I see related to https://github.com/CircleCI-Public/circleci-cli/issues/716

JulesFaucherre (approved):
Glad this finally gets out!

Issue #751, opened 2022-07-13
`circleci orb validate` fails with private dependent orb from same namespace


jschwanz:
## Meta:

**CircleCI CLI Version:**
* using image circleci/circleci-cli@sha256:eee9b00ad8929a528171d773e2ffc9d9f82fcd31c2e89a3da9f22bf17bf54bfc
* 0.1.19666+830e7ff (release)

**Operating System:**
Ubuntu 20.04

---

## Current behavior:

I'm attempting to author a private orb that has a dependency on another private orb in the same namespace. When `circleci orb validate` runs as a part of `orb-tools` I get:

```
Error: Cannot find my-namespace/dep-orb@1.2.3 in the orb registry. Check that the namespace, orb name and version are correct.
```

## Expected behavior:

I expect the `orb validate` process to be able to validate the use of private orbs from the same namespace.

jschwanz:
This functionality already exists in `circleci config validate` with the `--org-id` and `--org-slug` arguments. It would be really nice to have it here, too.

rafaelqueiroz93:
Unfortunatelly I'm having the same problem.

Waiting for a solution

adamdmharvey:
We are having this problem with our orbs as well. Effectively, it means we can't really use the `circleci orb validate` command as most of our orbs cross-reference each other now.

mcommons1:
Is there any update on this issue?

davidbarkhuizen:
Wow. CircleCI totally sucks at actually implementing necessary core functionality.  This is just one example of many that I have come across now.  CircleCI - YOU NEED TO RAISE YOUR GAME!!!

ADUDevEx:
After the security breach we experienced in CircleCI we migrated to private orbs and are now blocked by this.

kelvintaywl:
thank you for your sharing folks!

I am tagging our developer experience team here @CircleCI-Public/developer-experience to see if we can offer any updates on this 🙏

praveenkeshri:
?

ADUDevEx:
Thank you @kelvintaywl for your support on this issue. Hopefully the developer-experience team will find time to update this.

tdeekens:
Thanks for the feedback @JulesFaucherre. Has there been any progress on this. Running into this is quite tedious to get around and I am sure fixing it would uplift how companies can adopt and use Orbs more easily.

JulesFaucherre:
[...]

PR #968, opened 2023-07-19
refactor: Moved the orb validation in a orb API module

JulesFaucherre:
# Checklist

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/main/CONTRIBUTING.md).

## Changes

Started implementing an orb API abstraction by moving the orb validation. The interest of this module is that it will handle the different server version and return a client that is matching the Server version used

## Rationale

[An issue](https://github.com/CircleCI-Public/circleci-cli/issues/751) had been opened before to allow users to validate orbs that use private orbs. An `ownerId` has been added to the GraphQL API to enable this. Because of this change to the API, this created a potential difference between Server customers and standard customers. We handled this potential difference in the API call code directly but decided that we would handle in a better way in the near future. This is the new version of how we handle this backward compatibility.

Previous Jira ticket: https://circleci.atlassian.net/browse/DEVEX-856

## Considerations

Created the `OrbClient` as a lazy-evaluated singleton because it allows it to be imported from anywhere in the code

## Screenshots

Not much to show, the behaviour is still the same


abdelDriowya (approved):
LGTM & TESTED

api/orb/client.go
```diff
@@ -0,0 +1,105 @@
...
+					Name string `json:"name"`
+				} `json:"args"`
+			} `json:"fields"`
+		} `json:"queryType"`
+	} `json:"__schema"`
+}
+
+func orbQueryHandleOwnerId(gql *graphql.Client) (bool, error) {
+	query := `
+query ValidateOrb {
```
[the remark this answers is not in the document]
loderunner: There are [several](https://www.apollographql.com/blog/graphql/golang/using-graphql-with-golang/) [Go](https://blog.khanacademy.org/genqlient-a-truly-type-safe-go-graphql-client/) [projects](https://sr.ht/~emersion/gqlclient/) that do client code generation from schema, with type safety and everything. Maybe we should use one of those, some day.

api/orb/deprecated.go
```diff
@@ -0,0 +1,50 @@
...
+
+	query := `
+		query ValidateOrb ($config: String!) {
+			orbConfig(orbYaml: $config) {
+				valid,
+				errors { message },
+				sourceYaml,
+				outputYaml
+			}
+		}`
```
loderunner, line 33: 🧑‍🏫 

```suggestion
	const query = `query ValidateOrb ($config: String!) {
  orbConfig(orbYaml: $config) {
    valid,
    errors { message },
    sourceYaml,
    outputYaml
  }
}`
```

api/orb/client.go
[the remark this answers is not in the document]
loderunner: Or maybe we could write a simple-as-duck query builder? Using [graph-go-tools](https://github.com/wundergraph/graphql-go-tools) perhaps.

api/orb/deprecated.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/CircleCI-Public/circleci-cli/api"
+	"github.com/CircleCI-Public/circleci-cli/api/graphql"
+	"github.com/pkg/errors"
+)
+
+type deprecatedClient struct {
+	gql *graphql.Client
+}
+
+func (deprecated *deprecatedClient) OrbQuery(configPath string, ownerId string) (*api.ConfigResponse, error) {
```
loderunner, line 13: Tiny naming proposal: Could it just be named `client` inside the context of this function?

```suggestion
func (client *deprecatedClient) OrbQuery(configPath string, ownerId string) (*api.ConfigResponse, error) {
```

api/orb/latest.go
```diff
@@ -0,0 +1,48 @@
...
+
+	query := `
+		query ValidateOrb ($config: String!, $owner: UUID) {
+			orbConfig(orbYaml: $config, ownerId: $owner) {
+				valid,
+				errors { message },
+				sourceYaml,
+				outputYaml
+			}
+		}`
```
loderunner, line 29:
```suggestion
	const query = `query ValidateOrb ($config: String!, $owner: UUID) {
  orbConfig(orbYaml: $config, ownerId: $owner) {
    valid,
    errors { message },
    sourceYaml,
    outputYaml
  }
}`
```

api/orb/yaml.go
loderunner, line 1: Not important, but it seems a bit overkill to have a full file for a single generic function.

Could `loadYaml` maybe live in `client.go`?

cmd/orb.go
```diff
@@ -732,7 +733,11 @@ func validateOrb(opts orbOptions, org orbOrgOptions) error {
 		return fmt.Errorf("failed to get the appropriate org-id: %s", err.Error())
 	}
 
-	_, err = api.OrbQuery(opts.cl, opts.args[0], orgId)
+	client := orb.GetClient(opts.cfg)
+	if client == nil {
+		return fmt.Errorf("Unable to validate orb")
```
loderunner, line 738: This error message obscures the underlying problem. Even though we may want to hide the inner workings from users, this could be too generic when the time comes to answer issues on GitHub.

cmd/orb.go
```diff
@@ -754,7 +759,11 @@ func processOrb(opts orbOptions, org orbOrgOptions) error {
 		return fmt.Errorf("failed to get the appropriate org-id: %s", err.Error())
 	}
 
-	response, err := api.OrbQuery(opts.cl, opts.args[0], orgId)
+	client := orb.GetClient(opts.cfg)
+	if client == nil {
+		return fmt.Errorf("Unable to validate orb")
```
loderunner, line 764: Same here.

loderunner (changes requested):
A lot of comments, but mostly nitpicking. The general idea is great.

However, I feel this implementation could prove to be hard to extend in the future. Without changing too much, I think we can make this future-proof. Let's discuss.

api/orb/client.go
```diff
@@ -0,0 +1,142 @@
...
+	case v1_string:
+		return &v1Client{gql}, nil
+	case v2_string:
+		return &v2Client{gql}, nil
+	default:
+		return nil, fmt.Errorf("Unable to recognise your server orb API")
+	}
+}
+
+func detectClientVersion(gql *graphql.Client) (clientVersion, error) {
```
loderunner, line 62: The internals of this function aren't necessarily clear. It might be worth commenting why "handling owner ID" results in being able to differentiate between v1 and v2. What do you think?

api/orb/client.go
```diff
@@ -0,0 +1,142 @@
...
+	OrbQuery(configPath string, ownerId string) (*api.ConfigResponse, error)
+}
+
+func GetClient(config *settings.Config) (Client, error) {
+	once.Do(func() {
+		client, clientError = newClient(config)
+	})
+
+	return client, clientError
+}
```
loderunner, line 42: As we discussed offline, I think this not-quite-singleton pattern is actually more harmful than having potentially more than one client.

The risk here is a caller calling `GetClient` a second time, with a different config, and getting a `Client` not suited to their config, without error.

In this implementation, I think it's worth having callers rely on an explicit `NewClient(config)` even if it means wastefully creating clients, rather than using `GetClient` expecting certain results and facing hard-to-undertand bugs.

loderunner (changes requested):
Very nice. I'm not sure about the "kinda singleton" pattern. Let's discuss if we want to keep it.

loderunner (approved):
All good. Address the 2nd conversation if you think it's useful, or just go ahead and merge.

PR #973, opened 2023-07-27
[SNC-392] compile support for policy test command

davidmdm:
# Checklist

=========

### Internal Checklist

## Changes

=======

- Updated the CircleCI-Public/circle-policy-agent
- Modified the `circleci policy test` command by adding the policy-agent compile function
  - calls the orbs-service API via graphql to compile test input
- Fixes a bug in the policy API client that was causing failures across tests
- adds the compily policy and test to validate the compilation is working in e2e fashion

## Rationale

=========

To provide a better testing experience of CircleCI Config Policies, we needed to update the policy-agent that understood the compile and pipeline_parameters directive, and provide a compilation function for it to use.

### DIFF
--- a/cmd/policy/policy_test.go
+++ b/cmd/policy/policy_test.go
@@ -173,7 +173,7 @@ func TestPushPolicyBundleNoPrompt(t *testing.T) {
 			svr := httptest.NewServer(tc.ServerHandler)
 			defer svr.Close()
 
-			cmd, stdout, stderr := makeCMD("")
+			cmd, stdout, stderr := makeCMD("", "testtoken")
 
 			cmd.SetArgs(append(tc.Args, "--policy-base-url", svr.URL, "--no-prompt"))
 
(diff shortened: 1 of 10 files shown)


jbialy (approved):
From a high level it lgtm!

cmd/policy/policy_test.go
```diff
@@ -1330,13 +1331,31 @@ func TestTestRunner(t *testing.T) {
 				assert.Contains(t, s, "<?xml")
 			},
 		},
+		{
```
sagar-connect, line 1334: Test suggestion:
`compile: true`, but owner-id isn't provided.

sagar-connect (approved):

cmd/policy/policy_test.go
davidmdm: That is the current test case

rooneyshuman (approved):

cmd/policy/policy_test.go
sagar-connect: My understanding is if compilation is required, but owner-id isn't provided, it should fail in some way.
But the expectation is pass in this test?

davidmdm: Owner-ID is only required if you need to pull a private orb

JulesFaucherre (approved):

PR #977, opened 2023-08-01
refactor: Abstracted the different config validation routes

JulesFaucherre:
# Checklist

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/main/CONTRIBUTING.md).

## Changes

Abstracted the API that compiles config so that it's transparent whether you use the new API route or the GraphQL endpoint

## Rationale

The config compilation has recently been moved from the GraphQL to a new API route. At first it was implemented such as if the compilation request failed with a 404, we tried falling back to the legacy GraphQL endpoint because it mostly meant that we were on old Server version that did not have the new API route
Now you have to pass an `APIClient` to the config compiler upon creation. This client should be obtained with `config.NewWithConfig(*settings.Config)`. This API client is an interface which implementation either request the new API route or the GraphQL
To detect which implementation must be given, a call is made to the new route, on 404 the GraphQL implementation is returned, else the new API route implementation is returned

## Screenshots

Nothing to show as it only changes how the code operates internally but do not add any product change

### DIFF
--- a/api/rest/client.go
+++ b/api/rest/client.go
@@ -98,7 +98,7 @@ func (c *Client) DoRequest(req *http.Request, resp interface{}) (int, error) {
 		}{}
 		body, err := io.ReadAll(httpResp.Body)
 		if err != nil {
-			return httpResp.StatusCode, err
+			return 0, err
 		}
 		err = json.Unmarshal(body, &httpError)
 		if err != nil {
--- a/api/rest/client.go
+++ b/api/rest/client.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
@@ -97,7 +96,7 @@ func (c *Client) DoRequest(req *http.Request, resp interface{}) (int, error) {
 		httpError := struct {
 			Message string `json:"message"`
 		}{}
-		body, err := ioutil.ReadAll(httpResp.Body)
+		body, err := io.ReadAll(httpResp.Body)
 		if err != nil {
 			return httpResp.StatusCode, err
 		}
--- a/api/rest/client.go
+++ b/api/rest/client.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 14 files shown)


api/rest/client.go
```diff
@@ -96,10 +96,14 @@ func (c *Client) DoRequest(req *http.Request, resp interface{}) (statusCode int,
 		httpError := struct {
 			Message string `json:"message"`
 		}{}
-		err = json.NewDecoder(httpResp.Body).Decode(&httpError)
+		body, err := io.ReadAll(httpResp.Body)
 		if err != nil {
 			return httpResp.StatusCode, err
 		}
```
loderunner, line 102: Should we be returning an error from reading the response body along with the status code? It seems a little strange.

For example, we could have the server respond with a 200 OK, but for some reason, fail to read the body of the response. We'd then return `200` with an error message, unrelated to the HTTP protocol.

The way I see it, we should either return:
* 2xx HTTP status code - no error
* 4xx-5xx HTTP status code - HTTP error (from body or from status text)
* HTTP status code zero value - non-HTTP error

config/api_client.go
```diff
@@ -0,0 +1,55 @@
...
+	"net/url"
+
+	"github.com/CircleCI-Public/circleci-cli/api/graphql"
+	"github.com/CircleCI-Public/circleci-cli/api/rest"
+	"github.com/CircleCI-Public/circleci-cli/settings"
+)
+
+var (
+	compilePath = "compile-config-with-defaults"
+)
```
loderunner, line 14: This one doesn't seem to change. Can we make this a `const`?

```suggestion
const compilePath = "compile-config-with-defaults"
```

config/api_client.go
```diff
@@ -0,0 +1,55 @@
...
+	case v1_string:
+		return &v1APIClient{graphql.NewClient(config.HTTPClient, config.Host, config.Endpoint, config.Token, config.Debug)}, nil
+	case v2_string:
+		return &v2APIClient{restClient}, nil
+	default:
+		return nil, fmt.Errorf("Unable to recognise your Server's config file API")
+	}
+}
+
+func detectAPIClientVersion(restClient *rest.Client) (apiClientVersion, error) {
```
loderunner, line 41: I think the internals of this function could be explained. In a year's time, it's probable that another developer will face this and wonder why v1 results in a 404. 🙃 

Do you think a comment would be justified here?

loderunner (approved):
Very readable code, for a non-trivial problem. Well done! 👏 

Minor comments, none should be blocking. I'll let you decide what you want to do with them and resolve the conversations as needed.

config/api_client.go
JulesFaucherre: Agree! Thanks

PR #1008, opened 2023-09-25
Update wording of error to be slightly less confusing

jamonholmgren:
I found this error message to be slightly confusing -- like, I didn't know `circleci setup` was a command, although it's obvious in retrospect.

CleanShot 2023-09-25 at 16 44 58@2x

By changing the wording here, I think it's a bit more clear.


jamonholmgren:
(Will need to resubmit with tests passing -- thought I could get away with a quick wording update, but it looks like you have tests relying on it.)

PR #1013, opened 2023-10-26
feat: default http client timeout override

### LINKED ISSUE #1012: Allow configuring http client timeout
## What problem does this feature solve?:
- Our CI validates the circleci configuration size by running `circleci config process / validate` commands
- Ocassianally the `circleci` fails to execute due to timeout.
- Looking at the code it seems like the http client is configured to 10s timeout ([here](https://github.com/CircleCI-Public/circleci-cli/blob/v0.1.29041/api/rest/client.go#L43))
- I think this should be configurable by either env variable or global flag

## Provide an example:
```
export HTTP_TIMEOUT=30s
circleci config validate config.yml
```

or 
```
circleci --http_timeout 30s config validate config.yml
```

or-shachar:
# Checklist

=========

- [ x I have checked for similar issues and haven't found anything relevant.

## Changes
- Fixed #1012 
- Allows using `HTTP_TIMEOUT` env variable:
```bash
export HTTP_TIMEOUT=30s
circleci config validate config.yml
```

## Considerations

- 10 seconds should definitely be enough for the API server to respond
- This override feature is only for cases where there is some **temporal** misbehavior.

### DIFF
--- a/api/header/global.go
+++ b/api/header/global.go
@@ -1,5 +1,7 @@
 package header
 
+import "time"
+
 // When the CLI is initialized, we set this to a string of the current CLI subcommand
 // (e.g. `circleci orb list`) with no args or flags, so we include it as a header in API requests.
 var cliCommandStr string = ""
@@ -10,4 +12,10 @@ func SetCommandStr(commandStr string) {
 
 func GetCommandStr() string {
 	return cliCommandStr
+}
+
+const defaultTimeout = 60 * time.Second
+
+func GetDefaultTimeout() time.Duration {
+	return defaultTimeout
 }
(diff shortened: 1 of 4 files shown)


or-shachar:
@loderunner  - I've changed the name of the env variable to `CIRCLECI_CLI_TIMEOUT` - feel free to squash/merge  + introduce a new PR with changes to the GQL client as well 🙏

abdelDriowya:
Thanks @or-shachar for your contribution, I will add timeout to graphql client and we will merge the pr after that

JulesFaucherre (approved):
@abdelDriowya it would be great to add a little test on `NewFromConfig` to avoid regressions on this behaviour

PR #1034, opened 2024-01-08
update docs

georgelivas:
Update help command and docs for `create` command


cmd/namespace.go
```diff
@@ -86,13 +86,11 @@ Please note that at this time all namespaces created in the registry are world-r
 		},
 		Args:        cobra.RangeArgs(1, 3),
 		Annotations: make(map[string]string),
-		Example: `  circleci namespace create NamespaceName github OrgName
+		Example: `  circleci namespace create NamespaceName --org-id "your-org-id-here"
```
JulesFaucherre, line 89: I suppose this example was showing the two ways you could use the command which were `circleci namespace create NamespaceName <vcs-type> <org-name>` or `circleci namespace create NamespaceName --orgId <org-id>`.

This PR removes the first way of using the command so you should simply remove the first example instead of changing it.

JulesFaucherre (changes requested):
Little change on the example plus I believe you could take care of removing the `vcs-type/org-name` behaviour.
We can call if it's not clear.

cmd/namespace.go
```diff
@@ -84,15 +83,12 @@ Please note that at this time all namespaces created in the registry are world-r
 
 			return err
 		},
-		Args:        cobra.RangeArgs(1, 3),
+		Args:        cobra.RangeArgs(1, 1),
```
JulesFaucherre, line 86: For fixed number of arguments, cobra recommand using `cobra.ExactArgs` instead of `cobra.RangeArgs`.

JulesFaucherre (approved):
A minor remark but otherwise LGTM.

PR #1046, opened 2024-02-26
[DEVEX-1216] context commands can now be used with `--org-id` parameter.

JulesFaucherre:
[Jira ticket](https://circleci.atlassian.net/browse/DEVEX-1216)

## Changes

- Refactor of the context API so that it can use either `orgID` or `vcsType / orgName` to reference an org. The information is passed upon creation of the context API client.
- Allow all commands of `circleci context` to use the `--org-id`.
- Commands `circleci context store-secret`, `circleci context remove-secret`, `circleci context create` and `circleci context delete` show a text line upon success to inform user.

## Rationale

Having these commands use the `--org-id` parameter was necessary to support standalone projects.

## Considerations

The context API had to be adapted to handle using `orgID` instead of `vcsType / orgName`. A first approach was already implemented for `CreateContext`. The solution taken here was to add a method `CreateContextWithOrgID`.
After some reflexion, we thought that duplicating every call of the interface would not be very clear and would pass the decision of whether to use `orgID` or `vcsType / orgName` to the caller. Thus we decided to unite the interface into one that would be instantiated with the information about its org.

## Screenshots

N/A.


api/rest/client.go
```diff
@@ -129,6 +129,10 @@ func (c *Client) DoRequest(req *http.Request, resp interface{}) (int, error) {
 	return httpResp.StatusCode, nil
 }
 
+func (c *Client) Do(req *http.Request) (*http.Response, error) {
```
loderunner, line 132: I know it's not your fault that `rest.Client` is already a leaky abstraction, but it seems that wherever you call `Do`, you're doing things very similarly to `DoRequest`. 
Is there a way to re-use, maybe change, the existing `DoRequest` method, rather than creating a new, leakier method?

api/api.go
```diff
@@ -744,30 +741,54 @@ func CreateNamespaceWithOwnerID(cl *graphql.Client, name string, ownerID string)
 	return &response, nil
 }
 
-func getOrganization(cl *graphql.Client, organizationName string, organizationVcs string) (*GetOrganizationResponse, error) {
-	var response GetOrganizationResponse
+type GetOrganizationParams struct {
+	OrgName *string
+	VCSType *string
+	OrgID   *string
```
loderunner, line 747: Why use `*string` here? I suspect it's to use `nil` to mark the absence of the parameter?

If that's the case, `string` and its zero value `""` should be an adequate, idiomatic choice.

api/context/gql.go
```diff
@@ -0,0 +1,253 @@
...
+
+	"github.com/CircleCI-Public/circleci-cli/api/graphql"
+	"github.com/pkg/errors"
+)
+
+type ListContextsWithGQLParams struct {
+	OrgID   *string
+	OrgName *string
+	VCSType *string
+}
```
loderunner, line 16: Same here. I think it's much simpler to just use `string`.

api/context/gql_client.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/CircleCI-Public/circleci-cli/api"
+	"github.com/CircleCI-Public/circleci-cli/api/graphql"
+)
+
+type gqlClient struct {
+	client *graphql.Client
+
+	orgID   *string
+	vcsType *string
+	orgName *string
```
loderunner, line 15: Same here, I think `string` would work.

api/context/rest.go
```diff
@@ -0,0 +1,338 @@
...
+
+	"github.com/CircleCI-Public/circleci-cli/api"
+	"github.com/CircleCI-Public/circleci-cli/api/rest"
+)
+
+type ListContextsWithRestParams struct {
+	OwnerID   *string
+	OwnerSlug *string
+	OwnerType *string
+	PageToken *string
```
loderunner, line 18: Same here

api/context/rest.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+type listContextsResponse struct {
+	Items         []Context `json:"items"`
+	NextPageToken *string   `json:"next_page_token"`
+}
+
+// List all contexts for an owner.
+// Binding to https://circleci.com/docs/api/v2/index.html#operation/listContexts
+func ListContextsWithRest(client *rest.Client, params *ListContextsWithRestParams) (*listContextsResponse, error) {
```
loderunner, line 28: Why make `params` a pointer here? It will panic below if `params` is `nil`.

api/context/rest.go
```diff
@@ -0,0 +1,338 @@
...
+
+	dest := listContextsResponse{}
+	if err := json.Unmarshal(bodyBytes, &dest); err != nil {
+		return nil, err
+	}
+	return &dest, nil
+}
+
+// Gets all pages of ListContexts
+func ListAllContextsWithRest(client *rest.Client, params *ListContextsWithRestParams) ([]Context, error) {
```
loderunner, line 77: Why make `params` a pointer?

api/context/rest.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+var orgOwnerType = "organization"
+
+// Creates a new context.
+// Binding to https://circleci.com/docs/api/v2/index.html#operation/createContext
+func CreateContextWithRest(c *rest.Client, params *CreateContextWithRestParams) (*Context, error) {
+	if params == nil {
+		return nil, fmt.Errorf("parameters are nil when creating context")
+	}
```
loderunner, line 112: Why make `params` a pointer if you're going to return an error when it's `nil`?

api/context/rest_client.go
```diff
@@ -0,0 +1,105 @@
...
+	orgName *string
+}
+
+func toSlug(vcs, org string) *string {
+	slug := fmt.Sprintf("%s/%s", vcs, org)
+	return &slug
+}
+
+// createListContextsParams is a helper to create ListContextsParams from the content of the ContextRestClient
+func (c *restClient) createListContextsParams() (*ListContextsWithRestParams, error) {
```
loderunner, line 23: Why return a pointer, here?

api/context/rest.go
```diff
@@ -0,0 +1,338 @@
...
+
+type ListEnvVarsWithRestParams struct {
+	ContextID string
+	PageToken *string
+}
+
+type listEnvVarsResponse struct {
+	Items         []EnvironmentVariable `json:"items"`
+	NextPageToken *string               `json:"next_page_token"`
+}
```
loderunner, line 197: I think even your page tokens could be defined as `string`. In many cases, checking against `""` works just as well as checking against `nil`, with less complexity.

loderunner (changes requested):
Alright, the design seems great to me.

However I notice a pattern of using pointers a lot. Is there a justification for this that I'm missing? If it's to mark presence or absence of the variable, most of the time, the zero value works just as well, especially with `string`, where empty string often works just as well as `nil` (except in cases where empty string makes semantic sense).

If there's a justification, sorry for all the comments, let's discuss so I can understand better.

api/api.go
JulesFaucherre: Yes, I kinda ported this practice from the previous implementation, I'll remove it.

api/context/rest_client.go
JulesFaucherre: I think, as a Go practice I like to return a unusable value in case of error, returning a pointer here allow to return `nil` in case of error.

api/api.go
```diff
@@ -27,6 +27,10 @@ const (
 	Remove
 )
 
+type ErrorWithMessage struct {
+	Message *string `json:"message"`
+}
```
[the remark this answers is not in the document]
JulesFaucherre: Defined it as an error and used it for the `api/context/rest.go` file 👌 .

api/context/rest_client.go
loderunner: That's what the zero value is for. Returning `nil` is just returning the zero value of a pointer type.

For example, [here](https://go.dev/doc/tutorial/handle-errors), in the Go tutorial, the func `return "", errors.New()`. Other examples [here](https://github.com/golang/go/blob/master/src/net/tcpsock.go#L157-L159) and [here](https://github.com/golang/go/blob/master/src/strconv/atoi.go#L81-L83) from the Go sources.

api/context/rest.go
```diff
@@ -0,0 +1,247 @@
...
+	ContextID string
+	PageToken string
+}
+
+type listEnvVarsResponse struct {
+	Items         []EnvironmentVariable `json:"items"`
+	NextPageToken string                `json:"next_page_token"`
+}
+
+func ListEnvVarsWithRest(client *rest.Client, params ListEnvVarsWithRestParams) (*listEnvVarsResponse, error) {
```
loderunner, line 150: You could generalize for all return values. No need to return a pointer if you don't have anything hard to copy in the struct (like a mutex, a file or a network connection), or if you know the copying is a performance bottleneck. In many cases, not making it a pointer even lets the compiler do some cool optimizations.

Also, I think the return type should be exported, since it's returned from an exported func, and has exported fields.

```suggestion
func ListEnvVarsWithRest(client *rest.Client, params ListEnvVarsWithRestParams) (ListEnvVarsResponse, error) {
```

loderunner: This comment has sent me down a rabbit hole of when to:

* use pointers as receivers
* pass pointers as arguments
* return pointers from functions

Opinions vary, and many people inject different semantics into the simple `*` in Go. Optionals, mutability, pass-by-reference... Very different from Rust and its explicit `mut`, `&` references and `std::option`, or C++ and its many types of "smart pointers". We should probably take this offline and establish our own guide of "Pointer usage in Go". Hell, we could probably make a blog post out of it.

Do as you like for pointer types. You can resolve this, let's chat later.

loderunner (approved):
Looks good to me. Started a discussion that I think we could take offline. Resolve it and you can merge.

api/context/rest.go
JulesFaucherre: Nice, I don't know but maybe someone in CCI wrote some go practices for the company.

PR #1156, opened 2025-08-19
Fix cucumber tests for server

Klaney:
# Checklist

=========

- [ ] This is not a security issue (which should be reported here: https://circleci.com/security/)
- [ ] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/main/CONTRIBUTING.md).

### Internal Checklist

## Changes

=======

- Fix Snyk vulnerability by downgrading aruba to 0.14.14 and thor to 1.4.0
- Remove debug flag (-d) from Snyk configuration in CI/CD to prevent sensitive information warnings
- Add secure token injection for k9s server test using @k9s tag and Before/After hooks
- Fix git repository test by using shell command instead of aruba cd step
- Update Gemfile to specify compatible gem versions for Ruby 2.7

## Rationale

=========

This PR addresses a failing CI/CD pipeline caused by Snyk security vulnerability scans and an E2E cucumber test. The main branch had aruba 1.0.0 which depends on thor 1.0.1, creating an OS Command Injection vulnerability that was blocking deployments. 

Additionally, the k9s server test was failing because the server now requires authentication, while the integration tests weren't properly injecting the required token.

The overarching goal was to restore CI/CD pipeline stability while maintaining security best practices and ensuring the integration tests work reliably in both local and CI/CD environments.

## Considerations

==============

Ruby Version Compatibility: We chose to stay with Ruby 2.7 in CI/CD rather than upgrading to Ruby 3.x to minimize risk and avoid potential breaking changes. However, this creates technical debt that should be addressed in the future. The team should plan a migration to Ruby 3.x in a separate PR to:
- Take advantage of performance improvements
- Access newer gem versions with security patches
- Reduce maintenance burden of supporting older Ruby versions

Gem Version Strategy: We downgraded aruba from 1.0.0 to 0.14.14 instead of upgrading to the latest version because newer aruba versions have breaking changes that would require significant test refactoring. This is a pragmatic approach that fixes the immediate security issue while deferring the larger refactoring work.

Token Security: The k9s test token injection uses environment variables and Cucumber hooks rather than command-line arguments to ensure sensitive data isn't exposed in CI/CD logs. The token is properly scoped to only the @k9s tagged scenario and cleaned up after execution.

Aruba Compatibility: The git repository test required a workaround because aruba 0.14.14's cd step doesn't persist directory changes between steps. Using sh -c "cd /tmp && git status" ensures the directory change happens within the same shell session, which is more reliable than trying to fix aruba's directory handling.


integration_tests/features/circleci_config.feature
```diff
@@ -282,16 +283,15 @@ Feature: Config checking
     And the output should contain "command: date '+%Y-%m-%dT%T%z'"
 
   Scenario: Running validate in a directory that is not a git repo
-    When I cd to "/tmp"
+    When I run `sh -c "cd /tmp && git status"`
+    And the output should contain "fatal: not a git repository (or any of the parent directories): .git"
```
Klaney, line 287: Aruba @ 0.14.14 broke this test, this was the fix.

quinnallman (approved):

PR #1159, opened 2025-09-26
Store test results in CI

pete-woods:
This makes it much easier to see what tests have failed, and also will allow CircleCI to identify flaky tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,11 @@ jobs:
     steps:
       - checkout
       - gomod
-      - run: make test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -143,7 +143,11 @@ jobs:
     steps:
       - checkout
       - gomod
-      - run: make test
+      - run: gotestsum --junitfile test_results/linux.xml
```
pete-woods, line 146: Taken from Windows job

Klaney (approved):

PR #1193, opened 2026-03-17
Fix orb init failing when orb-publishing context already exists

### LINKED ISSUE #1192: circleci orb init fails with "A context already exists with this name" when orb-publishing context already exists
### Overview

`circleci orb init` fails when the `orb-publishing` context already exists in the organization (e.g., created by a teammate). The CLI has error handling for this case, but the expected error string doesn't match what the API actually returns, so the graceful "continuing on" path is never reached.

### Steps to Reproduce

1. Have an organization where the `orb-publishing` context already exists (e.g., a teammate previously ran `circleci orb init`)
2. Run:
   ```
   circleci orb init ./my-orb/
   ```
3. Walk through the interactive prompts
4. When asked "Automatically set up a publishing context for your orb?", select **"Yes, set up a publishing context with my API key."**

### Expected Behavior

The CLI should detect that the `orb-publishing` context already exists, print:

```
`orb-publishing` context already exists, continuing on
```

...and proceed with the rest of the setup.

### Actual Behavior

The CLI exits with:

```
Error: A context already exists with this name
```

### Root Cause

In [`cmd/orb.go` lines 1406-1412](https://github.com/CircleCI-Public/circleci-cli/blob/main/cmd/orb.go#L1406-L1412), the error handling checks for the string `"A context named orb-publishing already exists"`:

```go
err = contextAPI.CreateContext("orb-publishing")
if err != nil {
    if strings.Contains(err.Error(), "A context named orb-publishing already exists") {
        fmt.Println("`orb-publishing` context already exists, continuing on")
    } else {
        return err
    }
}
```

However, the CircleCI REST API (used for Cloud via [`NewContextClient`](https://github.com/CircleCI-Public/circleci-cli/blob/main/api/context/context.go#L49-L50)) returns a different error message: **`"A context already exists with this name"`**.

Since `strings.Contains("A context already exists with this name", "A context named orb-publishing already exists")` is `false`, the check fails and the error falls through to the `else` branch, which returns it and aborts the command.

**Note:** The GQL path (used for Server when REST is unavailable) would return `"Error creating context: ALREADY_EXISTS"` via [`gql.go` line 139](https://github.com/CircleCI-Public/circleci-cli/blob/main/api/context/gql.go#L139), which also wouldn't match.

### Suggested Fix

Replace the brittle string match with something that covers both API paths. For example:

```go
err = contextAPI.CreateContext("orb-publishing")
if err != nil {
    errMsg := strings.ToLower(err.Error())
    if strings.Contains(errMsg, "already exists") {
        fmt.Println("`orb-publishing` context already exists, continuing on")
    } else {
        return err
    }
}
```

This would match all known API error messages for a duplicate context:
- REST: `"A context already exists with this name"`
- GQL: `"Error creating context: ALREADY_EXISTS"`

### Workaround

Users can select **"No, I'll do this later."** at the publishing context prompt when the `orb-publishing` context already exists in their org. They may then need to manually ensure the `CIRCLE_TOKEN` environment variable is set in the existing `orb-publishing` context.

### Environment

- **CLI used with:** CircleCI Cloud (circleci.com)
- **API path:** REST (`POST /context`)

nanophate:
## Summary

- Fixes a string mismatch in `initOrb` that causes `circleci orb init` to fail with `"Error: A context already exists with this name"` when the `orb-publishing` context already exists in the organization (e.g., created by a teammate).
- The existing check looked for `"A context named orb-publishing already exists"`, which doesn't match what the REST API (`"A context already exists with this name"`) or GQL API (`"Error creating context: ALREADY_EXISTS"`) actually returns.
- Replaced with a case-insensitive check for `"already exists"` so the CLI gracefully continues regardless of API path.

Fixes #1192

## Test plan

- [x] Run `circleci orb init` in an org where `orb-publishing` context already exists, select "Yes, set up a publishing context with my API key" — should print `` `orb-publishing` context already exists, continuing on `` and proceed
- [x] Run `circleci orb init` in an org with no existing `orb-publishing` context — should create the context normally
- [x] Run `circleci orb init` and select "No, I'll do this later" — should skip context creation as before

### DIFF
--- a/cmd/orb.go
+++ b/cmd/orb.go
@@ -1405,7 +1405,7 @@ func initOrb(opts orbOptions) error {
 		contextAPI := context.NewContextClient(opts.cfg, "", vcsProvider, ownerName)
 		err = contextAPI.CreateContext("orb-publishing")
 		if err != nil {
-			if strings.Contains(err.Error(), "A context named orb-publishing already exists") {
+			if strings.Contains(strings.ToLower(err.Error()), "already exists") {
 				fmt.Println("`orb-publishing` context already exists, continuing on")
 			} else {
 				return err


cmd/orb.go
```diff
@@ -1405,7 +1405,7 @@ func initOrb(opts orbOptions) error {
 		contextAPI := context.NewContextClient(opts.cfg, "", vcsProvider, ownerName)
 		err = contextAPI.CreateContext("orb-publishing")
 		if err != nil {
-			if strings.Contains(err.Error(), "A context named orb-publishing already exists") {
+			if strings.Contains(strings.ToLower(err.Error()), "already exists") {
```
anthony-j-castro, line 1408: I think you might need another case here in order to match the error returned by the graphql endpoint? The example you provided has an underscore in that string: `"Error creating context: ALREADY_EXISTS"`

pete-woods:
Closing: targets v0's orb init. The v1 CLI no longer string-matches the API error — it falls back to ListContexts to reuse an existing orb-publishing context, which handles this case robustly across REST/GQL wording.

PR #1200, opened 2026-04-02
Use go.mod tool directive, upgrade golangci-lint v2

danmux:
## Summary
- Add `tool` directive to `go.mod` for golangci-lint v2 and gotestsum
- Migrate `.golangci.yml` to v2 format
- Update Taskfile, Makefile and CI to use `go tool` invocations
- Remove dedicated `golangci/golangci-lint:v1.64.8-alpine` Docker image from CI lint job, use standard `go` executor instead


pete-woods (approved):

Makefile
meeech, line 1: should the makefile jsut be aliases to the go tasks? seems silly to have to update both? can you see any reason for it?

meeech (approved):

Makefile
danmux: Makes sense, (will keep for a a follow up)

PR #1203, opened 2026-04-20
[DR-342] feat: add `circleci deploy init` command

avalcepina:
## Changes

- Add a new `cmd/deploy` package exposing a top-level `deploy` command group with an `init` subcommand.
- `deploy init` reads `.circleci/config.yml`, detects jobs whose names contain deploy-related keywords (`deploy`, `release`, `publish`, `ship`), prompts for a component name (defaulting to the inferred repo name) and an environment name (inferred from the job name when possible), and appends a `circleci run release log` step to each detected job.
- Register the new command group in `cmd/root.go`.
- Bump the expected top-level subcommand count in `cmd/root_test.go` from 29 to 30.
- Add unit tests and YAML fixtures covering detection, patching, idempotency, the no-deploy-jobs path, and the full interactive flow (via an injected `UserInputReader`).

## Rationale

Setting up deploy markers today requires reading docs, understanding the `circleci run release` CLI syntax, and hand-editing `config.yml`. This PR ships the zero-to-one experience described in [DR-342](https://linear.app/circleci/issue/DR-342/circleci-deploy-init-zero-to-one-deploy-marker-setup-command): an engineer runs one command from their repo, answers at most two questions per detected job, and their config is wired up for deploy markers.

Design principles (from the ticket):

- Two questions maximum per job; infer when possible.
- Never fails the engineer's pipeline — missing config, no deploy jobs, or an already-instrumented config all exit cleanly.
- Idempotent — re-running the command leaves an already-instrumented config untouched.
- Works offline — no API calls during setup, just local file parsing.
- Never commits or pushes on the user's behalf. The command prints a `git diff` / `git commit` / `git push` checklist and stops.

### Dashboard URL

The output prints the base dashboard URL `https://app.circleci.com/deploys`. Building an org-scoped URL reliably would require either a network round-trip to `/me/collaborations` or parsing `CIRCLE_BUILD_URL` (not available on a developer's laptop). Neither fits the "works offline" design principle, so the base URL is printed and WUC redirects the user to their most recently viewed org on arrival.

### YAML round-trip

Patching uses `gopkg.in/yaml.v3`'s node API, which preserves comments and relative ordering. It does not preserve blank lines between top-level keys — a known `yaml.v3` limitation. In practice this means a reformatted but semantically identical config; the step we add is always well-formed.

### Job detection heuristics

The keyword list (`deploy`, `release`, `publish`, `ship`) is intentionally conservative. Jobs with non-obvious names (e.g. `production-push`) won't be detected; the no-deploy-jobs branch prints a ready-to-paste snippet so users can wire those up by hand.

### Prompts

Follows the same `UserInputReader` pattern used by `cmd/pipeline`, so tests can run the whole flow without a TTY. The public `CustomReader` option lets future integration tests (or embedders) override the reader.

## Testing

- `go test ./cmd/deploy/...` — 13 new tests covering detection, patching (including idempotency), and the full init flow via scripted prompts.
- `go test ./cmd/...` — existing suites still pass; `cmd/root_test.go` updated for the new subcommand count.
- `go build ./...` and `go vet ./...` clean.
- Manual smoke tests against all three branches (no deploy jobs / already instrumented / happy path).

## Related

- Linear: [DR-342](https://linear.app/circleci/issue/DR-342/circleci-deploy-init-zero-to-one-deploy-marker-setup-command)
- Complements DR-341 (rich terminal output from `run release` commands in `task-agent-subcommand-release`), which prints a matching dashboard link from inside CI jobs.

## Checklist

- [x] Self-review of the diff
- [x] Comments on non-obvious intent (YAML node construction, idempotency, URL trade-off)
- [x] Checked for similar issues / prior work
- [x] Not a security issue
- [x] Read the Contributing guidelines

### DIFF
--- a/cmd/deploy/init.go
+++ b/cmd/deploy/init.go
@@ -1,0 +1,221 @@
+package deploy
+
+import (
+	"fmt"
+	"io"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/CircleCI-Public/circleci-cli/config"
+	"github.com/CircleCI-Public/circleci-cli/git"
+	"github.com/CircleCI-Public/circleci-cli/settings"
+)
+
+// deploysDashboardURL is the base URL printed at the end of a
+// successful init. We intentionally do not try to build an org-scoped
+// URL here: doing so reliably across OAuth orgs (gh/bb slug) and
+// standalone CircleCI orgs (opaque CIAM id) requires an authenticated
+// API call, which would break the "works offline" design goal. The
+// dashboard landing page redirects users to their most recent org.
+const deploysDashboardURL = "https://app.circleci.com/deploys"
+
+type initOptions struct {
+	// configPath points to the .circleci/config.yml the command reads
+	// and writes back. Defaults to config.DefaultConfigPath; exposed as
+	// a hidden flag so tests can redirect it at a temp file.
+	configPath string
+}
+
+func newInitCommand(_ *settings.Config, deployOpts *deployOpts) *cobra.Command {
+	iopts := initOptions{
+		configPath: config.DefaultConfigPath,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "Wire deploy markers into your .circleci/config.yml",
+		Long: `Scan the project's .circleci/config.yml, detect jobs that look like
+deployments, and append a step that records a deploy marker on every run.
+
+The command is idempotent: re-running it against an already instrumented
+config leaves it untouched. It never commits or pushes any changes —
+after it finishes you are prompted to review the diff and commit
+yourself.`,
+		Example: `  # Run from the root of your repository:
+  circleci deploy init`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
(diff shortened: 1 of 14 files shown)


cmd/deploy/init.go
```diff
@@ -0,0 +1,234 @@
...
+func allInstrumented(jobs []DetectedJob) bool {
+	for _, j := range jobs {
+		if !j.AlreadyInstrumented {
+			return false
+		}
+	}
+	return len(jobs) > 0
+}
+
+func joinJobs(names []string) string {
```
quinnallman: why reimplement this instead of using strings.join?

avalcepina: ah sorry, I meant to replace it but I forgot. good catch.

avalcepina: Fixed.

Chewypewy:
check out not knowing too much what it can do

quinnallman (approved):

avalcepina:
> check out not knowing too much what it can do

I can provide some context. 
It just checks your config for jobs naming a pattern that could imply they are deploy jobs and adds the simplest form of deploy marker (simple logs) to them. It double checks most of the values with the user and the gives some instruction on how to check the diff, commit, look it up in the ui or check the docs.
That's basically all.

PR #1209, opened 2026-04-22
Some standardization

pete-woods:
- Use `golangci-lint` for formatting (one less tool, and it does a better job).
- Run `ci:diff` task in linter (to make sure formatting, generated code, go.mod and so on are up to date).
- Extract some common things in the CI config.
- Store the linter cache.


gcirclest (approved):

.circleci/config.yml
```diff
@@ -3,38 +3,43 @@ version: 2.1
...
-      - run: task mod-download
+      - setup
       - go/with-cache:
+          golangci-lint: true
           steps:
             - run: task ci:lint
-            - store_test_results:
-                path: ./test-reports
+            - store_results
+            - run: task ci:diff
```
pete-woods, line 33: Make sure we check for diff-drift.

PR #1219, opened 2026-04-23
Use system keyring to store token

pete-woods:
- This library doesn't actually need CGO, so simplifies things a bunch.

### DIFF
--- a/internal/keyring/keyring.go
+++ b/internal/keyring/keyring.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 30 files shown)


internal/keyring/keyring.go
```diff
@@ -0,0 +1,99 @@
...
+			return "", ErrNotFound
+		}
+		return res.password, res.err
+	case <-ctx.Done():
+		return "", ctx.Err()
+	}
+}
+
+func hostname(service string) string {
+	return "cci:" + service
```
calvis: ~ i dont know best practices here, should this be the full `circleci:` binary name or something `com.circleci`?

calvis (approved):

internal/keyring/keyring.go
pete-woods: GitHub uses `gh:` as their prefix? Not really sure

pete-woods: Can't hurt to use a fqdn, though

PR #1230, opened 2026-04-26
Improve watch refresh

z00b:
In non-TTY mode, report more regularly, even if there is no new information.

In TTY mode, separate the ticks on "elapsed" time from the actual API polling, so we can back off polling but refresh regularly.

Also fix add pagination because claude guessed that was the problem. It wasn't but it would be for someone!


pete-woods (approved):

pete-woods:
Closing for now, as things have drifted a lot

PR #1236, opened 2026-04-27
Proposal: Remove section on "delight"

gordonsyme:
The other parts of this CLI guide already cover the things that make a
CLI pleasant to use:
* stable output that's easily piped to Big Data Tools like awk/sed/cut
  etc.
* good error messages
* consistency in switches
* consistent exit codes
* easily accessible documentation
* thoughtful use of colour in interactive contexts

The examples given in the removed "Delight" section are examples of what
not to do.
* Don't use the CLI as a billboard for marketing/advertising.
* Don't use wit in error messages. Something has gone wrong, the user
  has to work out what/why, and whimsical error messages diminish
  that fact.

### DIFF
--- a/agents/04-output.md
+++ b/agents/04-output.md
@@ -190,23 +190,6 @@ As a rule of thumb: if a user would pipe the output, it's an output command. If 
 
 ---
 
-## Delight
-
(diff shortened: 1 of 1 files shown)


calvis (approved):
~ agree that these examples dont make sense for this context; wonder if we can recapture intent somewhere else? although it looks like delight is mentioned in the design philosophy already

PR #1245, opened 2026-04-28
Add `--jq` switch to support processing JSON output

pete-woods:
- With this you don't need `jq` installed where you're executing the CLI.
- You also can ask for exactly the data you want, if you want to pass to `sed`, `awk`, etc.

##
Screenshot 2026-04-28 at 17 52 54


internal/jq/testdata/empty_array_colorized.txt
```diff
@@ -0,0 +1 @@
+[1;37m[[m[1;37m][m
```
calvis, line 1: filenames are a lil gross here

calvis (approved):

internal/jq/testdata/empty_array_colorized.txt
pete-woods: Yeah, I wanna do a bit more on this internal package. Will clean those up a bit :)

PR #1261, opened 2026-04-30
Fix IntelliJ warnings

pete-woods:


acceptance/acceptance_test.go
```diff
@@ -30,17 +30,17 @@ import (
 	"github.com/CircleCI-Public/circleci-cli-v2/internal/testing/binary"
 )
 
-const testToken = "testtoken"
+const testToken = "test-token"
```
calvis, line 33: just curious why this is an intellij warning

calvis (approved):

acceptance/acceptance_test.go
pete-woods: it thought it was a spelling mistake, and I couldn't be bothered arguing

PR #1266, opened 2026-05-02
fix: prevent terminal garbage after spinner (Ghostty/Kitty keyboard responses)

meeech:
[regened chunk stuff too]

## What

Fixes garbage characters (`1u`) appearing on the shell prompt after commands that use the spinner (e.g. `circleci pipeline get`) in Ghostty and other modern terminals.

## Why it happened

The spinner runs as a bubbletea program with `tea.WithInput(nil)` — stdin is kept out of raw mode so Ctrl+C still generates SIGINT. On the first render frame, bubbletea unconditionally sends a Kitty keyboard query (`\x1b[?u`) to probe terminal support. Ghostty responds with `\x1b[?1u`. Because bubbletea never reads stdin (input is disabled), the response sits there until the shell reads it — the `\x1b[?` prefix is consumed by readline and `1u` appears as typed text.

There were also separate DECRQM queries (modes 2026/2027) triggered by bubbletea's `shouldQuerySynchronizedOutput` based on `TERM`/`TERM_PROGRAM`, which produced additional garbage like `2026;2$y`.

## Fix

Two layers:

1. **`noQueryEnviron()`** — passed to bubbletea via `tea.WithEnvironment`. Injects `SSH_TTY=1` (suppresses the TERM_PROGRAM-based mode 2026/2027 path) and normalises `TERM` for Ghostty, WezTerm, Alacritty, Kitty, and Rio to `xterm-256color` (suppresses the TERM-based path). This eliminates the DECRQM garbage entirely.

2. **`drainStdinBuffer()`** — called in `Stop()` after the bubbletea program exits. Switches stdin to raw mode (clearing ICANON), waits up to 100ms via `select(2)` for the Kitty keyboard response, then non-blocking reads everything out. The raw mode switch is critical: in canonical mode `select` only returns when a complete line (`\n`-terminated) is buffered — the `\x1b[?1u` response has no newline, so without this the drain always timed out silently. Unix-only; Windows uses a no-op stub.


meeech:
Closing in favour of #1264 which uses TIOCFLUSH/TCFLSH to discard the TTY input buffer directly — a cleaner approach than the MakeRaw+select drain used here.

PR #1268, opened 2026-05-02
chore: regen chunk context

meeech:
Regenerates the `.chunk/context/` files from the current review prompt.


meeech:
Closing. It looks like chunk automagically uses main, which isn't what we want.

PR #1269, opened 2026-05-04
FACT-139: add search command

schurchleycci:
## Summary

- Adds `circleci pipeline search` wrapping `POST /api/v2/pipeline/search`. Takes a raw filter expression via `--filter` plus `--after`/`--before` for date scoping. For simple branch filtering use `circleci pipeline list`; this command is the door to the filter-expression API.
- Adds `apiclient.FlexTime` so JSON decoding tolerates the empty timestamp strings the search endpoint returns on some pipelines.
- Extends the fake `CircleCI` server with a `/pipeline/search` handler that records the most recent request body so tests can assert the filter passes through unchanged.

## Notes

- `--project` selects the project (slug like `gh/org/repo`) and defaults to the current git remote, matching the rest of the `pipeline` group.
- `actor.id` requires a UUID, not a login. Called out in `Long`.
- If `--after`/`--before` are omitted, the API applies its own ~2-week default window. Documented in `Long`.

## Test plan

- [x] `task build` succeeds
- [x] `task lint` clean (only the pre-existing `gosec` finding in `internal/iostream/drain_bsd.go` remains, unrelated to this PR)
- [x] Full acceptance + unit suites pass (`go test -count=1 ./...`)
- [x] Live smoke against `gh/CircleCI-Public/circleci-cli`: default search, `--filter 'pipeline.git.branch == "next"'`, `pipeline.git.branch starts-with "feat/"`, `--after`, `--json --jq`
- [x] New acceptance coverage:
  - `TestPipelineSearch` — golden table output
  - `TestPipelineSearch_JSON` — golden JSON output
  - `TestPipelineSearch_WithFilter` — raw filter passes through unchanged
  - `TestPipelineSearch_EmptyTimestamp` — verifies `FlexTime` handles the empty-string `created_at`/`updated_at` the API returns
  - `TestPipelineSearch_EmptyResults` — empty result path

### DIFF
--- a/internal/cmd/pipeline/search.go
+++ b/internal/cmd/pipeline/search.go
@@ -93,6 +93,12 @@ func newSearchCmd() *cobra.Command {
 		`),
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if limit < 1 {
+				return clierrors.New("search.invalid_limit", "Invalid limit",
+					fmt.Sprintf("--limit must be a positive integer, got %d", limit),
+				).WithExitCode(clierrors.ExitBadArguments)
+			}
+
 			var afterTime, beforeTime *time.Time
 			if after != "" {
 				t, err := time.Parse(time.RFC3339, after)
--- a/internal/cmd/pipeline/search.go
+++ b/internal/cmd/pipeline/search.go
@@ -1,0 +1,269 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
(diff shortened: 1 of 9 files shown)


internal/cmd/pipeline/search.go
```diff
@@ -0,0 +1,275 @@
...
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/spf13/cobra"
+
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/apiclient"
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/cmdutil"
+	clierrors "github.com/CircleCI-Public/circleci-cli-v2/internal/errors"
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/gitremote"
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/iostream"
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/mdtable"
```
z00b, line 41: these references shouldn't be to `v2`

z00b (commented):
@schurchleycci it looks like this PR has drifted a bit, but the intent seems fine. are you still trying to land it?

internal/cmd/pipeline/search.go
schurchleycci: Ah no Dan was saying a while back that the plan was to have the search API power the list command - just forgot to close this

PR #1270, opened 2026-05-04
Add `deploy list` command

gcirclest:
- Lists the 10 most recent releases for a project via `GET /api/v2/deploy/projects/{project_id}/releases` (`page-size=10`, hardcoded for now).
- Auto-detects project from git remote, or pass `--project gh/org/repo`.
- Renders markdown table in TTY, supports `--json` and `--jq`.
- Acceptance tests for default output, JSON, empty, and no-token.


acceptance/testdata/TestDeployList.txt
```diff
@@ -0,0 +1,6 @@
+# Releases
```
calvis: imo.. `deploy list` shouldnt give a table of `Releases`. or is this just an inconsistency we're committed to

pete-woods (approved):

PR #1273, opened 2026-05-05
[FACT-139] More publishing from Go releaser

pete-woods:
Go releaser supports all of this now, and quill supports macOS signing and notarizing on any platform, so we don't need a large matrix of publishing jobs.

- Choco (via Docker image wrapper)
- Snap
- rpm/deb
- Commented out macOS signing/notarizing (need to dig out our developer account creds).


.goreleaser.yml
```diff
@@ -37,6 +37,18 @@ builds:
       - arm64
     ldflags:
       - -s -w -X main.version={{.Version}}
+    hooks:
```
pete-woods: @emilywoods here

calvis (approved):

Taskfile.yml
```diff
@@ -173,8 +173,7 @@ tasks:
   ci:release:
     desc: Tag, build, and publish a release
     vars:
-      VERSION: 'v1.0.{{env "CIRCLE_BUILD_NUM"}}+{{.COMMIT}}{{.DIRTY}}.preview'
-
+      VERSION: 'v1.0.{{env "CIRCLE_BUILD_NUM" | default "0"}}-pre'
```
pete-woods, line 176: goreleaser already included commit info in the version

PR #1274, opened 2026-05-05
feat(pipeline watch): suggest log commands when pipeline fails

gcirclest:
## Summary
- When a watched pipeline ends in failure, attach actionable suggestions to the existing structured error so the user knows exactly which command to run next.
- Per failed job (capped at 3): \`circleci logs <job-number>\`. If more than 3 jobs failed, point at \`circleci pipeline get\` for the full list. Always include \`circleci logs --last-failed\` as a generic fallback.
- Suggestions are emitted via \`CLIError.WithSuggestions\`, so they render under the existing \`Suggestions:\` block on stderr and are included in the JSON error payload when \`--json\` is active.

## Test plan
- [x] Updated \`TestPipelineWatch_Failed\` to assert the new suggestion text appears on stderr.
- [x] \`go test ./acceptance/ -run TestPipelineWatch -count=1\` passes.
- [x] \`go test ./internal/cmd/pipeline/ -count=1\` passes.
- [x] \`task lint\` reports 0 issues.

### DIFF
--- a/acceptance/watch_test.go
+++ b/acceptance/watch_test.go
@@ -137,6 +137,42 @@ func TestPipelineWatch_Failed(t *testing.T) {
 	assert.Check(t, cmp.Contains(result.Stderr, "circleci logs --last-failed"), "stderr: %s", result.Stderr)
 }
 
+// --- failed pipeline with a failed job → suggests `circleci logs <num>` ---
+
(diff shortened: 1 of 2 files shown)


pete-woods (commented):
Need a test for that output

pete-woods (approved):

PR #1277, opened 2026-05-05
feat(extension): implicit plugin mechanism via circleci-<name> PATH lookup

michael-webster:
## Summary

- Adds implicit extension support: any `circleci-<name>` binary in PATH is transparently invoked as `circleci <name>`, following git's plugin convention
- New `internal/extension` package handles PATH scan, environment injection, and exit code propagation
- Extensions appear as real Cobra commands — they show up in `circleci --help` under an "Extensions" group and participate normally in cobra routing
- Root-level flags (`--config`, `--debug`, etc.) are parsed for CLI use; only the extension's own args are forwarded to the subprocess

## How it works

At startup, `root.NewRootCmd` calls `extension.FindAll()` which scans PATH for executables named `circleci-*`. Each discovered extension is registered as a real Cobra command with `DisableFlagParsing: true` so cobra routes to it directly without touching the extension's flags.

When the command runs, the extension's `RunE` handler:
1. Splits `os.Args` at the command name to separate root flags from extension args — root persistent flags (`--config`, `--insecure-storage`, etc.) are parsed so auth config is available, while only the extension's own args are forwarded
2. Resolves `CIRCLECI_TOKEN`, `CIRCLECI_HOST`, and best-effort git-remote project metadata, overlaying them onto the current environment
3. Execs the `circleci-<name>` binary found via `exec.LookPath` with the extension's args and the augmented environment

Built-in commands always win on name conflicts — extensions cannot shadow them. Extensions appear under an "Extensions" group in `circleci --help`.

## Test plan

- [x] `task test` — all 464 tests pass
- [x] `task check` — lint clean (pre-existing license failure in `cmd/.circleci/update_check.yml` unrelated to this change)
- [x] Manually tested with a fake `circleci-hello` extension: args, `CIRCLECI_TOKEN`, and `CIRCLECI_HOST` all injected correctly
- [x] Manually tested `circleci testsuite "ci tests"` — dispatches to `circleci-testsuite` correctly, exit code propagated
- [x] Unknown commands with no matching extension still show the original cobra error
- [x] Unknown subcommands within a group (e.g. `circleci pipeline foo`) are not intercepted

Example running this:
image

### DIFF
--- a/acceptance/extension_test.go
+++ b/acceptance/extension_test.go
@@ -1,0 +1,155 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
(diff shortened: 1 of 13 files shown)


internal/extension/extension.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/gitremote"
+)
+
+// FindAll scans PATH for executables named "circleci-<name>" and returns the
+// extension names (the part after "circleci-"). The first entry in PATH wins
+// for duplicate names, matching exec.LookPath semantics.
+func FindAll() []string {
+	seen := map[string]bool{}
+	var names []string
+	for _, dir := range filepath.SplitList(os.Getenv("PATH")) {
```
pete-woods: Please no env var access in here

acceptance/extension_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/CircleCI-Public/circleci-cli-v2/internal/testing/binary"
+	testenv "github.com/CircleCI-Public/circleci-cli-v2/internal/testing/env"
+)
+
+// skipUnlessShell skips the test on platforms where shell scripts cannot be
+// executed as extension binaries (i.e. Windows).
+func skipUnlessShell(t *testing.T) {
+	t.Helper()
+	skip.If(t, runtime.GOOS == "windows", "shell-script extensions are not supported on Windows")
```
pete-woods: In here, I'd like to see us build a small go program and put it into the path. Then to make sure it works on Windows.

pete-woods (approved):

PR #1283, opened 2026-05-08
Add --failfast to 'circleci run watch'

gcirclest:
## Summary
- Adds \`--failfast\` to \`circleci run watch\`. When set, the command exits as soon as any job in the run is observed to have failed, rather than waiting for the remaining workflows to reach a terminal state.
- Exit code is 1 (\`ExitGeneralError\`) — same as the existing run-failed path. The error includes \`circleci logs <num>\` suggestions for each failing job.

## Test plan
- [x] \`task test\` — full suite (489 tests) passes
- [x] \`task check\` — lint, license, mod-tidy clean
- [x] New \`TestRunWatch_FailFast\` exercises the case where the workflow is still running but contains a failed job, and asserts: exit code 1, "failing job" in stderr, and a per-job log suggestion (\`circleci logs 200\`)
- [x] Updated golden help text in \`testdata/usage/circleci/run/watch.txt\`

### DIFF
--- a/acceptance/watch_test.go
+++ b/acceptance/watch_test.go
@@ -238,6 +238,46 @@ func TestRunWatch_SHA_NotFound(t *testing.T) {
 	assert.Check(t, cmp.Contains(result.Stderr, "No run found"), "stderr: %s", result.Stderr)
 }
 
+// --- --failfast: exit immediately when a job fails, without waiting for the rest of the run ---
+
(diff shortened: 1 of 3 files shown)


meeech (approved):
gabe needs this for demo, so approving. we can revert if its an issue for people (though seems useful)

PR #1286, opened 2026-05-11
Implement namespace management

z00b:
Note that this uses the gRPC APIs, since that's what's in current version and we don't appear to have another option just yet. This should clearly be migrated to an appropriate v3 API when one is available.


pete-woods (approved):
I'm gonna stamp this for the commands, and come back later to strip out the graphql.

PR #1301, opened 2026-05-13
Add telemetry enable/disable commands

schurchleycci:
Adds `circleci settings telemetry enable` and `circleci settings telemetry disable`. Preference is persisted to the config file; env vars (`CIRCLECI_NO_TELEMETRY`, `NO_ANALYTICS`, `DO_NOT_TRACK`, `CI`) always win. `telemetry_enabled` is also surfaced in `circleci settings list`.

## Design notes

- Telemetry commands live under `circleci settings telemetry` rather than as a top-level command group — telemetry is a CLI setting, so it belongs alongside `set` and `list`
- No `--json` on enable/disable — pure mutation commands; read path is covered by `settings list --json`
- Confirmation message uses the resolved config path (respects `--config`), not the XDG default

Screenshot 2026-05-14 at 11 04 04 AM

## Test plan

- [x] `circleci settings telemetry enable` / `disable` write the correct value and print the right path
- [x] `circleci --config /tmp/custom.yml settings telemetry enable` confirmation shows `/tmp/custom.yml`
- [x] `CIRCLECI_NO_TELEMETRY=1 circleci settings telemetry enable` saves the preference but prints the override notice
- [x] `circleci settings list` and `--json` both include `telemetry_enabled`
- [x] `task test`

### DIFF
--- a/internal/cmd/settings/telemetry_enable.go
+++ b/internal/cmd/settings/telemetry_enable.go
@@ -20,7 +20,7 @@
 //
 // SPDX-License-Identifier: MIT
 
-package telemetry
+package settings
 
 import (
 	"context"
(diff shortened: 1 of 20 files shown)


pete-woods:
This looks good, but I believe we should manage the setting under the `settings` command, like we do other settings.

schurchleycci:
> This looks good, but I believe we should manage the setting under the `settings` command, like we do other settings.

Good call - so pretty much take what I've got here and put it under e.g. `circleci settings telemetry enable` instead of a top level command?

schurchleycci:
@pete-woods this is updated to be `circleci settings telemetry` now

internal/cmd/settings/telemetry_enable.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
```
pete-woods, line 1: These files still look like they need moving.

Stamping, assuming that will be done, though.

pete-woods (approved):

pete-woods:
Have stamped under the assumption of moving those files.

schurchleycci:
> Have stamped under the assumption of moving those files.

Ahhh sorry about that, I've clearly been juggling too many Claude sessions. They're fully moved over now rather than duplicated.

PR #1302, opened 2026-05-13
Implement config validate/process/pack

z00b:
Adds `circleci config validate`, `config process`, and `config pack`.

Deliberate scope decisions:
- v2 REST only (`/compile-config-with-defaults`); v1 GraphQL fallback dropped
- `--org-slug`/`--org-id` carried forward for private orb resolution
- `--json` on validate only; pack/process output YAML as primary data
- Deprecated-image check omitted; API surfaces these natively
- `config generate` deferred (external dependency); `config migrate` dropped (legacy 2.0 migration command)

Also fixes a flag-shadowing bug: root's `--config/-c` (CLI settings path) and `validate`'s local `--config/-c` (pipeline YAML path) have the same name. `PersistentPreRunE` now reads from `cmd.Root().PersistentFlags()` to avoid picking up the child command's local value.

### DIFF
--- a/acceptance/config_test.go
+++ b/acceptance/config_test.go
@@ -23,13 +23,34 @@
 package acceptance_test
 
 import (
+	"encoding/json"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
 	"gotest.tools/v3/golden"
 
 	"github.com/CircleCI-Public/circleci-cli/internal/testing/binary"
 	testenv "github.com/CircleCI-Public/circleci-cli/internal/testing/env"
+	"github.com/CircleCI-Public/circleci-cli/internal/testing/fakes"
+)
+
+const (
+	// language=yaml
+	testConfigYAML = `version: "2.1"
+jobs:
+  build:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - checkout
+`
+	// language=yaml
+	testCompiledYAML = `# compiled output
+version: "2.1"
+`
 )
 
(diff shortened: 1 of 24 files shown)


internal/apiclient/config.go
```diff
@@ -0,0 +1,124 @@
...
+// Collaboration is an org the authenticated user belongs to.
+type Collaboration struct {
+	VCSType string `json:"vcs_type"`
+	Slug    string `json:"slug"`
+	Name    string `json:"name"`
+	ID      string `json:"id"`
+}
+
+// GetCollaborations returns all orgs the authenticated user is a member of.
+func (c *Client) GetCollaborations(ctx context.Context) ([]Collaboration, error) {
```
pete-woods: Pretty surprised it needs this! :D

acceptance/config_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+	"gotest.tools/v3/golden"
+
+	"github.com/CircleCI-Public/circleci-cli/internal/testing/binary"
+	testenv "github.com/CircleCI-Public/circleci-cli/internal/testing/env"
+	"github.com/CircleCI-Public/circleci-cli/internal/testing/fakes"
+)
+
+const testConfigYAML = "version: \"2.1\"\njobs:\n  build:\n    docker:\n      - image: cimg/base:stable\n    steps:\n      - checkout\n"
```
pete-woods: Would recommend a multi line string for readability

Also good to add `//language=YAML` comment to help IDEs.

z00b:
after an embarrassing number of lint and test failures, i believe i have addressed the open comments 😬

pete-woods (approved):

PR #1304, opened 2026-05-13
Add reposcan domain package

Klaney:
## Summary

- Adds `internal/reposcan/` domain package: `Result`, `SetupStep`, `Scanner` interface, `Render`, and `NewDefaultScanner` backed by `github.com/CircleCI-Public/chunk-cli/envbuilder`
- `Scanner.Scan(ctx, dir)` calls `envbuilder.DetectEnvironment` and maps the result to an insulating `reposcan.Result` type — keeps the public surface stable across env-builder upgrades
- `Render` writes human-readable detection output to stderr (stack, image, setup steps, or fallback message for empty detection)
- Full unit test coverage (offline); env-builder types don't appear in the package's public surface
- Adds `chunk-cli` dependency to `go.mod` / `go.sum`

No user-facing command is wired up in this PR — this is the reusable library that `circleci init` (and other commands) will call into.

## Test plan

- [ ] `task test` passes (unit tests in `internal/reposcan/`)
- [ ] `task check` passes (lint, license headers, mod tidy)
- [ ] `go test ./internal/reposcan/...` green with no network calls


Klaney:
## For temporary testing I exposed the functionality in a command

It has since been removed

Screenshot 2026-05-13 at 12 14 19 PM

internal/reposcan/scanner.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+
+	"github.com/CircleCI-Public/chunk-cli/envbuilder"
+)
+
+// Scanner detects the stack and setup commands for a repository at dir.
+// Implementations should treat dir as a read-only filesystem path.
+type Scanner interface {
```
pete-woods: Is there another scanner implementation? This interface should only exist if there is.

internal/reposcan/scanner.go
```diff
@@ -0,0 +1,68 @@
...
+
+// NewDefaultScanner returns a Scanner backed by chunk-cli's env-builder
+// library. The default scanner may make network calls (e.g. to Docker Hub)
+// while resolving image versions, so callers should pass a context with a
+// timeout if they need to bound the operation.
+func NewDefaultScanner() Scanner {
+	return &envbuilderScanner{}
+}
+
+type envbuilderScanner struct{}
```
pete-woods: This should be an exported struct

internal/reposcan/scanner.go
```diff
@@ -0,0 +1,68 @@
...
+// Implementations should treat dir as a read-only filesystem path.
+type Scanner interface {
+	Scan(ctx context.Context, dir string) (*Result, error)
+}
+
+// NewDefaultScanner returns a Scanner backed by chunk-cli's env-builder
+// library. The default scanner may make network calls (e.g. to Docker Hub)
+// while resolving image versions, so callers should pass a context with a
+// timeout if they need to bound the operation.
+func NewDefaultScanner() Scanner {
```
pete-woods: Constructors should return concrete types

internal/reposcan/scanner.go
Klaney: Good point. Got a little overzealous

pete-woods (approved):

PR #1310, opened 2026-05-14
Add auth signup using login fallback flow

Fab10-CircleCi:
## Summary

This PR adds `circleci auth signup` to the next CLI auth flow.

The command starts a signup OAuth flow, opens the signup URL in the browser, then asks the user to return to the terminal after verifying their email. When the user presses Enter, the CLI first checks whether the original signup OAuth callback has already completed. If it has, the CLI exchanges that code and saves the token. If not, it falls back to the existing `auth login` browser flow.

This supports the temporary product/backend constraint where signup may not always complete the CLI auth handoff through the loopback callback.

## Decision Context

- Terminals can render printed URLs as clickable links, but the CLI cannot detect whether the user clicked one.
- Because of that, the command opens the signup URL immediately instead of asking the user to press Enter just to open it.
- The signup URL still includes the CLI loopback callback, so if signup does redirect back successfully, the CLI consumes that callback and avoids a second login flow.
- Reused the existing login token persistence path for successful signup callback completion.
- Added a Bubble Tea continue prompt for the signup verification step.
- Removed the old `PromptLine` helper.
- Added acceptance coverage for:
  - non-interactive signup URL output
  - already-authenticated guard
  - Ctrl+C cancellation at the verification prompt
  - fallback to regular login when no signup callback is available
  - direct completion when the signup callback is available

## Testing

\`\`\`sh
go test ./internal/httpcl ./internal/oauth ./internal/cmd/cmdauth
go test ./acceptance -run 'TestAuthSignup' -count=1
go test ./acceptance -count=1
\`\`\`


acceptance/auth_signup_test.go
```diff
@@ -0,0 +1,347 @@
...
+
+	assert.Assert(t, t.Run("token persisted to YAML", func(t *testing.T) {
+		cfgPath := filepath.Join(env.HomeDir, ".config", "circleci", "config.yml")
+		body, err := os.ReadFile(cfgPath)
+		assert.NilError(t, err)
+		assert.Check(t, cmp.Contains(string(body), "test-signup-callback-token"))
+	}))
+}
+
+func runSignupAndInterrupt(t *testing.T, env *testenv.TestEnv) (int, string) {
```
pete-woods: Most of this function is duplicated from the binary package.

I'd also like to restrict construction of the console to that package.

internal/cmd/cmdauth/login.go
```diff
@@ -124,6 +129,9 @@ func runLogin(ctx context.Context, configPath string, noBrowser, secureStorage b
 	defer func() { _ = flow.Close() }()
 
 	iostream.ErrPrintf(ctx, "Open this URL in your browser to continue:\n\n  %s\n\n", flow.AuthorizeURL)
+	if openBrowser {
```
pete-woods: Isn't this the non interactive path?

internal/cmd/cmdauth/signup.go
```diff
@@ -0,0 +1,176 @@
...
+				"Run 'circleci auth logout' to clear it, then re-run 'circleci auth signup'",
+				"If you already have a CircleCI account, use 'circleci auth login'",
+			).
+			WithExitCode(clierrors.ExitAuthError)
+	}
+
+	host := cfg.EffectiveHost()
+	deviceID := config.EnsureDeviceID(ctx, configPath)
+
+	flow, err := oauth.StartSignup(ctx, host, deviceID, runtime.GOOS)
```
pete-woods, line 108: All this feels like it should be a single bubbletea program

acceptance/auth_signup_test.go
Fab10-CircleCi: Good point. I’ll remove the signup-specific `runSignupAndInterrupt` helper and the dedicated acceptance test rather than duplicating console construction outside `internal/testing/binary`.
Now that the prompt is Bubble Tea-based, I don’t think this PR needs a add interrupt signup test.

internal/cmd/cmdauth/login.go
Fab10-CircleCi: Yes, the naming/comment are now misleading. The normal auth login non-interactive path still passes openBrowser=false. Signup fallback passes openBrowser=true because it needs to launch the known-good browser login flow without running the full login TUI. I’ll rename or document this helper so it is clearly “browser OAuth without the full login TUI,” not strictly “non-interactive.”
      - runLoginNoBrowser(...)
      - runLoginBrowserFallback(...)
Or move the fallback into a Bubble Tea signup model so runLoginBrowser(..., openBrowser) is less awkward. 🤔

internal/cmd/cmdauth/signup.go
Fab10-CircleCi: I agree. The current implementation uses Bubble Tea only , which fixed Ctrl+C but leaves signup orchestration out. I can refactor interactive signup into SignupFlowModel., browser open, Enter-to-continue  state, signup callback check, and fallback-to-login decision.

pete-woods (approved):

PR #1312, opened 2026-05-15
Prevent auth acceptance tests from opening browser tabs

Fab10-CircleCi:
## Summary

Acceptance tests were launching real browser tabs because the interactive auth UI models called `browser.OpenURL` directly.

This adds an injectable browser opener to the login and signup flow models. In normal usage, the opener still defaults to `browser.OpenURL`, so production behavior is unchanged. In acceptance tests, `TestEnv` now sets `CIRCLECI_NO_BROWSER_OPEN=true`, which makes the command layer pass a no-op opener instead.

`CIRCLECI_NO_BROWSER_OPEN` is intentionally undocumented; it is a test seam, not a supported user-facing configuration flag.

## Context

`auth login` already opened a browser tab during the interactive browser-login path. PR #1310 added signup coverage that exercises browser auth more often, so running acceptance tests became noisy and disruptive locally.

Before this change, the login interactive path, signup start, and signup-to-login fallback could all open browser tabs during acceptance tests.

With this change:
- Normal `circleci auth login` and `circleci auth signup` still open the browser.
- Acceptance tests still exercise the full interactive login/signup flows.
- Acceptance tests no longer open real browser tabs.

## Testing

Unit tests cover the `browserOpener()` env-var switch and the signup model's use of the injected opener with and without `NoBrowser`.

\`\`\`sh
go test ./internal/ui ./internal/cmd/cmdauth
go test ./acceptance -run 'TestAuthSignup|TestAuthLogin' -count=1
go test ./acceptance -count=1
go test ./...
task ci:check
\`\`\`

### DIFF
--- a/internal/cmd/cmdauth/browser.go
+++ b/internal/cmd/cmdauth/browser.go
@@ -1,0 +1,38 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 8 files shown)


pete-woods:
Resolved the issues with signup on the backend

PR #1316, opened 2026-05-18
[WEBXP-990] Add circleci config generate

Klaney:
[Link to Ticket](https://linear.app/circleci/issue/WEBXP-990/config-generation-via-env-builder-step-3-of-circleci-init)

## Summary
- Adds `circleci config generate <path>`, which scans a repo and writes a starter `.circleci/config.yml`.
- New `internal/configgen` package exposes a `Generate` API; `internal/cmd/config` hosts the verb and group.
- Skips when an existing config is present; surfaces scan errors; falls back to a generic template for unknown stacks.
- Unit, golden, and acceptance test coverage for happy path, skip-existing, path-not-found, scan failure, and fallback template.

## Test plan
- [ ] `task test`
- [ ] `./dist/circleci config generate ./some-repo` against a Node project — verifies detection
- [ ] `./dist/circleci config generate ./empty-dir` — verifies fallback template
- [ ] Re-run on a repo that already has `.circleci/config.yml` — verifies skip

## Some random questions
Writing acceptance tests for this was difficult due to the "side effect" nature of the outputs being dependent on the users files/project etc. I introduced fixtures as a way to get halfway decent Acceptance tests, however if this pattern is something we don't like I'm very open to alternatives, fixtures just seemed like the best way forward to me. 

I created it under `configgen`, as it seemed different enough from the `config` domain that it warranted its own space (`config` is the circleci `config`, config gen is creating a circleci config, but it's not a fully fledged config in the system... but maybe I'm overthinking it)

## Random comment

After the onboard scaffold stuff is done, we should look into breaking the env builder into it's own shareable module so we don't keep coupling `cli` to `chunk-cli`

### testing different projects

Screenshot 2026-05-14 at 9 40 04 AM

Screenshot 2026-05-14 at 9 39 27 AM

### Test including path

Screenshot 2026-05-14 at 9 42 59 AM

### Test no directory

Screenshot 2026-05-14 at 9 44 16 AM


pete-woods:
For the acceptance tests, what I'd expect is a set of subdirectories in testdata somewhere that look like a skeleton node project or whatever is enough to get the scaner to work.

I don't see any tests that actually exercise the scanning capability.

Also, for golden tests, use regular file extensions like `.yml`, `.txt` like the other golden tests do.

Klaney:
> For the acceptance tests, what I'd expect is a set of subdirectories in testdata somewhere that look like a skeleton node project or whatever is enough to get the scaner to work.
> 
> I don't see any tests that actually exercise the scanning capability.
> 
> Also, for golden tests, use regular file extensions like `.yml`, `.txt` like the other golden tests do.

Yeah it's tough cause the project scanner reaches out to docker hub which makes it not deterministic because then dockerhub could fail or block requests or whatever. But the .NET one doesn't do that. So I'll write a mock .net proj and assert the differences.

I'll fix the golden txt files extensions.

Klaney:
I have a "node skeleton project" on deck for this as well, it works locally. the only reason I don't push it up is it will make the acceptance test flaky because scanner, for node projects, reaches out to dockerhub to resolve the latest pinned `cimg/node` to build the config. This means if dockerhub is down or we get rate limited the test will fail.

Maybe there's a workaround? The .net one is the most deterministic and reliable though, as it doesn't reach out to dockerhub.

acceptance/config_generate_test.go
```diff
@@ -0,0 +1,172 @@
...
+	env := testenv.New(t)
+	result := binary.RunCLI(t, binary.RunOpts{
+		Binary:  binaryPath,
+		Args:    []string{"config", "generate", dir},
+		Env:     env.Environ(),
+		WorkDir: t.TempDir(),
+	})
+
+	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
+	assert.Equal(t, result.Stdout, "")
```
pete-woods: It seems like we'd want some output here, to tell the user we generated some config?

internal/cmd/config/generate_test.go
```diff
@@ -0,0 +1,153 @@
...
+	assert.Check(t, cmp.Equal(generate.Name(), "generate"))
+}
+
+// runGenerate invokes 'circleci config generate' in-process against the given
+// project directory and returns the captured stderr (with dir replaced by the
+// placeholder "<DIR>" and any Windows path separators normalized to "/" so
+// golden files are stable across platforms) plus the error from RunE. Cobra's
+// default error rendering is silenced because the production root command
+// does the same — errors are formatted by main.go.
+func runGenerate(t *testing.T, dir string, extraArgs ...string) (stderr string, err error) {
```
pete-woods: These tests seem to duplicate the sort of thing we are doing in the acceptance tests.

pete-woods:
Those changes are a good step in teh right direction. Can you now move all the tests inline with the commands to fully fledged acceptance tests that don't need to interfere with the internal state of the CLI.

Let's avoid testing the same thing twice in different ways, and pull e.g. all the error handling tests out to acceptance tests (looks like there's a bunch of those already)

acceptance/config_generate_test.go
Klaney: Yeah, we do we say "Generated .circleci/config.yml" I'll add it to the test

internal/cmd/config/generate_test.go
Klaney: yeah thats a good point. i got a little overzealous testing everywhere

Klaney:
The golden looks weird but that's indeed the output. Maybe an odd dotnet convention

Screenshot 2026-05-20 at 10 28 21 AM

internal/cmd/config/generate_test.go
Klaney: https://github.com/CircleCI-Public/circleci-cli/pull/1316/changes/46740397e7bdea778bf9b20bbe15b29461e7b050

internal/reposcan/render_test.go
```diff
@@ -109,5 +109,5 @@ func TestRender_SystemSetupStep_IsRendered(t *testing.T) {
 		},
 	})
 
-	assert.Check(t, golden.String(errBuf.String(), t.Name()+".txt"))
+	assert.Check(t, golden.String(outBuf.String(), t.Name()+".txt"))
```
Klaney, line 112: I noticed in my previous PR and in this one I was writing Errors instead of Printf which didn't seem right so i changed these

pete-woods (approved):

PR #1326, opened 2026-05-20
Make httpcl always return status, so debug can work properly

pete-woods:
Screenshot 2026-05-20 at 10 39 19

### DIFF
--- a/internal/httpcl/client.go
+++ b/internal/httpcl/client.go
@@ -81,6 +81,7 @@ func New(cfg Config) *Client {
 
 	rc := retryablehttp.NewClient()
 	rc.RetryMax = 3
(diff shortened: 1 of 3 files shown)


internal/httpcl/client.go
```diff
@@ -180,7 +185,11 @@ func (c *Client) Call(ctx context.Context, r Request) (status int, err error) {
 
 	resp, err := c.http.Do(req)
 	if err != nil {
-		return 0, err
+		submatch := statusRegex.FindStringSubmatch(err.Error())
```
danmux: resp is right there :D

pete-woods: (but nil)

danmux (approved):

PR #1332, opened 2026-05-20
Simplify auth signup to reuse login flow with signup=true [CIAM-152]

lavin-ds:
## Ticket
[CIAM-152](https://linear.app/circleci/issue/CIAM-152)

## Changes
- Signup now reuses the login flow (host selection, method selection, browser OAuth) with `signup=true` on the authorize URL
- Added callback timeout to the interactive TUI (was missing, would hang forever)


pete-woods (approved):

Chewypewy (commented):
No comment

PR #1342, opened 2026-05-25
Implement `circleci project dlc purge`

z00b:
Include `circleci dlc` alias group.

### DIFF
--- a/internal/cmd/project/follow.go
+++ b/internal/cmd/project/follow.go
@@ -25,7 +25,6 @@ package project
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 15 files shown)


pete-woods:
Instead of adding a new domain, can we do a 3 line PR to expose this internal API at least on the regular API domain. This will avoid us having to special case all this for server:

https://github.com/circleci/output/blob/main/deploy/values.yaml#L211

https://github.com/circleci/output/blob/main/api/public/api.go#L81

internal/cmd/project/follow.go
```diff
@@ -103,12 +102,3 @@ func runProjectFollow(ctx context.Context, client *apiclient.Client, projectSlug
 	iostream.Printf(ctx, "%s Now following %s\n", iostream.SymbolOK(ctx), projectSlug)
 	return nil
 }
-
-// parseSlug splits "gh/org/repo" → ("gh", "org", "repo").
-func parseSlug(slug string) (vcs, org, repo string, err error) {
```
pete-woods, line 108: 👍

pete-woods (approved):

pete-woods:
I think this is unblocked now with the V3 delete DLC API merged

PR #1346, opened 2026-05-26
WEBXP-989 Add local test runner via env-builder

Fab10-CircleCi:
## Summary
- adds `internal/testrunner` to run the env-builder detected test command from a `reposcan.Result`
- adds `circleci test run [path]` as a standalone command for manual verification before `circleci init` orchestration lands
- streams test output live, captures a bounded output tail, and prints an agent-ready prompt when tests fail
- adds distinct structured errors for no test command, test failure, runner error, scan failure, and path-not-found cases
- adds acceptance coverage and a regression guard that env-builder still emits a `test` setup step for .NET projects

## Notes
`chunk-cli/envbuilder` does not expose a local test-runner API, so this uses the detected setup step named `test` and trusts the subprocess exit code instead of parsing output.

`testrunner.Run(ctx, dir, result)` intentionally does not render the scan summary; callers own the preamble. For WEBXP-993, watch for `configgen.Generate` rendering the scan summary internally, which can double-print if the init orchestrator also renders Step 1.

## Test plan
- `GOCACHE=/private/tmp/circleci-go-cache go test ./internal/...`
- `GOCACHE=/private/tmp/circleci-go-cache go test ./acceptance -run '^TestTestRun'`

### DIFF
--- a/acceptance/test_run_test.go
+++ b/acceptance/test_run_test.go
@@ -1,0 +1,155 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+// SPDX-License-Identifier: MIT
+
+package acceptance_test
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/CircleCI-Public/circleci-cli/internal/reposcan"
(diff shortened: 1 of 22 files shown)


internal/testrunner/prompt.go
```diff
@@ -0,0 +1,50 @@
...
+Project stack: %s
+Image: %s
+Command run: %s
+Exit code: %d
+
+Failing test output:
+%s
+
+Please diagnose the failure from the output above and propose a fix.
+Once fixed, I will re-run ` + "`circleci init`" + `.
```
Klaney: I think for now we wouldn't want `circleci init` verbage right? maybe just change this to "I will rerun circleci test run"

internal/testrunner/testrunner.go
```diff
@@ -0,0 +1,141 @@
...
+			exitCode := exitErr.ExitCode()
+			outputTail := combineOutput(stdoutTail.String(), stderrTail.String())
+			iostream.Printf(ctx, "\n%s", RenderPrompt(result.Stack, image(result), testCommand, exitCode, outputTail))
+			return clierrors.New(
+				"test.tests_failed",
+				"Tests failed",
+				fmt.Sprintf("Test command exited with status %d.", exitCode),
+			).WithSuggestions(
+				"Paste the prompt above into your AI assistant",
+				"Re-run circleci init once tests pass",
```
Klaney: Same here, probably just "circleci test run". Maybe later we can configure it to use whatever the parent command was that ran this.

acceptance/test_run_test.go
Klaney, line 1: Can we also add a "No Arg" and "Pass in a file, not a directory" test to these?

Klaney (commented):
I think other than my comments I wasn't sure if these should be running in a docker container or not. But let's assume not not for now, we can add that later with a flag or something.

internal/testrunner/prompt.go
Klaney: I also know the ticket says to emit "circleci init" but I don't think that should be there imo.

Fab10-CircleCi:
@Klaney  Going with local-shell for this PR per your "let's assume not for now". In fact I wanted to put more eyes in this problem. We agreed we want to keep this ass simple as possible to ship something first then improve from there.

internal/testrunner/prompt.go
Fab10-CircleCi: agreed,  swapped circleci init for circleci test run in the template. We can revisit the wording once the orchestrator lands. Fixed in last push

internal/testrunner/testrunner.go
Fab10-CircleCi: Done. Swapped to circleci test run

acceptance/test_run_test.go
Fab10-CircleCi: Good catch. Thanks! added both tests.

Klaney (approved):
Tested locally, this does work. 

I'm approving mostly so as this can be a placeholder for a more robust testing apparatus in the future. Currently this just runs tests locally, not a big value add. 

In the future we will be modifying this command to actually run the tests in a `runner` to better emulate a users tests/config in a circleci environment. 

We should also talk to the execution team about how to make that happen referencing this thread in slack https://circleci.slack.com/archives/C0AUWSGBFQD/p1779817290560409

PR #1348, opened 2026-05-27
WEBXP-993 Add circleci onboard

Fab10-CircleCi:
[Linear ticket](https://linear.app/circleci/issue/WEBXP-993/orchestrate-the-4-steps-into-circleci-init-flow) — phase 1 only.

## Summary
- adds `circleci onboard [path]` as a thin signup-handoff orchestrator
- chains the existing local steps already on `next`: scan → tests → config gen → signup
- adds `internal/onboarder` (library) + `internal/cmd/onboard` (Cobra command)
- adds `cmdauth.SignupIfNeeded` — idempotent variant of signup for orchestrators; valid token → "Already signed in as X", stale token → `auth.signup.stale_token` clierror with logout suggestion. `circleci auth signup` is unchanged.
- refactors `configgen.Generate` to no longer call `reposcan.Render` internally; `circleci config generate` now renders explicitly with a `!result.IsEmpty()` guard mirroring `circleci test run`

## What's out of scope (deliberate)
Phase 1 ends after signup/auth. The following are explicit follow-ups, blocked on the SOC + CIAM API contract being confirmed:
- project creation, trigger creation, GitHub App install
- automated `git push` of the generated config
- mid-flow resume / state persistence

Success message reflects this:
> ✓ Onboarded
> Commit `.circleci/config.yml`. After your project is connected in CircleCI, pushing will start your first pipeline.

## Idempotency (per step)
Rerunning `circleci onboard` is safe:
- scan / render / tests: rerun every time (no persistent state)
- config gen: skip if `.circleci/config.yml` exists, print confirmation
- signup: skip if valid token exists, print "Already signed in as X"; stale token returns a structured error with a logout suggestion (never silently overwritten)

## Test plan
- `go test ./internal/cmd/cmdauth` — new `SignupIfNeeded` unit tests (already-authenticated path + stale-token path via `fakes.CircleCI`)
- `go test ./internal/cmd/root -run TestUsage` — usage goldens updated for `onboard`
- `go test ./acceptance -run '^(TestOnboard|TestConfigGenerate)'` — five onboard acceptance tests + config-generate empty-stack golden update
- `task test` and `task check`

Acceptance tests use a fake CircleCI server serving `/api/v2/me` for the already-authenticated branch. Fresh-signup OAuth callback is **not** covered by `TestOnboard_*` — it's exercised by the existing `acceptance/auth_signup_test.go` infrastructure plus manual verification.

## Notes for reviewers
- `internal/onboarder` imports `internal/cmd/cmdauth` for `SignupIfNeeded`. This is the only library→cmd import in the tree. Kept for v1 since there's exactly one consumer; if `SignupIfNeeded` gets more callers, extracting auth helpers into a non-cmd package is the natural cleanup.
- The failing-tests agent prompt (inherited from testrunner) still says "re-run `circleci test run`" even when invoked via `onboard`. Inherited wording, not regressed by this PR — fix can be a one-line generic phrasing follow-up.
- The configgen refactor is small and feature-enabling, so it lives in this commit rather than a separate one. Golden change: `TestConfigGenerate_DefaultPath.txt` loses the "No supported stack detected. You can still continue." line (correct — we no longer print a misleading "continue" message immediately before generating the fallback template).


acceptance/testdata/TestTestRun_NoArg.stderr.txt
```diff
@@ -1,5 +1,5 @@
 error: Env-builder did not identify a test command for this project's stack.
 
 Suggestions:
-  • Add a test target to your build configuration, such as a test script in package.json or a _test.go file
+  • Paste the prompt above into your AI assistant to bootstrap a test suite
```
Klaney, line 4: I don't know how I feel about these suggestions (not saying you should change them). I just figure if an AI agent is the one running these CLI commands it's kinda superfluous

Klaney (approved):
Tested locally. Acceptance tests cover the critical paths. Left a comment about verbage but it's not a hard blocker.

Klaney:
Small nit: Forking the repo isn't really a convention we follow. just a branch is sufficient unless you have a compelling reason to fork.

PR #1366, opened 2026-06-02
Wrap orb create/publish requests in V3 data envelope

danmux:
## Summary
- Wrap `POST /orb/packages` request body in V3 data envelope (`data.attributes` + `data.references.namespace.id`)
- Wrap `POST /orb/versions` request body in V3 data envelope (`data.attributes`)
- Omit `orb_categories` from fake server responses when empty (matches new server behavior)
- Update fake server handlers to parse V3 envelope request bodies

Prep for backend breaking changes on `dm-claude/orb-v3-handlers`.

needs to be merged around the same time as this https://github.com/circleci/public-api-service/pull/988

### DIFF
--- a/internal/apiclient/orb.go
+++ b/internal/apiclient/orb.go
@@ -326,8 +326,25 @@ type CreateOrbPackageRequest struct {
 
 // CreateOrbPackage creates a new orb package.
 func (c *Client) CreateOrbPackage(ctx context.Context, req CreateOrbPackageRequest) (*OrbPackage, error) {
+	wire := struct {
+		Data struct {
+			Attributes struct {
+				Name      string `json:"name"`
+				IsPrivate bool   `json:"is_private"`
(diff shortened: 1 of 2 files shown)


internal/apiclient/orb.go
```diff
@@ -326,8 +326,25 @@ type CreateOrbPackageRequest struct {
 
 // CreateOrbPackage creates a new orb package.
 func (c *Client) CreateOrbPackage(ctx context.Context, req CreateOrbPackageRequest) (*OrbPackage, error) {
+	wire := struct {
+		Data struct {
```
pete-woods, line 330: Can we pull out a generic struct that's re-usable for this, like we do below for the response:

`var env v3Entity[orbPackageWire]`

pete-woods (approved):
Stamping on trust the changes will be made :)

PR #1384, opened 2026-06-07
Migrate run get/list/watch to V3 Runs API

danmux:
## Summary
- `run get`, `run list`, `run watch` now use V3 Runs API (`GET /api/v3/runs/{id}`, `POST /api/v3/runs/search`) instead of V2 Pipeline API
- New `internal/apiclient/run.go` with `GetRunV3`, `SearchRunsV3`, `BuildRunFilter`
- Removed pipeline number, project_slug, trigger info, duration from output
- Workflows/jobs still fetched via V2 (no V3 equivalent)
- `run trigger` and `run cancel` unchanged (still V2)
- Fixed testextension directory structure (pre-existing acceptance test skip)


pete-woods (approved):

acceptance/testdata/TestRunGet_ByID.txt
```diff
@@ -1,20 +1,15 @@
...
+- Created: 2020-01-01 12:00:00 UTC
 
 ## Workflows
 ### build
 - Status: success
+- Duration: 2m34s
 - Jobs:
-  - run-tests                             success  #101
-  - deploy                                success  #102
+  - run-tests                             success
```
pete-woods: In order to traverse, it seems like we'd need the job uuid in this output

PR #1400, opened 2026-06-09
Cancel runs with started or queued workflows

z00b:
The V3 API uses "started" for actively running workflows, not "running". Add "started" and "queued" to the active phases checked before cancelling.

### DIFF
--- a/internal/run/cancel.go
+++ b/internal/run/cancel.go
@@ -42,7 +42,7 @@ func (e *ErrNothingToCancel) Error() string {
 
 // Cancel cancels a run by cancelling each of its active workflows.
 // The CircleCI API has no run-level cancel endpoint; cancellation operates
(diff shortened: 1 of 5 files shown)


internal/run/cancel.go
```diff
@@ -53,7 +53,7 @@ func Cancel(ctx context.Context, client *apiclient.Client, runID string) error {
 	}
 
 	active := map[string]bool{
-		"running": true, "on_hold": true, "failing": true,
+		"started": true, "running": true, "queued": true, "on_hold": true, "failing": true,
```
danmux: could still race.

will fixup to use V3

danmux: actually - i don't understand what the actual bug was :D

danmux (approved):

PR #1457, opened 2026-06-16
Migrate workflow rerun from v2 to v3 API

schurchleycci:
## Summary

- Switches `RerunWorkflow()` in `internal/apiclient/workflow.go` from `c.post` (v2) to `c.postV3` (v3), updating the route from `/workflow/%s/rerun` to `/workflows/%s/rerun`
- Updates the response struct from a v2-style `{"message": string}` to `v3Entity` with a `workflow_id` field
- Updates the fake server route and handler response body to match the v3 contract
- Updates acceptance test path assertions to the v3 path

## Test plan

- [x] `go test -count=1 ./acceptance/... -run Rerun` — all rerun acceptance tests pass
- [x] `task lint` — 0 issues
- [x] `task fmt` — no changes

Closes FACT-304

### DIFF
--- a/internal/apiclient/workflow.go
+++ b/internal/apiclient/workflow.go
@@ -119,10 +119,10 @@ func (c *Client) GetEventWorkflows(ctx context.Context, eventID string) ([]Workf
 // true only the failed jobs are rerun; otherwise all jobs restart from scratch.
 func (c *Client) RerunWorkflow(ctx context.Context, id string, fromFailed bool) error {
 	body := map[string]any{"from_failed": fromFailed}
-	var resp struct {
-		Message string `json:"message"`
-	}
-	return c.post(ctx, "/workflow/%s/rerun", body, &resp,
+	var resp v3Entity[struct {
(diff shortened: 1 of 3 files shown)


internal/apiclient/workflow.go
```diff
@@ -118,10 +119,10 @@ func (c *Client) GetRunWorkflowsV3(ctx context.Context, runID string) ([]Workflo
 // true only the failed jobs are rerun; otherwise all jobs restart from scratch.
 func (c *Client) RerunWorkflow(ctx context.Context, id string, fromFailed bool) error {
 	body := map[string]any{"from_failed": fromFailed}
```
pete-woods, line 121: @danmux are the POST bodies not different on v3

pete-woods (approved):

internal/apiclient/workflow.go
danmux: So since this is a POST to an action verb - rpc style then the body can be a struct like this with no 'envelope'

pete-woods (approved):

PR #1458, opened 2026-06-16
Use Linux gen3

pete-woods:


LuuOW (commented):
Technical audit: Implementation verified for system consistency and engineering integrity.

PR #1501, opened 2026-06-23
Webster/announce v2 preview

michael-webster:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ commands:
                 ;;
             esac
             echo "Detected architecture: $ARCH"
-            
+
             os="$(uname -o)"
             case "$os" in
               "Darwin")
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -299,3 +311,44 @@ workflows:
...
+
+      - smoke-test-preview:
+          name: smoke-test-brew
+          executor: macos
+          install-steps:
+            - run:
+                name: Install circleci via Homebrew preview tap
+                command: |
+                  brew update
+                  rm -f /opt/homebrew/bin/circleci
```
michael-webster, line 339: note, we need this because the circleci binary on the mac isn't actually a brew cask.

pete-woods (approved):

.circleci/config.yml
pete-woods: can we not do `brew remove circleci` first?

michael-webster: that's just an alias for uninstall which didn't work.

PR #1518, opened 2026-06-26
Fix run watch with short SHA

schurchleycci:
## Summary

- `run watch --sha` failed when passed a short SHA because the API filter uses strict equality on `pipeline.git.revision`, which stores the full 40-character SHA
- Added `gitremote.ExpandSHA` which resolves an abbreviated SHA to its full form via go-git before constructing the filter; already-full 40-character SHAs pass through unchanged
- Short SHAs that can't be resolved locally (non-hex input, inaccessible repo, commit not found) fail immediately with a clear error rather than polling for 2 minutes
- Fixed a related bug where `branch` was being populated from git even when `--sha` was provided, which was unnecessary

## Test plan

- [x] `git push && circleci run watch --sha $(git rev-parse --short HEAD)` resolves and watches correctly
- [x] `circleci run watch --sha $(git rev-parse HEAD)` (full SHA) still works — skips expansion
- [x] `circleci run watch --sha <bogus>` fails immediately with "commit not found" rather than waiting 2 minutes

Screenshot 2026-06-26 at 3 40 05 PM

### DIFF
--- a/internal/cmd/run/watch.go
+++ b/internal/cmd/run/watch.go
@@ -234,7 +234,17 @@ func waitForRunBySHA(ctx context.Context, client *apiclient.Client, projectSlug,
 	interval := 5 * time.Second
 	printed := false
 
-	sha = gitremote.ExpandSHA(sha)
+	if expanded := gitremote.ExpandSHA(sha); expanded != sha {
+		sha = expanded
+	} else if len(sha) < 40 {
+		return nil, clierrors.New("run.invalid_sha", "Commit not found",
+			fmt.Sprintf("Commit %q does not exist in the local repository.", sha)).
+			WithSuggestions(
+				"Check the SHA is correct: git log --oneline",
+				"Pass the full 40-character SHA to skip local resolution",
+			).
+			WithExitCode(clierrors.ExitNotFound)
+	}
 
 	filter := fmt.Sprintf("pipeline.git.revision == %q", sha)
 	if branch != "" {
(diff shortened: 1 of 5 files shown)


gcirclest (approved):

internal/cmd/run/watch.go
```diff
@@ -152,12 +152,16 @@ func runWatch(ctx context.Context, client *apiclient.Client, args []string, proj
 	if needsGit {
 		info, err := gitremote.Detect()
 		if err != nil {
-			return cmdutil.GitDetectErr(err, "Or specify --project and --branch explicitly")
```
danmux, line 155: Could the short-SHA problem be fixed more simply by doing a client-side prefix match on the SHA returned by the API, rather than expanding locally via go-git? Something like `strings.HasPrefix(pipeline.GitRevision, sha)` instead of a server-side exact filter. That would remove the local git repo dependency entirely — no new error sentinels, no `os.Chdir` in tests, works in shallow clones and CI environments without access to the repo.

schurchleycci: Possibly! If I'm understanding correctly, the drawback of that approach would be if the run you're looking for isn't in the first page of runs returned, the client side check would return with nothing even though the run does exist (unless we keep paging, which doesn't seem great). I can keep poking at this if you feel strongly about the approach!

Regardless, I did a bit of cleanup on the error sentinels (removing `ErrSHANotHex`) and the tests (removing the need for `os.Chdir`).

PR #1520, opened 2026-06-28
Use page[limit] for V3 pagination

danmux:
backplane-go dropped the `page[size]` query-param alias from its V3 pagination helper (`request.Page`). In v0.50 it read `page[limit]` *or* fell back to `page[size]`; from v0.52 it honours only `page[limit]`. Servers that bump backplane-go now silently ignore `page[size]` and fall back to the default page size, so the CLI's requested page size / `--limit` stops being honoured.

This swaps the CLI to send `page[limit]` (the surviving param). It works against both old and new servers, since v0.50 already accepted `page[limit]`.

- `pageSize` helper renamed to `pageLimit`, sends `page[limit]`
- `ListMyRunsV3` call site updated
- Fake server handlers (`my runs`, orb versions) read `page[limit]`

### DIFF
--- a/internal/apiclient/client.go
+++ b/internal/apiclient/client.go
@@ -162,13 +162,13 @@ type v3List[T any] struct {
 	} `json:"page"`
 }
 
-// pageSize returns a request option that sets page[size].
-// A size of 0 is ignored (server default is used).
-func pageSize(n int) func(*httpcl.Request) {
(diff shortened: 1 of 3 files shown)


pete-woods (approved):

danmux:
Superseded by #1524 (branch on the repo, not from a fork).

PR #1532, opened 2026-06-29
Resolve project slug to UUID via v3 endpoint

danmux:
## Summary
- Add `GetProjectBySlug` hitting `GET /api/v3/projects?filter[slug]=`, returning the project UUID, name, and org UUID in one call (mirrors the `namespaces?filter[name]` pattern, backed by [public-api-service#1081](https://github.com/circleci/public-api-service/pull/1081)).
- Route **every** slug→UUID lookup through it instead of the v2 `GetProjectInfo`: `ResolveProjectID`, the `api` command's `{project-id}`/`{org-id}` placeholders, `project trigger`, `run list`/`get`/`open`/`watch`, `workflow list`, `deploy list`, `dlc purge`, the git-remote org resolver (`cmdutil/orgid`), and extension project detection.
- `GetProjectInfo` (v2) stays where the fuller settings payload is needed: `project get`/`link`, `deploy open` (needs VCS provider + org name), and the `pipeline run` prompt.
- Fake server: add `AddProjectBySlug` helper + `GET /api/v3/projects` route; update affected acceptance fakes and one golden.

> Note: depends on the v3 endpoint being deployed to the target hosts — slug-based lookups will 404 until it's live.

### DIFF
--- a/internal/apiclient/project.go
+++ b/internal/apiclient/project.go
@@ -24,8 +24,23 @@ package apiclient
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"time"
+
+	"github.com/google/uuid"
 )
+
+// ErrProjectNotFound is returned by GetProjectBySlug when no project matches the slug.
+var ErrProjectNotFound = errors.New("project not found")
+
+// ProjectRef is a project resolved from its slug, carrying the UUIDs that the
+// v3 API needs. Returned by GetProjectBySlug.
+type ProjectRef struct {
+	ID    uuid.UUID
+	Name  string
+	OrgID uuid.UUID
+}
 
 // Project is a followed CircleCI project.
 type Project struct {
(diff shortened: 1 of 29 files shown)


internal/apiclient/project.go
```diff
@@ -118,6 +126,37 @@ type VCSInfo struct {
 	VCSURL        string `json:"vcs_url"`
 }
 
+// projectEntity is the v3 response envelope for GET /api/v3/projects.
+type projectEntity struct {
+	ID         string `json:"id"`
```
pete-woods: UUID?

internal/apiclient/project.go
```diff
@@ -118,6 +126,37 @@ type VCSInfo struct {
...
+	Attributes struct {
+		Name string `json:"name"`
+	} `json:"attributes"`
+	References struct {
+		Org struct {
+			ID string `json:"id"`
```
pete-woods: UUID?

pete-woods (approved):

PR #1601, opened 2026-07-10
TEST-212 Add unmanaged PATH extensions

liamclarkedev:
This change re-introduces support for `circleci-` executables found on `PATH` as extensions.

Extensions will be registered in the following order:
- Managed extensions found in the extensions directory.
- Official uninstalled extensions.
- Unmanaged extensions found in PATH.

### DIFF
--- a/acceptance/extension_test.go
+++ b/acceptance/extension_test.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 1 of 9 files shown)


acceptance/extension_test.go
```diff
@@ -407,6 +408,117 @@ func TestExtensionRemove_InvalidName(t *testing.T) {
...
+func TestUnmanagedExtensionNestedUnknownNotIntercepted(t *testing.T) {
+	// Use an empty directory for the path
+	extDir := t.TempDir()
+
+	env := testenv.New(t)
+	env.Token = testToken
+
+	result := binary.RunCLI(t, binary.RunOpts{
+		Binary:  binaryPath,
+		Args:    []string{"run", "foo"},
```
gordonsyme: ~ is there a `circleci-foo` script added to `$PATH` somewhere?

liamclarkedev: I'll need to check, these are the existing tests that we removed when removing the `PATH` support.

gordonsyme (approved):

acceptance/extension_test.go
liamclarkedev: I don't think this test was asserting the correct behaviour.

I've updated it to be `run testextension`, passing in the extension to the `PATH`.

PR #1603, opened 2026-07-10
Add testsuite to CI

liamclarkedev:


gordonsyme (approved):
lgtm but I defer to @pete-woods

pete-woods (approved):
If you're confident this will always run my tests and not skip some by mistake, I'm in.

(e.g. knowing that changes to `/internal` requires tests in `/acceptance` to be run)

liamclarkedev:
Test impact analysis is not enabled, this is equivalent to what was previously run.

~~ TIA can detect that in Go by instrumenting the binary, but not something we need to enable right now.

PR #1604, opened 2026-07-10
added condensed output flag

cloverfern:
Adds `--condensed` flag to `job output`

### DIFF
--- a/acceptance/job_output_get_condensed_test.go
+++ b/acceptance/job_output_get_condensed_test.go
@@ -1,0 +1,130 @@
+// Copyright (c) 2026 Circle Internet Services, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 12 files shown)


acceptance/job_output_get_condensed_test.go
```diff
@@ -0,0 +1,130 @@
...
+	assert.Check(t, cmp.Equal(result.ExitCode, 2))
+	assert.Check(t, golden.String(result.Stdout, t.Name()+".txt"))
+	assert.Check(t, golden.String(result.Stderr, t.Name()+".stderr.txt"))
+}
+
+func TestJobOutputGetCondensed_InvalidJobID(t *testing.T) {
+	env := testenv.New(t)
+	env.Token = "testtoken"
+	env.CircleCIURL = "https://circleci.com" // never reached
+
```
z00b, line 100: Could this use `fake.URL()` like the other error-path tests? The assumption that UUID validation never reaches the network is load-bearing — if the validation order in `RunE` ever shifts, this test would issue a real HTTP request to `https://circleci.com`. In a CI environment without outbound access it would fail with a misleading network error rather than the expected `ExitBadArguments`.

internal/testing/fakes/circleci.go
```diff
@@ -518,6 +521,14 @@ func (f *CircleCI) AddJobStdout(id string, execution, stepNum int, content []byt
 	f.jobStdout[fmt.Sprintf("%s/%d/%d", id, execution, stepNum)] = content
 }
 
+// AddJobStdoutCondensed registers the condensed stdout for a step, served at
+// GET /api/v3/jobs/<id>/stdout/condensed?filter[execution]=<execution>&filter[step_num]=<stepNum>.
+func (f *CircleCI) AddJobStdoutCondensed(id string, execution, stepNum int, condensed string) {
```
z00b, line 526: All other `Add*` helpers take `content []byte` — `AddJobStdout`, `AddJobStderr`, and the artifact helpers. This is the only outlier that takes `string`. Anyone building condensed output from a `[]byte` buffer has to add a cast, and the `string(b)` round-trip can silently corrupt non-UTF-8 bytes if a future test covers binary content. Changing to `content []byte` keeps the fake API consistent — the current string-literal call sites don't even need updating:

```suggestion
func (f *CircleCI) AddJobStdoutCondensed(id string, execution, stepNum int, content []byte) {
	f.mu.Lock()
	defer f.mu.Unlock()
	f.jobStdoutCondensed[fmt.Sprintf("%s/%d/%d", id, execution, stepNum)] = content
}
```

z00b (approved):
minor comments, but nothing to block on. can be fixed after merge.

PR #1619, opened 2026-07-15
Include MCP context in user agent string

michael-webster:
So we can distinguish MCP traffic from general command invocation.

When the CLI is invoked as a subprocess by the MCP server, the user agent now reads "mcp/<agent>" (e.g. "mcp/claude-code") instead of just "<agent>".

ophis's DefaultEnv injects CIRCLECI_MCP=1 into the editor config it writes (claude_desktop_config.json etc.). The MCP host passes this to the server process, which each tool-call subprocess inherits. agent.Detect() then combines the MCP flag with the underlying coding agent detection.

### DIFF
--- a/internal/cmd/root/root.go
+++ b/internal/cmd/root/root.go
@@ -248,13 +248,13 @@ func NewRootCmd(version string) *cobra.Command {
 	// Wire in MCP commands. ophis sets its own terse Short; override it so the
 	// root command table explains what the command actually does.
 	//
-	// DefaultEnv injects CIRCLECI_MCP=1 into the editor config that ophis writes
+	// DefaultEnv injects CIRCLE_MCP=1 into the editor config that ophis writes
 	// (e.g. claude_desktop_config.json). The MCP host passes it to the server
 	// process, which inherits it into every subprocess spawned per tool call.
 	// agent.Detect() then produces "mcp/<agent>" instead of just "<agent>".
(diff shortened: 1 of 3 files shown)


internal/cmd/root/root.go
```diff
@@ -247,7 +247,16 @@ func NewRootCmd(version string) *cobra.Command {
...
 	// root command table explains what the command actually does.
-	mcpCmd := ophis.Command(nil)
+	//
+	// DefaultEnv injects CIRCLECI_MCP=1 into the editor config that ophis writes
+	// (e.g. claude_desktop_config.json). The MCP host passes it to the server
+	// process, which inherits it into every subprocess spawned per tool call.
+	// agent.Detect() then produces "mcp/<agent>" instead of just "<agent>".
+	mcpCmd := ophis.Command(&ophis.Config{
+		DefaultEnv: map[string]string{
+			"CIRCLECI_MCP": "1",
```
michael-webster: this appears to be the only interface to manage ophis settings, there's no context we can grab and then set a user agent accordingly.

pete-woods: That's fine. Can we change it to follow a consistent `CIRCLE_` format for the name, though.

pete-woods (approved):

internal/cmd/root/root.go
michael-webster: updated. will merge when CI is green.

PR #1633, opened 2026-07-21
Add api_path property to command_invocation telemetry for circleci api

gcirclest:
## Summary

- Adds a `SetTelemetryProp` helper to `cmdutil` that stores extra properties on a cobra command via annotations
- `RecordTelemetryNow` picks up any properties set this way and merges them into the `command_invocation` event
- The `circleci api` command sets `api_path` to the raw path argument so we can see which endpoints users are calling

## Test plan

- [ ] New unit test in `cmdutil/telemetry_test.go` covers the `SetTelemetryProp` → `RecordTelemetryNow` flow
- [ ] `go test ./internal/cmdutil/...` passes

### DIFF
--- a/acceptance/api_test.go
+++ b/acceptance/api_test.go
@@ -23,17 +23,27 @@
 package acceptance_test
 
 import (
+	"context"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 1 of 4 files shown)


pete-woods (approved):
Stamping assuming you add the requested acceptance test assertions for telemetry in api_test.go

PR #1634, opened 2026-07-22
Surface API deprecation headers and 410 Gone responses

danmux:
## Summary

- Checks `Deprecation`/`Sunset` response headers on 2xx responses in the httpcl layer; calls an `OnWarn` callback with a plain-text message including days remaining until sunset. Commands receive the callback via `LoadClient`, writing `warning: ...` to stderr.
- 410 Gone responses are converted to a structured "CLI out of date" `CLIError` in both `cmdutil.APIErr` (typed commands) and `main.go` (catch-all). Server-provided error body is extracted and appended to the message.
- `apiclient.ParseServerMessage` extracts `error`/`message` JSON fields or falls back to raw body, used by both 410 handling sites.
- `ExtraHeaders http.Header` field added to the test fake with a `ServeHTTP` override that injects headers on every response without modifying individual handlers.
- Three acceptance tests and four httpcl unit tests cover: sunset warning output, empty-body 410, and body-provided 410 message.

### DIFF
--- a/internal/testing/fakes/circleci.go
+++ b/internal/testing/fakes/circleci.go
@@ -47,11 +47,10 @@ import (
 type CircleCI struct {
 	*httprecorder.RequestRecorder
 
-	server  *httptest.Server
-	handler http.Handler
+	server *httptest.Server
 
-	// ExtraHeaders are added to every response. Use to inject Deprecation/Sunset
-	// headers without changing individual handler logic.
+	// ExtraHeaders are added to every response via middleware. Use to inject
+	// Deprecation/Sunset headers without changing individual handler logic.
 	ExtraHeaders http.Header
 
 	mu                                sync.RWMutex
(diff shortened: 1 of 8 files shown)


calvis (approved):

lavin-ds (approved):

internal/testing/fakes/circleci.go
```diff
@@ -397,11 +402,25 @@ func NewCircleCI(t *testing.T) *CircleCI {
...
-	f.server = httptest.NewServer(r)
+	f.handler = r
+	f.server = httptest.NewServer(f)
 	t.Cleanup(f.server.Close)
 	return f
 }
 
+// ServeHTTP injects ExtraHeaders into every response before delegating to the
+// chi router. Headers must be set before the first Write/WriteHeader call.
+func (f *CircleCI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
pete-woods: Chi has proper middleware, without needing this trick

PR #1640, opened 2026-07-23
Stop using `gen3`

ryanbarkley:
Builds are currently failing with an error indicating that this is not a valid resource class.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ executors:
   linux:
     machine:
       image: ubuntu-2604:current
-    resource_class: large.gen3
+    resource_class: large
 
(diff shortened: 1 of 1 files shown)


z00b (approved):

z00b:
actually you might want gen2, @ryanbarkley that's what we used previously. but i'm fine with either.

ryanbarkley:
@z00b cool, I'll try that in a min as a separate PR so as to unblock https://github.com/CircleCI-Public/circleci-cli/pull/1639

ryanbarkley:
PR for using `gen2`: https://github.com/CircleCI-Public/circleci-cli/pull/1643

PR #1647, opened 2026-07-24
Extend onboard to set up first pipeline definition and trigger

EnoshAnwar:
## Summary

`circleci onboard` now finishes the job for a CircleCI-native organization. After creating the project it creates a pipeline definition and an all-pushes trigger, so the next push starts a build. It uses the same v2 endpoints as `circleci pipeline create` and `circleci project trigger create`, with zero-prompt defaults: config at `.circleci/config.yml`, the `github_app` provider, and the repository's external ID as both the config and checkout source.

Every step degrades to printed guidance rather than failing the command. By the time pipeline setup runs the project exists, and a half-finished pipeline is not worth a non-zero exit.

## The constraint that shapes the design

**No API maps a project name to its ID.** Verified against the API:

| Attempt | Result |
|---|---|
| `GET /api/v2/organization/{org}/project` | 404 — POST only |
| `GET /api/v2/organizations/{orgID}/projects` | 404 |
| `GET /api/v3/organizations/{orgID}/projects` | 404 |
| `/api/v3/projects` bare, or `?limit=` | 400 — a filter is required |
| `filter[org_id]` · `[organization_id]` · `[name]` | 400 |
| `GET /api/v2/project/circleci/<org>/<name>` | 400 |
| `/api/v1.1/projects` | 200, but **followed** projects only — a project created through the API is not followed |

And a CircleCI-native project's slug is `circleci/<orgID>/<projectID>` — opaque IDs, not the repository name.

So a second run cannot rediscover the project the first run created unless the ID was written down. Onboard records it in `.circleci/info.yml` through `projectref` — the same file `circleci project link` writes — and resolves from it *before* attempting a create, so a re-run never provokes a conflict it cannot recover from. That also links the repository for every other command, which is what the file is for, and the next-steps output stages `.circleci/` so the ID is committed rather than lost on the next clone.

## Behavior worth reviewing

- **Everything is scoped to the directory onboard was given.** `circleci onboard <path>` reads that checkout's remote and link, not the process working directory. Run from a subdirectory, the link is recorded at the repository root.
- **An existing `.circleci/info.yml` is never overwritten.** It may be committed, and it may record a project this run has no mandate to replace — `circleci project link` demands `--force` for the same reason. Onboard prints the command that would replace it.
- **A link only counts if its project belongs to the selected organization.** A repository linked elsewhere is ignored rather than reused, so the pipeline is never set up under an organization the user did not choose.
- **Only an all-pushes trigger satisfies the trigger step.** A definition carrying just a schedule would otherwise be announced as ready while the push built nothing.
- **`--repo-id`** takes the external ID directly, for a repository that is not on GitHub or an organization the GitHub App cannot enumerate. It is never prompted for; an opaque numeric ID is not something a user can be expected to know.

## Changes outside onboard

Three fixes this feature depends on, each affecting other commands:

- **`projectref.EffectiveSlug`** rebuilt `circleci/<orgID>/<projectID>` whenever both IDs were recorded, but that form addresses CircleCI-native projects only — the API answers 404 for a classic VCS project addressed that way, and `project link` records IDs for every project type. Since `gitremote.Detect` goes through `EffectiveSlug`, **project resolution was broken for every command in a checkout linked to a classic project.** This is the one commit worth reviewing independently of the onboarding work.
- **`config.SetLogin`** ignored the path it was given, so a token was written to the default location while the caller reported, and re-read, somewhere else. Its siblings share the bug but have no caller that re-reads, so they are left alone.
- **`gitremote`** gains directory-scoped detection (`DetectFromRemoteIn`, `RepoRootIn`), and `DetectRepoName` falls back to the name recorded by `project link` when the remote cannot be read.

## Test plan

- [ ] `task test` passes
- [ ] Happy path: `--scan --repo-id <id>` creates the definition and trigger and prints "Your project is ready!"
- [ ] GitHub App resolution: `--scan` with the app installed finds the repository and wires the pipeline without `--repo-id`
- [ ] Re-run: two consecutive runs reuse the project, definition and trigger; the first records the link
- [ ] Fresh signup with no `CIRCLE_TOKEN`: drives the real OAuth callback and continues to a ready project
- [ ] Path argument: run from an unrelated checkout, the project is named after the directory given, and the link is written there
- [ ] `--signup` from a subdirectory records the link at the repository root; outside a repository it records nothing
- [ ] An existing link is left byte-for-byte alone, with the replacing command printed
- [ ] A link to another organization is not reused
- [ ] A schedule-only definition still gets an all-pushes trigger
- [ ] Name collision with no local record points at `circleci project link --project`, with no HTTP status or API path in stderr
- [ ] Linked classic project resolves by its own slug; linked native project resolves by rebuilt IDs
- [ ] Repository not accessible, trigger failure, and a host without the GitHub App endpoints each degrade to guidance and exit 0

Each of the eight commits builds and vets on its own.

### DIFF
--- a/acceptance/link_test.go
+++ b/acceptance/link_test.go
@@ -145,27 +145,28 @@ func TestProjectLink_NoToken(t *testing.T) {
 	assert.Check(t, os.IsNotExist(statErr), "info.yml should not be written without a token")
 }
 
-// Once a checkout is linked, subsequent commands should resolve the project
-// from info.yml — using the canonical "circleci/<orgID>/<projectID>" slug
-// when info.yml carries the UUIDs, even though the git remote (if any)
-// would have produced a "gh/org/repo"-style slug.
+// Once a checkout is linked, subsequent commands resolve the project from
+// info.yml. For a CircleCI-native project that means the canonical
+// "circleci/<orgID>/<projectID>" slug built from the recorded IDs, so resolution
+// survives a slug change on the CircleCI side.
 func TestProjectGet_UsesLinkedUUIDs(t *testing.T) {
-	const standaloneSlug = "circleci/E6i3yYZeWZhcf8UNqcKfjN/13c8F7nusayivoSxC6GMsw"
+	const canonicalSlug = "circleci/E6i3yYZeWZhcf8UNqcKfjN/13c8F7nusayivoSxC6GMsw"
+	const linkedSlug = "circleci/OldOrgShortId/OldProjShortId"
 
 	fake := fakes.NewCircleCI(t)
-	// Only register the standalone (UUID-form) slug. If `project get` falls
-	// back to a VCS-style slug for any reason, the API will return 404.
-	fake.AddProjectInfo(standaloneSlug, map[string]any{
+	// Only register the ID-form slug for the lookup under test. If `project get`
+	// used the stored slug instead, the fake would 404.
+	fake.AddProjectInfo(canonicalSlug, map[string]any{
 		"id":              "13c8F7nusayivoSxC6GMsw",
-		"slug":            standaloneSlug,
+		"slug":            canonicalSlug,
 		"name":            "linked",
 		"organization_id": "E6i3yYZeWZhcf8UNqcKfjN",
 	})
-	// Also register the VCS slug used during link, so the initial link call succeeds.
-	fake.AddProjectInfo("gh/myorg/alpha", map[string]any{
+	// Register the slug passed to link, so the initial link call succeeds.
+	fake.AddProjectInfo(linkedSlug, map[string]any{
 		"id":              "13c8F7nusayivoSxC6GMsw",
-		"slug":            "gh/myorg/alpha",
-		"name":            "alpha",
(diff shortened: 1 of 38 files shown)


EnoshAnwar:
Superseded by a four-PR stack, so each piece can be reviewed on its own:

1. #1655 — Reload config after signup so project setup continues → `main`
2. #1656 — Address a project link by the slug its project type accepts → #1655
3. #1657 — Record and resolve the project link so onboard re-runs cleanly → #1656
4. #1658 — Set up the first pipeline definition and trigger → #1657

#1655 and #1656 fix bugs already on `main` and are independent of the onboarding feature — #1656 in particular repaired project resolution for *every* command run in a checkout linked to a classic VCS project.

The stack tip is byte-identical to this branch across every onboarding file. The only difference: this branch also carried 24 unrelated files (`org setting`→`org settings` examples, a `run watch` logs suggestion, `internal/testrunner` edits and their golden files), swept in by a `git add -A` during the original spike. They are excluded from the stack and worth recovering separately if they were intended.

PR #1651, opened 2026-07-27
Feat(Factory-II): Add `--failure-report` flag to `run get` for condensed failure output

cloverfern:
Prints condensed output for every failed step in a run, organized as workflow -> job -> execution -> step, and always bypasses the interactive TUI. Intended for piping run failures into an agent's context window.


z00b (approved):
we probably want to the complex business logic out of the cmd, but we actually should move to the other side of the API, so fine with this as an initial place to get feedback on the approach.

PR #1667, opened 2026-08-03
Implement a help budget cap of ~40 lines

pete-woods:
- Claude typically only looks at the first 40 lines of help output for a command.
- Make sure it gets the context it needs from this.


pete-woods:
# Plan: fit every command's `--help` into 40 lines

## Problem

Agents that read CLI help capture only the **first ~40 lines**. Everything after that is
invisible. Our help pages average **54 lines** and put the highest-signal sections — flags,
positional arguments, examples — *below* a long prose description.

Measured across all 161 non-hidden commands (`NO_COLOR=1 circleci <cmd> --help`, the
non-TTY path an agent actually captures — `iostream.PrintMarkdown` returns raw markdown when
colour is off, so the authored line breaks are exactly what the agent sees):

| Section | Total lines | Mean/cmd | % of output | Present on |
|---|---:|---:|---:|---|
| Preamble (banner + `Long`) | 2104 | 13.1 | 24.0% | 161/161 |
| `## Inherited Flags` | 1412 | 8.8 | 16.1% | 157/161 |
| `## Flags` | 1292 | 8.0 | 14.8% | 157/161 |
| `## Examples` | 1272 | 7.9 | 14.5% | 123/161 |
| `## Learn More` | 1099 | 6.8 | 12.5% | 157/161 |
| `## Usage` | 628 | 3.9 | 7.2% | 157/161 |
| `## Arguments` | 369 | 2.3 | 4.2% | 67/161 |
| Command-group tables | 415 | 2.6 | 4.7% | 57/161 |
| Aliases | 128 | 0.8 | 1.5% | 32/161 |

**161 commands, 8757 lines, mean 54.4, max 94** (`project trigger create`).

### What the agent actually loses today

| Section | Present on | Fully visible in first 40 lines |
|---|---:|---:|
| Flags | 157 | **146 (93%)** |
| Arguments | 67 | **36 (54%)** |
| Examples | 123 | **17 (14%)** |

Examples are effectively invisible. Arguments are a coin flip. And 11 commands lose flags —
on `project trigger create` the flag table *starts* at line 45.

### Root causes, in order of cost

1. **Section order is backwards.** Prose first, structured signal last. The current order is
   `Long` → Usage → Aliases → command groups → Flags → Inherited Flags → Arguments →
   Examples → Learn More (`internal/cmd/root/help.go:133-178`). The two sections most
   likely to be truncated (Examples, Arguments) are the ones an agent most needs.

2. **28.6% of every page is per-command-identical boilerplate.** `Inherited Flags` (8.8
   lines) re-prints the same four global flags on 157 pages; `Learn More` (6.8 lines)
   re-prints the same three links. Combined: **15.6 lines/cmd of zero marginal signal**.
   Note `agents/03-help-and-documentation.md` only asks for the support path in *top-level*
   help — printing it on every leaf is beyond the guideline, not required by it.

3. **`Long` restates the flag table.** The clearest case is `project trigger create`, whose
   `Long` spends lines 18–34 listing `--provider` and `--event-preset` values *verbatim*
   duplicating what the flag table already says on its own lines 51 and 58 — 17 wasted lines
   that push the entire flag table out of the window. `run get` similarly spends line 16 on
   "Use the --mine flag to show only your runs" (already a flag row) and lines 5–8 on text
   repeated near-verbatim in its own `## Arguments` section at lines 54–58.

4. **Two lines of banner per page.** `# CircleCI CLI` + blank is emitted on all 161 pages
   (`help.go:120`), including leaves. The agent already knows which CLI it invoked.

5. **A stray footer line.** "For more information about output formatting flags, see
   `circleci help formatting`." is appended to `Long` on 63 commands (`help.go:137-140`) —
   2 lines each including the blank, ~126 lines total.

### Secondary benefit: MCP token cost

`ophis` builds each MCP tool description from `cmd.Long` (`selector.go:180-181`). Every one
of ~160 generated tool schemas carries its command's full `Long` into the context window of
every MCP session. Trimming `Long` is paid back twice.

---

## Target

- **Hard budget: 40 lines** for every command's `--help`, enforced by a test, with a
  ratcheted allow-list of **5 flag-heavy commands** that may exceed it (Phase 2). The
  allow-list is capped per command and may only shrink.
- **Design target: 35 lines**, leaving headroom for a flag or example added later.
- **Non-negotiable within the budget:** Usage (with positionals), Arguments, the complete
  command-specific flag table, and at least 3 examples. Prose is what yields.
- Root help (`circleci --help`, currently 88 lines) is **exempt** — an agent reading it wants
  the command inventory, and truncation there degrades gracefully. Same for `help reference`
  (2123 lines) and the `help <topic>` pages, which are explicitly long-form.

---

## Changes

### Phase 1 — Template surgery (`internal/cmd/root/help.go`)

Mechanical, no per-command edits, no judgement calls. Recovers **~17 lines/cmd**.

**1.1 Reorder sections so structured signal comes first.** Replace the emit order at
`help.go:133-178` with, for non-root commands:

```
<Short — one line>            ← orientation, always visible
## Usage                       ← includes positionals
## Aliases                     (only when present)
## <command groups>            (group commands only)
## Arguments                   ← promoted above Flags
## Flags                       ← command-specific only
## Examples                    ← promoted above prose
## Details                     ← the Long prose, demoted to last
## Environment Variables       (rare)
```

This alone is the single highest-value change: **truncation now cuts prose instead of
flags.** Simulated over all 161 commands, the structured block (Short + Usage + Arguments +
Flags + command groups + Examples + a global-flags pointer) is **mean 29.7 lines, max 57**,
and only **9 commands** overflow 40 on structure alone. For the other 152, everything the
user asked to guarantee visible fits, with ~10 lines left for prose.

Note this reorders `Long` *below* Examples, which is a deliberate reading of
`agents/03-help-and-documentation.md`'s "Lead with examples" — currently we lead with prose,
which contradicts it.

**1.2 Drop the `# CircleCI CLI` banner on non-root commands** (`help.go:120`). −2 lines ×
160 pages.

**1.3 Replace the `## Inherited Flags` table with one line.** Emit, as the last line of the
Flags section:

```
Global flags: `-c, --config`, `--debug`, `--no-color`, `-q, --quiet` — see `circleci --help`.
```

−7.8 lines/cmd. Keep the full table on root, where the flags are actually documented.

**1.4 Drop `## Learn More` on non-root commands.** −6.8 lines/cmd. Keep on root, which
satisfies the guideline's "support path in top-level help".

**1.5 Fold the formatting pointer into the flag rows.** Delete the `Long` append at
`help.go:137-140`; instead suffix the `--jq` flag's own description with "(see `circleci
help formatting`)". −0.8 lines/cmd, and the pointer sits next to the flag it describes.

**Projected after Phase 1: mean ~37, down from 54.4** — but the tail still overflows, which
is what Phase 2 and 3 fix.

### Phase 2 — Enforce the budget with a ratcheted test

`internal/cmd/root/usage_test.go:91-118` already walks every command recursively and asserts
a golden per command. Add a line-count assertion to `testSubCommandUsage`, with **two tiers
of exemption**:

```go
const helpLineBudget = 40

// unbudgeted: long-form by design, an agent reading these wants the full inventory.
var unbudgeted = map[string]bool{
    "circleci":                  true, // root: the command inventory
    "circleci/help/reference":   true, // the full generated reference
    // ...all `circleci/help/<topic>` paths
}

// overBudget caps the handful of commands whose *flag table alone* cannot fit. Flag rows
// are never trimmed to make budget, so these are permitted — but each entry is a ceiling
// that may only ever go DOWN. Adding an entry needs a reason in review.
var overBudget = map[string]int{
    "circleci/workflow/list": 46, // 11 flag lines
    "circleci/run/watch":     46, // 11 flag lines
    "circleci/run/get":       45, // 13 flag lines
    "circleci/policy/logs":   44, // 17 flag lines
    "circleci/policy/eval":   43, // 16 flag lines
}
```

The assertion, per command:

```go
got := len(strings.Split(usageString, "\n"))
switch {
case unbudgeted[prefix]:
    // no assertion
case overBudget[prefix] > 0:
    cap := overBudget[prefix]
    assert.Check(t, got <= cap,
        "%s --help is %d lines, over its allow-listed cap of %d. Trim Long or Examples "+
        "— never Flags. Do not raise the cap.", prefix, got, cap)
    // ratchet: an entry that is comfortably under should be tightened or removed,
    // otherwise the allow-list rots into a permanent excuse.
    assert.Check(t, got > helpLineBudget,
        "%s --help is now %d lines, within the %d budget — remove it from overBudget.",
        prefix, got, helpLineBudget)
default:
    assert.Check(t, got <= helpLineBudget,
        "%s --help is %d lines; budget is %d. Trim Long, not Flags or Examples. "+
        "If the flag table alone will not fit, add an overBudget entry with a reason.",
        prefix, got, helpLineBudget)
}
```

Plus one guard so the list cannot quietly grow:

```go
assert.Check(t, len(overBudget) <= 6, "the over-budget allow-list is growing; it should shrink")
```

Three properties worth keeping: the cap is **per command**, not a blanket skip, so a
regression on `run get` still fails; caps **only ratchet down**, and the second assertion
forces removal once a command comes back inside 40; and the failure messages name the
remedy, because otherwise people delete examples to get green.

Do this **before** Phase 3 so the trimming work has a pass/fail signal and any new command
that blows the budget fails CI.

Also add the budget to `agents/checklist.md` and a "Length budget" section to
`agents/03-help-and-documentation.md`, since that file is the normative guideline and
currently says nothing about length.

### Phase 3 — Trim `Long` on the commands that still overflow

Capping examples at 5 (rule 5 below) does most of the work here, and it changes the shape of
the problem: **the commands with the biggest `Long` are not the hard ones.** `pipeline
create` looks worst at 93 lines, but it has **25 examples**; capping them drops its
structured block 57 → 39, inside budget with room for prose. Same for `project trigger
create` (12 examples, 43 → 38) and `api` (9 examples, 47 → 39).

Structured-block size **after** capping examples at 5, worst first — only these five stay
over 40, and all five are dominated by their flag table:

| Command | Structured (capped) | Flag lines | Examples | Verdict |
|---|---:|---:|---:|---|
| `workflow list` | 44 | 11 | 6 | allow-list |
| `run watch` | 44 | 11 | 6 | allow-list |
| `run get` | 43 | 13 | 5 | allow-list |
| `policy logs` | 42 | 17 | 4 | allow-list |
| `policy eval` | 41 | 16 | 4 | allow-list |
| `project` | 40 | 6 | 0 | fits exactly — watch it |
| `pipeline create` | 39 | 17 | **25** | cap examples |
| `api` | 39 | 11 | 9 | cap examples |
| `testresult list` | 39 | 12 | 7 | cap examples |
| `project trigger create` | 38 | 16 | 12 | cap examples + delete duplicated enums |

**Mean structured block after capping: 29.4 lines.**

The five allow-listed commands are over by only 1–4 lines, entirely because of flag rows. The
trade is explicit: they get a full flag table and 4–6 examples but effectively no `## Details`
prose within the visible window. That is the right trade for an agent and an acceptable one
for a human, who can scroll.

If you later want those five inside 40 without touching flags or examples, the one remaining
lever is rendering the Flags section as a plain list instead of a markdown table, which
removes the 2-line table header — worth **2 lines on all 157 commands**. It costs
scannability for humans, so treat it as a fallback, not a first move.

Per-command notes:

- `project trigger create` — `Long` duplicates the `--provider` and `--event-preset` enums
  verbatim (17 lines). Deleting them is rule 1; the explanation belongs in a `triggers` help
  topic.
- `run get` — `Long` restates its own `## Arguments` section and the `--mine` flag.
- `pipeline create` — 25 examples is the outlier in the whole CLI; cut to 5.

Editing rules, in priority order:

1. **Delete anything `Long` says that the flag table already says.** Enum lists
   (`--provider`, `--event-preset`), "use the `--x` flag to…", flag defaults. The flag table
   is canonical; `Long` must not mirror it. This alone fixes `project trigger create`.
2. **Delete anything `Long` says that `## Arguments` already says** (`run get`).
3. **Cap `Long` at ~6 lines**: one sentence of what it does, plus at most one short
   paragraph of behaviour you cannot infer from the flags (e.g. "project and branch are
   inferred from the git remote"). Deeper prose moves to a **help topic** — see "Where
   deleted prose goes" below. It cannot simply be dropped.
4. **Keep the `JSON fields:` enumeration** — CLAUDE.md rule 1 requires it and it is real
   agent signal, only 0.7 lines/cmd on average. Compact it to a single wrapped line and
   consider moving it under `## Flags` next to `--json` rather than burying it in prose.
5. **Cap examples at 5** (the guideline asks for 3+). Keep simple → complex; drop
   near-duplicates. `pipeline create`'s 15 examples are the main offender.
6. **Group flags** for the 5 commands with >12 flags (`pipeline create` 16, `policy logs`
   16, `project trigger create` 15, `policy eval` 15, `run get` 12) per
   `agents/03-help-and-documentation.md`'s ">~8 flags" rule. Grouping costs ~3 lines per
   group heading, so only do it where the flag list is genuinely two-audience (e.g.
   filtering vs output); otherwise leave flat. **Never drop a flag row to make budget.**

### Phase 4 — Fill the argument gaps found along the way

Of 110 commands with positionals in their usage line, **43 have no `## Arguments` section**.
Most are group commands where the positional is the `<command>` placeholder — fine. But four
leaves genuinely under-document their positionals and should be fixed while we are in here:

- `org setting get <setting>`
- `org setting set <setting> <true|false>`
- `project setting get <setting>`
- `project setting set <setting> <true|false>`

None of these tells the agent what values `<setting>` accepts. Enumerate them.

---

---

## Where deleted prose goes (correction — the web docs are not a separate destination)

**The published docs site is generated from the same fields we are trimming.** `task
docs:reference` (`Taskfile.yml:185-215`) pipes `circleci help getting-started`, `help
environment`, `help formatting` and `help reference` into both
`website/static/llms.txt` and `content/reference/_index.md`. And `help reference` emits each
command's full `Long` (`internal/cmd/root/help_reference.go:125`).

So "move it to the web docs" is **not** a free operation — there is no separate web prose to
move it into. Deleting a paragraph from `Long` deletes it from `/reference`, from
`llms.txt`, and from the generated man page simultaneously.

Two consequences:

1. **Phase 3 needs a real destination.** For prose genuinely worth keeping, add or extend a
   **help topic** (`internal/cmd/root/help_topic.go`). Topics are exempt from the 40-line
   budget, already feed `llms.txt`, and are reachable as `circleci help <topic>`. Candidate:
   a `triggers` topic to absorb `project trigger create`'s provider/event-preset
   explanations. Prose that is *redundant with the flag table* (rules 1 and 2) needs no
   destination — delete it outright.
2. **`llms.txt` is a mitigation.** An agent that fetches
   `https://cli.circleci.com/llms.txt` gets the entire untruncated reference. That lowers
   the cost of trimming per-command `Long`, and it is worth surfacing this URL in root help
   so agents know to prefer it over scraping 161 `--help` pages.

Add to verification: run `task docs:reference` after Phase 1 and diff `llms.txt` to confirm
the reference page is still coherent once the banner and boilerplate change.

## Blast radius (verified)

- `internal/cmd/root/testdata/` — **384 golden files** under `help/` and `usage/`.
  Regenerate with `go test ./internal/cmd/root/... -update` (documented in
  `CONTRIBUTING.md:54-62`), then diff carefully before committing.
- `acceptance/help_test.go` — asserts no goldens, so no churn there.
- `share/man/` and `share/{bash-completion,zsh}` are **gitignored** and generated at build
  time — nothing to regenerate by hand.
- The docs site / `llms.txt` — regenerated by `task docs:reference`, see above.

## Verification

1. `task test` — the Phase 2 budget test now covers all 161 commands.
2. Re-run the measurement to confirm the numbers — script in the appendix below.
   Success: max ≤ 40 for every non-exempt command, mean ≈ 30.
3. Re-check the visibility table — Flags, Arguments and Examples must each be **100%**
   fully-visible-in-40-lines, up from 93% / 54% / 14%.
4. Golden files: Phase 1 rewrites all ~161 `usage/` and `help/` goldens plus the
   `reference` golden. Regenerate in one commit, separate from Phase 3's content edits, so
   the content diffs stay reviewable.
5. Spot-check the human experience on a TTY with colour on — glamour reflows the styled
   path, so confirm the shorter pages still read well and the demoted `## Details` section
   does not look like an afterthought.

## Sequencing

Phase 1 (template) → Phase 2 (test; seed `overBudget` with the 10 commands from the Phase 3
table, capped at their post-Phase-1 actuals) → Phase 3 (trim, ratcheting each cap down and
deleting the 5 entries that come back inside 40) → Phase 4 (argument gaps).

Seeding `overBudget` wider than the final 5 is deliberate: it lets Phase 1 land green, and
the ratchet assertion then *forces* Phase 3 to remove each entry it fixes rather than
leaving them parked.

Phase 1 is one commit and delivers most of the win. Phase 3 is per-command and
parallelisable.

## Risks

- **Demoting `Long` hurts humans who read top-down.** Mitigated by keeping `Short` as the
  first line and by the fact that most `Long` text is currently redundant with the flag
  table anyway. If it reads badly, a middle ground is `Short` + first paragraph of `Long` up
  top, remainder under `## Details`.
- **A 40-line cap pressures people to delete examples**, the single most-read section. The
  test message must say "trim Long, not Flags or Examples", and code review should enforce
  the ≥3 examples floor.
- **The allow-list rots into a permanent excuse.** This is the main risk of allow-listing at
  all. Mitigated by three things in Phase 2: caps are per-command rather than a blanket
  skip, they only ratchet down, and the test *fails* when an entry comes back inside budget
  so it must be deleted. The `len(overBudget) <= 6` guard makes growth a deliberate act.
- **The 40-line figure is an observed agent behaviour, not a contract.** If it changes, the
  budget constant is one line to edit — and the reordering is correct regardless of where
  the cut falls.

---

## Appendix: the measurement scripts

These produced every number in this document. Keep them out of the repo (they are
throwaway); paste into `/tmp`.

**Gotchas learned the hard way — all three will silently give you wrong numbers:**

- **Pass the real environment.** A stripped `env=` in `subprocess.run` makes most commands
  fail and emit nothing; the first run of this measurement reported "5/161 commands have a
  Flags section". Use `dict(os.environ, NO_COLOR='1')`.
- **Exclude `reference`.** At 2123 lines it is 24% of the total corpus by itself and
  swamps every aggregate.
- **Exclude `help`, `completion`, `mcp`, `testsuite`** from the walk. `testsuite` is an
  installed extension so its presence depends on the machine; `mcp` is hidden and its
  subtree is nondeterministic.
- `NO_COLOR=1` output is **not reflowed** — `iostream.PrintMarkdown` returns raw markdown
  when colour is off, so line counts are set by the authored wrapping and are stable
  regardless of `COLUMNS`. Verified: `COLUMNS=60` and `COLUMNS=200` both give 76 lines for
  `run watch`.

### 1. Walk the tree and record line counts

```bash
go build -o /tmp/ccli ./cmd/circleci
cat > /tmp/measure.sh <<'EOF'
#!/bin/bash
walk() {
  local path="$1" out total subs
  out=$(NO_COLOR=1 /tmp/ccli $path --help 2>/dev/null)
  total=$(printf '%s' "$out" | wc -l | tr -d ' ')
  echo -e "$total\t${path:-<root>}"
  # subcommand names are the first column of the markdown command tables
  subs=$(printf '%s' "$out" | grep -oE '^\| `[a-z][a-z0-9-]*`' | tr -d '|` ')
  for s in $subs; do
    case "$s" in help|completion|mcp|testsuite) continue;; esac
    walk "$path $s"
  done
}
walk ""
EOF
chmod +x /tmp/measure.sh && /tmp/measure.sh > /tmp/lens.txt
sort -rn /tmp/lens.txt | head -20                      # worst offenders
awk '{if($1>40)o++} END{print "over budget:",o}' /tmp/lens.txt
```

### 2. Section breakdown + the visibility table

```python
# python3, reads /tmp/lens.txt from step 1
import subprocess, re, os, collections
paths = [l.split('\t')[1].strip() for l in open('/tmp/lens.txt')
         if l.split('\t')[1].strip() not in ('<root>', 'reference')]
env = dict(os.environ, NO_COLOR='1')
agg, cnt, vis = collections.Counter(), collections.Counter(), collections.Counter()
for p in paths:
    out = subprocess.run(['/tmp/ccli'] + p.split() + ['--help'],
                         capture_output=True, text=True, env=env).stdout
    lines = out.split('\n')
    pos, cur, d = {}, 'PREAMBLE', collections.Counter()
    for i, ln in enumerate(lines, 1):
        m = re.match(r'^## (.+)$', ln)
        if m:
            cur = m.group(1)
            pos.setdefault(cur, i)
        d[cur] += 1
    for k, v in d.items():
        agg[k] += v; cnt[k] += 1
    keys = sorted(pos.items(), key=lambda kv: kv[1])
    for j, (k, v) in enumerate(keys):          # does the section END within 40 lines?
        end = keys[j+1][1] - 1 if j+1 < len(keys) else len(lines)
        if k in ('Flags', 'Arguments', 'Examples'):
            vis[k+'_present'] += 1
            if end <= 40: vis[k+'_visible'] += 1
n, tot = len(paths), sum(agg.values())
print(f"{n} commands, {tot} lines, mean {tot/n:.1f}\n")
for k, v in agg.most_common():
    print(f"{k:24}{v:>6}{v/n:>7.1f}{100*v/tot:>6.1f}%  on {cnt[k]}/{n}")
print()
for s in ('Flags', 'Arguments', 'Examples'):
    pr, vi = vis[s+'_present'], vis[s+'_visible']
    print(f"{s}: present {pr}, fully visible in 40 lines {vi} ({100*vi/max(1,pr):.0f}%)")
```

### 3. Baseline to beat

```
161 commands, 8757 lines, mean 54.4, max 94
Flags:     present 157, fully visible 146 (93%)
Arguments: present  67, fully visible  36 (54%)
Examples:  present 123, fully visible  17 (14%)
```

gcirclest (approved):

calvis (approved):

PR #1691, opened 2026-08-05
Bullet each line of a multiline config validation error

SashaLsnko:
## Summary
- `circleci config validate`/`process` bullets each entry in `result.Errors`, but the API will now return one error whose message spans multiple newline-separated lines instead of several separate error strings
- Previously only the first line of such an error got a bullet, breaking the visual alignment other lines relied on
- Added `printValidationErrors` (shared by validate and process) which splits each error string on `\n` and bullets every resulting line
- Added `TestConfigValidate_MultilineError` acceptance test covering this case, since nothing previously asserted on the bullet formatting
- Added `TestConfigValidate_MultipleErrors`, which sends the same content as separate error strings and asserts against the same golden files — proving both API response shapes (one multiline string vs. several separate strings) render identically

## Test plan
- [x] `task test -- ./acceptance/... -run TestConfigValidate`
- [x] `go build ./...`

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -32,6 +32,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


internal/cmd/config/config.go
```diff
@@ -122,6 +123,19 @@ func resolveOrgID(ctx context.Context, client *apiclient.Client, org, cmdName st
...
 }
 
+// printValidationErrors writes each compilation error as a bulleted line to
+// stderr. An error's message may itself span multiple newline-separated
+// lines (the API sometimes returns one error with embedded detail lines
+// rather than several separate errors), so each line gets its own bullet to
+// keep the output visually aligned.
+func printValidationErrors(ctx context.Context, errs []string) {
+	for _, e := range errs {
+		for _, line := range strings.Split(e, "\n") {
```
briceicle, line 133: ~ might be worth stripping trailing new lines (i.e. `strings.TrimRight(e, "\n")`) before splitting. In case the error message has trailing new lines. As these would produce empty bullet points

briceicle (approved):

internal/cmd/config/config.go
SashaLsnko: I prefer to leave them - that way errors will look the same in UI and CLI

PR #1695, opened 2026-08-05
Bullet each line of a multiline config validation error (v0)

SashaLsnko:
# Checklist

=========

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have read [Contribution Guidelines](https://github.com/CircleCI-Public/circleci-cli/blob/main/CONTRIBUTING.md).

### Internal Checklist

## Changes

=======

- Split each config compilation error's `Message` on `\n` before bulleting it in `configErrorsAsError` (`config/v2_api_client.go`), so every line gets its own `- ` prefix instead of only the first.
- Applied the same fix to `GQLErrorsCollection.Error()` (`config/v1_api_client.go`), used by the legacy GraphQL compile path.
- Added unit tests covering a single multiline error message for both the v2 REST and v1 GraphQL error-formatting paths.

## Rationale

=========

This is the `v0` (old CLI) counterpart to [CircleCI-Public/circleci-cli#1691](https://github.com/CircleCI-Public/circleci-cli/pull/1691), which fixed the current CLI (`main`) for the same underlying change: [orb-service#2001](https://github.com/circleci/orb-service/pull/2001) makes the config-compilation API return one combined error message (with embedded newlines) instead of several separate error strings when `next=true`.

The old CLI doesn't print errors via a shared bulleting helper the way `main` does — instead each API client wraps `response.Errors` into a single `error` value (`config compilation contains errors:\n\t- ...`) at the point the error is returned. That join only bulleted the first line of an error's message, so a combined multiline error would render with a hanging, unbulleted continuation.

## Considerations

==============

- Scoped to the two `Error`/error-formatting functions that actually do the line-joining (`config/v2_api_client.go`, `config/v1_api_client.go`). The `!response.Valid` branches in `config/commands.go` are effectively dead code for both API clients today, since both clients already return an `error` whenever `Errors` is non-empty, short-circuiting before that check is reached — so they were left untouched to keep this change minimal.
- Did not touch the pre-existing doubled `"config compilation contains errors:"` prefix in `v2APIClient.CompileConfig` (line wraps `configErrorsAsError`'s already-prefixed message in another `"config compilation contains errors: %s"`) — unrelated to this fix.

## Screenshots

============

N/A - CLI text output change, no visual/UI screenshots. See test assertions for exact before/after formatting.

Before:
Screenshot 2026-08-05 at 4 53 02 PM

### DIFF
--- a/config/v1_api_client.go
+++ b/config/v1_api_client.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/CircleCI-Public/circleci-cli/api/graphql"
 	"github.com/pkg/errors"
@@ -31,7 +32,11 @@ func (errs GQLErrorsCollection) Error() string {
 	message := "config compilation contains errors:"
 
 	for i := range errs {
-		message += fmt.Sprintf("\n\t- %s", errs[i].Message)
+		// A single error's message may itself span multiple newline-separated
+		// lines, so bullet every line rather than only the first.
+		for _, line := range strings.Split(errs[i].Message, "\n") {
+			message += fmt.Sprintf("\n\t- %s", line)
+		}
 	}
 
 	return message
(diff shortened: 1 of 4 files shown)


SashaLsnko:
We decided that we will only update v1 cli, which was done in https://github.com/CircleCI-Public/circleci-cli/pull/1691

PR #1702, opened 2026-08-06
Delete resource classes by UUID

christian-stephen:
The API addresses resource classes by UUID, but delete sent the `namespace/name` slug, so every attempt failed with `400 {"message":"id is not a valid uuid"}`. The command was unusable against the real API.

Resolve the slug through the namespace listing first, as the legacy CLI did, then delete by UUID. A slug that matches nothing now exits 5 from the lookup rather than surfacing a raw API 404 body.

Delete uses the `/force` endpoint so tokens go with the class, which is what the command already documents and its prompt already warns about. Without it the API returns 409 whenever a token exists.

The fake server routed the slug form, matching the CLI rather than the API, so no test could catch this. It now addresses resource classes by UUID and drops their tokens on delete.

Stacked on #1701, which fixes the namespace listing this lookup depends on. Review that one first.

### DIFF
--- a/internal/apiclient/runner.go
+++ b/internal/apiclient/runner.go
@@ -24,6 +24,9 @@ package apiclient
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/google/uuid"
 )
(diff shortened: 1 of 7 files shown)


internal/apiclient/runner.go
```diff
@@ -109,10 +112,35 @@ func (c *Client) CreateResourceClass(ctx context.Context, resourceClass, descrip
...
+		if rc.ResourceClass == resourceClass {
+			return &rc, nil
+		}
+	}
+	return nil, fmt.Errorf("%w: %q", ErrResourceClassNotFound, resourceClass)
+}
+
+// DeleteResourceClass deletes a runner resource class by its UUID, along with
+// any tokens issued for it.
+func (c *Client) DeleteResourceClass(ctx context.Context, id string) error {
```
pete-woods: id should be a uuid.UUID

pete-woods (approved):

PR #1748, opened 2026-08-18
Add --definition-id flag to run trigger

michael-webster:
Adds `--definition-id` to `circleci run trigger`. When set, routes to `POST .../pipeline/run` with `definition_id` in the body instead of the default `/pipeline` endpoint. Includes an acceptance test covering the new flag.


michael-webster:
Manually verified that triggering and branch defaulting works.

PR #1763, opened 2026-08-25
Cache Go modules and build output in CI

stiyyagura0901:
Every job downloaded Go modules from scratch. One setup-go call now installs the toolchain and restores the module and build caches, each saved automatically at the end of the job — so a job that is cancelled or times out still keeps what it downloaded and compiled.

No cache steps needed anywhere else in the config.

## Verified

The `check` job's steps, from a run on this branch:

```
circleci run setup-go@v0.3.0-145bffb -- --version ... --toolchain-checksum ... --checksum ...
Restoring cache      toolchain
Restoring cache      modules
Restoring cache      build
go version
Install go-task
task ci:check
Uploading test results
Saving cache         at job end
Saving cache         at job end
```

Three restores from one call, both saves at the end. An earlier revision measured the round trip as 394 MiB of module cache restored in ~4s instead of re-downloading. All jobs green.

Every job shares the same cache key, so the first to finish writes it and the rest log `Skipping cache generation, cache already exists`. That is expected and cannot fail a job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,9 @@ parameters:
   go_version_with_patch:
     type: string
     default: 1.26.4
+  setup_go:
+    type: string
+    default: github.com/circleci-functions/setup-go@v0.3.2-1bb7a7a
 
 commands:
   setup:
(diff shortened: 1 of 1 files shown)


pete-woods (commented):
I'd expect this all to be possible with a single invocation of the function. Is it required that module restore is its own call?

Also hoping that go build cache is covered with the same single call.

stiyyagura0901:
> I'd expect this all to be possible with a single invocation of the function. Is it required that module restore is its own call?
> 
> Also hoping that go build cache is covered with the same single call.

@pete-woods 
yes, updated the commit to so one command Installs the toolchain and restores the module and build caches, saved at job end

image

.circleci/config.yml
```diff
@@ -12,9 +15,14 @@ commands:
       - install-go
       - install-task
 
+  # Installs the toolchain and restores the module, build and lint caches, saved at job end.
   install-go:
     steps:
-      - run: circleci run circleci/setup-go@0.0.10-5577d2a -- --version << pipeline.parameters.go_version_with_patch >> --go-cache-checksum '{{ checksum "go.sum" }}_{{ checksum "clikit/go.sum" }}_{{ checksum "tools/go.sum" }}'
+      - run: circleci run << pipeline.parameters.setup_go >> --
+          --version << pipeline.parameters.go_version_with_patch >>
+          --toolchain-checksum '{{ checksum "go.sum" }}_{{ checksum "clikit/go.sum" }}_{{ checksum "tools/go.sum" }}'
```
pete-woods:
```suggestion
```

pete-woods (approved):

PR #1772, opened 2026-08-27
Send CircleCI-Caller header from condensed step output CLI commands

joaquin-sm-cci:
## Summary

- `GetJobStdoutCondensed()` now accepts a `caller string` parameter and sets it as the `CircleCI-Caller` request header when non-empty
- `job output get --condensed` sends `circleci-cli/job-output-get`
- `run get --failure-report` sends `circleci-cli/run-get-failure-report`

llmops-service already reads this header on the `GET /api/v3/jobs/:id/stdout/condensed` endpoint and forwards it as `source` in the `api-step-output-condensed` Segment event. Until now that field was always blank for CLI calls — CLI usage was indistinguishable from web or direct API calls in analytics. public-api-service proxies this route with `httputil.ReverseProxy` which forwards all headers verbatim, so no changes are needed outside the CLI.

Closes CF2-907

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./internal/apiclient/... ./internal/cmd/job/... ./internal/cmd/run/...` passes
- [ ] Verify `CircleCI-Caller` header appears in an HTTP trace against a real or fake server for both flows

### DIFF
--- a/internal/apiclient/job.go
+++ b/internal/apiclient/job.go
@@ -411,14 +411,20 @@ func (c *Client) GetJobStdout(ctx context.Context, jobID uuid.UUID, execution, s
 // GetJobStdoutCondensed fetches a step's stdout condensed to its most
 // error-relevant lines (noisy/repetitive output filtered out server-side).
 // The endpoint returns raw text (octet-stream).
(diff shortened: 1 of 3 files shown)


cloverfern:
We may move away from adding these headers (according to a recent review I got) we may need to go back to the drawing board on how we would achieve this

pete-woods:
We already send clear user agents via the standard headers, that identify the CLI.

PR #1776, opened 2026-08-31
Add runner smoke tests against the live API

olcarl-circleci:
## Summary

Adds a `smoke/` package that drives the built `circleci` binary against the **live**
CircleCI API, covering the management cycle for a self-hosted runner. ONP-3563.

## Why

`acceptance/runner_test.go` runs the real binary against `internal/testing/fakes`.
That proves the CLI's own wiring — flags, exit codes, output shape — but not that
the CLI and the runner API still agree with each other. Nothing in this repo has
ever tested against a live API, so a compatibility break between the CLI and
runner-admin would reach a customer before it reached a test.

## Changes

- `smoke/runner_test.go`, behind a `smoke` build tag so `./...` excludes it and
  `task test` never sees it. Skips unless credentials are present.
  - `TestRunnerSmoke` — create a resource class, generate an agent config, claim as
    an agent, read resource classes / tokens / instances back, tear it all down.
  - `TestRunnerSmokeFailures` — no token, bad token, missing resource class,
    insufficient permissions.
  - `TestRunnerSmokeClaimQueuedTask` — queue a real job on the resource class it
    just created, claim the task, hand it back. Opt-in on a project slug.
  - `TestCreatedAtFromName` — pure, no credentials; pins which names the orphan
    sweep will and will not delete.
- `smoke/README.md` — the three variables to export, and the pipeline parameter a
  project's config needs for the queued-task test. One optional variable
  covers the insufficient-permission case.
- `Taskfile.yml` — `test:smoke`, and `lint:smoke` so a tagged-out file cannot rot
  unnoticed.
- `CLAUDE.md` — pointer, plus renaming the existing "Smoke tests" heading to
  "Sanity checks", since it meant `go run ./cmd/circleci --help` and now collides.

## Notes for review

**The agent is simulated, not installed.** The tests make the same requests an
agent makes — `POST /runner/claim`, `GET /api/v2/task/config` for the task id,
`POST /runner/unclaim` — so no runner binary, container, or executed job is
involved. The claim handler registers the runner before it asks the distributor
for work, so a claim against an empty resource class still produces an instance
and returns 204.

**Order matters in the queued-task test.** A claimed task holds its job queued
until the claim times out, and cancelling the run reports success while the job
sits there for minutes. Unclaiming first releases the task and lets the cancel
take effect immediately; cancelling first drops the task from the dispatcher,
after which unclaim answers 404. Both directions are commented in place.

**No golden files.** Every response carries live UUIDs and timestamps, so
assertions are on exit codes and `--json` fields. This is a deliberate departure
from rule 7 in `agents/14-testing.md`; `internal/conventions` only polices the UI
packages, so nothing enforced is being bypassed.

**Reaping is in two parts,** because `t.Cleanup` does not survive a Ctrl-C or a
`go test -timeout` kill. Each run registers deletion of its own resource class
before creating it, and sweeps `cli-smoke-*` classes left behind by an interrupted
earlier run. Deleting a resource class cascades its tokens server-side. The prefix
is deliberately specific: the shared `circleci-runner` namespace already holds
`prod-smoke-test-*` and `provisioner-smoke-test-*` from other suites, and the
sweep must never touch those.

## Testing

- `task test` — 2285 tests, 2 skipped, 0 failures.
- `task test:smoke` against production — 28 tests, 0 failures, 0 skips, ~21s,
  nothing left behind in the target org.
- `gofmt`, `go vet` (tagged and untagged), `task lint:smoke`, and the license
  header check all clean. `go.mod` untouched.
- `go list ./...` omits the package, so `task test` and `task check` are
  unaffected by it.

Not wired into CI: ONP-3564 was closed as Won't Do, so this is documented for
manual runs.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,6 +98,8 @@ cmd/circleci/main.go      Entry point. Cobra bootstrap + top-level error handlin
                           (Lives under cmd/circleci/ so `go install .../cmd/circleci`
                           produces a binary named `circleci`, not `circleci-cli`.)
 acceptance/               Acceptance tests — exec the compiled binary against fake servers.
+smoke/                    Smoke tests — exec the compiled binary against the live API.
+                          Build-tagged `smoke`, credential-gated. See smoke/README.md.
 agents/                   Design guidelines (normative — see above).
 clikit/                   Terminal I/O + presentation layer — its own Go module, published
                           for extension authors (below).
@@ -374,6 +376,7 @@ task test        # all tests (unit + acceptance), root module + clikit, -race -c
 task check       # all static checks: lint (both modules), license headers, mod-tidy, release-check
 task fix         # auto-fix what `task check` flags: fmt, license, mod-tidy, lint --fix
 task build       # build the circleci binary to dist/circleci
+task test:smoke  # runner smoke tests against the live API; needs credentials, see smoke/README.md
 ```
 
 All three cover both Go modules. `clikit` is separate, so `./...` does not reach it — `task test`
@@ -405,7 +408,9 @@ That last one matters here: `internal/cmd/root/testdata/help/` holds a golden `-
 for every command in the tree, so touching a `Short`/`Long`/`Example` or adding a flag updates
 dozens of files. Run it with `-update` and commit the regenerated goldens alongside the change.
 
-**Smoke tests** (compile-and-run without building a binary):
+**Sanity checks** (compile-and-run without building a binary). Not to be confused with
+the runner smoke suite in [`smoke/`](smoke/README.md), which drives the built binary
+against the live API:
 
 ```sh
 go run ./cmd/circleci --help             # basic smoke test
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -72,6 +72,13 @@ tasks:
     cmds:
       - go tool -modfile tools/go.mod gotestsum -- -race -count=1 {{.ARGS}}
 
+  test:smoke:
+    desc: Run the runner smoke tests against the live API (needs credentials, see smoke/README.md)
+    vars:
+      ARGS: '{{default "./smoke/..." .CLI_ARGS}}'
+    cmds:
+      - go tool -modfile tools/go.mod gotestsum -- -tags smoke -race -count=1 -timeout 20m {{.ARGS}}
+
   check:
     desc: Run all static checks
     cmds:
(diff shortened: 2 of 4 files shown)


olcarl-circleci:
Superseded — closing in favour of building this in `circleci/runner-smoke-tests` instead.

## Why

That repo is purpose-built for self-hosted runner smoke tests, is owned by the same team, and already solves the parts that made this PR awkward:

- **It already installs the latest CLI.** `cli_setup` in `.circleci/config/shared.yml` does `curl .../CircleCI-Public/circleci-cli/main/install.sh | DESTDIR=~/bin bash`, so it tests the released artifact rather than a locally built binary.
- **Credentials already exist.** `CIRCLE_API_TOKEN` is provisioned in the `runner-smoke-tests` context. Wiring live credentials into this repo's CI was the blocker here, and it is worse than usual because this repo is public: a context on a public project needs care that a private test repo does not.
- **It already runs the CLI's runner CRUD nightly** — `resource-class create --generate-token --json`, `resource-class delete --force`, `resource-class list --namespace --json` — but as fixtures, never asserted. The real gap is assertions, not invocation.
- **Orphan reaping, unique naming and cleanup schedules already exist** there (`delete-all-resource-classes` with a pattern, `*-cleanup` scheduled pipelines, `<< pipeline.number >>`-based names), so a good deal of this PR was rebuilding what was already written.

## Trade-off accepted

A CLI regression in `circleci runner` will now be caught by a nightly in another repo rather than by the PR that caused it. That is slower feedback, deliberately traded for the test living with the team that owns the runner API.

## Carried forward

The findings this work produced are independent of where the tests live and are worth filing regardless:

- `AppURL` prefixes `app.` unconditionally, so `runner open` builds an unreachable URL on CircleCI Server.
- `--force` on `resource-class delete` changed meaning from v0 (cascade tokens) to v1 (skip the prompt); the non-force API path is now unreachable from the CLI.
- `runner open --help` documents a `--host` flag that does not exist.
- `runner open` returns a bare `fmt.Errorf` for a malformed `--org`, contrary to the repo's structured-error rule, and exits 1 rather than `ExitBadArguments`.
- `runner.not_enabled` is mapped to any 404 on the list paths, so "namespace not claimed", "no permission" and "runners unavailable" all report as a token or plan problem.
- `runner config` emits only `api.auth_token` — not `runner.name`, which the agent requires — see ONP-3558.

ONP-3564 stays closed: the nightly and cleanup schedules come for free with the existing setup in `runner-smoke-tests`.

PR #1826, opened 2026-09-16
Show the organization slug in org list

EnoshAnwar:
The slug is what every `--org` flag takes, and `circleci org list` was the one place you could not find it. It printed id, name and VCS only, so the value had to come from the raw API or the web app.

```
# Organizations
| Slug                              | Name           | Organization ID                        | VCS    |
| --------------------------------- | -------------- | -------------------------------------- | ------ |
| `gh/myorg`                        | myorg          | `a0000000-0000-4000-8000-0000000c0002` | github |
| `circleci/2V3EWrY7hKnRnfkk8cm4Ai` | standalone-org | `a0000000-0000-4000-8000-0000000c0003` | -      |
```

`slug` is added to `--json` too, so `circleci org list --json --jq '.[].slug'` is now the way to get them.

## Why it was missing

`GET /api/v3/orgs` carries no slug, and a slug cannot be derived from a UUID, which is why moving `org list` to that endpoint dropped it. A standalone org's slug is `circleci/<short id>`, and that short id is not an encoding of the org UUID, so there is no client-side shortcut.

The listing is now decorated from `GET /api/v2/me/collaborations`, keyed on the org UUID. Slug leads the table because it is the value a reader is there for; the ID column stays because a few flags take it instead.

## Failure behaviour

That lookup is supplementary, so a failure is reported on stderr and the table still prints with the slug column blank, rather than losing the whole listing over it.

A blank slug is a real case rather than a hypothetical, which is worth knowing before reading the `-` in that column as a bug: an organization can appear in `/api/v3/orgs` without appearing in collaborations. I hit exactly that against production, on a Bitbucket org.

The cost is one extra API call per `org list`.

## Tests

`TestOrgList` and `TestOrgList_JSON` now assert the slug, with the fake serving both orgs through collaborations the way the real API does. `TestOrgList_NoSlugsAvailable` covers the degraded path.

`task test` and `task check` are clean, aside from pre-existing `internal/gitremote` and `internal/orbinit` failures that reproduce on an untouched tree when `commit.gpgSign` is set locally. Also verified against production, where it returns slugs for 8 organizations and a blank for the one described above.


conormcd (changes requested):
Please don't do this. The entire point of moving to /api/v3/orgs was to stop using the collaborations endpoints (which are terrible for many reasons).

If we need slugs we can do this another way.

